PR #77, opened 2023-03-01
Fix Storage Example

wyattjoh:
The example had a typo referencing the wrong variable `r` instead of the correct variable `a` when accessing `db`.


rgrandl (approved):
Thanks for fixing the typo.

PR #82, opened 2023-03-02
add documentation about registry store folder environment variable XD…

shivas:
add documentation about the registry store folder environment variable XDG_DATA_HOME.
Otherwise, Windows won't be able to find registry data and all `weaver` commands won't see any deployments


shivas:
Closing issue, it's not windows related

PR #101, opened 2023-03-04
Fix ssh deployer

rgrandl:
While renaming to service weaver, we renamed some things that didn't suppose to be renamed.


spetrovic77 (approved):
Sorry, this was my bad.

PR #102, opened 2023-03-05
fix typos

mouuii:


internal/net/call/call_test.go
```diff
@@ -29,12 +29,12 @@ import (
 	"testing"
 	"time"
 
-	sdktrace "go.opentelemetry.io/otel/sdk/trace"
-	"go.opentelemetry.io/otel/trace"
 	"github.com/ServiceWeaver/weaver/internal/cond"
 	"github.com/ServiceWeaver/weaver/internal/net/call"
 	"github.com/ServiceWeaver/weaver/internal/traceio"
 	"github.com/ServiceWeaver/weaver/runtime/logging"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
```
rgrandl, line 36: Can you don't change the order of these imports?

rgrandl (commented):
Thanks for fixing the typos. I left a small comment.

internal/net/call/call_test.go
spetrovic77: I've seen this happen to me too, I think different IDEs have different ideas maybe?

spetrovic77: Maybe okay to let it happen in a few files, as long it's not the entire repo :).

spetrovic77 (approved):
Wath typos? ;)

Thanks for fixing!

internal/net/call/call_test.go
mouuii: thank you, this is because of the goland IDE , I have turned it off

PR #107, opened 2023-03-05
Move to a non-cgo SQLite implementation.

spetrovic77:
This will help avoid some compilation issues on MacOS and Windows.


examples/chat/sqlstore.go
```diff
@@ -23,7 +23,7 @@ import (
 
 	"github.com/ServiceWeaver/weaver"
 	_ "github.com/go-sql-driver/mysql"
```
rgrandl, line 25: Any reason we keep this dependency here?

examples/chat/sqlstore.go
```diff
@@ -158,8 +158,8 @@ func (s *sqlStore) Init(ctx context.Context) error {
 		},
 	} {
 		query := q.query
-		if cfg.Driver == "sqlite3" {
-			// sqlite3 does not work with AUTO_INCREMENT specified on primary keys.
+		if cfg.Driver == "sqlite" {
+			// sqlite does not work with AUTO_INCREMENT specified on primary keys.
```
rgrandl, line 162: Is this still true?

rgrandl (approved):

johanbrandhorst (approved):
Thanks for this 😄

examples/chat/sqlstore.go
spetrovic77: Yes, I believe it is.

examples/chat/sqlstore.go
spetrovic77: For the GKE case, which uses the Cloud SQL MySQL instance.

spetrovic77:
> 

You're welcome, thanks for bringing it to our attention.

PR #115, opened 2023-03-06
Improved ExportListener API.

mwhittaker:
Before this PR, the flow for exporting a listener was a little complicated. A weavelet would receive two listener related options via a protos.Weavelet: `use_localhost` and `process_picks_ports`. If `use_localhost` was true, a weavelet would listen on localhost. Otherwise, it would listen on `$HOSTNAME`. If `process_picks_ports` was true, the weavelet would listen on port 0. Otherwise, it would call ExportListener to get a port. Thus, ExportListener sometimes exported a listener and sometimes didn't (and returned a port).

This PR simplifies the API by removing `use_localhost` and `process_picks_ports` options completely and by splitting ExportListener in two. First, a weavelet calls GetAddress to get the address it should listen on for a listener.  For example, the single process deployer may return "localhost:0", while the SSH deployer may return "$HOSTNAME:0". Second, the weavelet listens on this address and calls ExportListener to export the listener.

This flow is much simpler, but it does require an additional RPC over the pipe. However, exporting listeners is not on critical path, plus there is already significant traffic over the pipe (for every log entry, to export metrics, etc.). Overall, I think the performance hit is negligible.


ghemawat (approved):
Much nicer. Thanks.

PR #116, opened 2023-03-06
Add a section on "what to contribute to" to the contrib file.

spetrovic77:
This is in response to a request from a community member on how they can get involved.


CONTRIBUTING.md
```diff
@@ -131,7 +131,19 @@ To make sure changes are well coordinated, we ask you to discuss any significant
...
 [new issue][new_issue] or claim an [existing one][issues].
 
+## What to contribute to?
+
+Here are the current areas where the community contributions would be most
+valuable:
+
+* Implement deployers for platforms other than local, SSH, and GKE.
+* Bug fixes for the existing libraries.
+* Non-Go language support. (Note: please reach out to us on our
```
ghemawat, line 141: I think we need some kind of agreed upon design before asking people to work on it.

Asking before starting work is good, but we should be up-front about the design work needed before work on language support can start.

PR #117, opened 2023-03-06
Added GitHub workflow to generate & publish site.

mwhittaker:
This PR adds a GitHub workflow to build and publish the website every time the main branch is pushed to. The website is currently hosted on https://github.com/ServiceWeaver/website, but we'd like to switch to the main weaver repo.

In the future, we may not want to update the website on every change. We might want to update it when we update the version with a new tag?


rgrandl (approved):
Nice, thanks Michael.

PR #121, opened 2023-03-07
Fixing escape issues with curl and zsh

danielmapar:
Issue [here](https://github.com/ServiceWeaver/website/issues/2)

### DIFF
--- a/website/docs.md
+++ b/website/docs.md
@@ -158,7 +158,7 @@ hello listener available on 127.0.0.1:12345
 In a separate terminal, curl the server to receive a greeting:
 
 ```console
-$ curl localhost:12345/hello?name=Weaver
+$ curl localhost:12345/hello\?name\=Weaver
 Hello, Weaver!
 ```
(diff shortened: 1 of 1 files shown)


mwhittaker:
Hi @danielmapar! Thanks for the fix. I'm not super familiar with zsh. Could we escape things using single quotes instead of backslashes like this: 

```zsh
curl 'localhost:12345/hello?name=Weaver'
```

If that does work, I think it's a bit easier to read?

danielmapar:
@mwhittaker very good point, and yes `"` works as well. I updated the PR to use `"` instead of escaping chars with `\`.

mwhittaker (approved):
LGTM!

danielmapar:
I re-based the PR branch @mwhittaker, let me know if I need to do any further steps to get it merged.

mwhittaker:
Merged! Thanks so much for your contribution @danielmapar!

PR #122, opened 2023-03-07
Fixed some visual bugs on mobile website.

mwhittaker:
This PR fixes two bugs that appeared when the website was viewed on mobile. First, the jumbotron on the landing page was squeezed very thin. I fixed this by removing the left and right margins on small pages. Second, the top navigation header had elements overlapping vertically. I fixed this by adjusting padding.

| Before | After |
| --- | --- |
| 3KtjtCAcCaUrmME | 5hrdzVnRDefiM7D |

### DIFF
--- a/website/assets/css/common.css
+++ b/website/assets/css/common.css
@@ -69,9 +69,7 @@ header .active {
 header a {
   text-decoration: none;
   color: black;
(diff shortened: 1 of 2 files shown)


spetrovic77 (approved):
Nice!

Issue #90, opened 2023-03-03
Enabling the errcheck linter


dnephin:
Congratulations on the first release! ServiceWeaver looks very interesting and I'm looking for ways to contribute.

I noticed that `errcheck` is disabled with a comment about too many errors. I've found this linter to be useful because unhandled errors can often cause bugs.

Would you be interested in a few small PRs to fix or ignore the current warnings so that the `errcheck` linter can be enabled?

I'll leave some more detail in a comment below about how I'd approach that change.

dnephin:
Initially I'd want to enable it only for production code (not tests), to limit the number of changes required to get it enabled. Something like this:

```yaml
issues:
  exclude-rules:
    - path: '_test\.go'
      linters: [errcheck]
```

I think a bunch of the unhandled errors are safe to ignore (ex: writing an HTTP response from a handler). Those can be ignored with either:

```go
_, _ = w.Write(...)
// or
w.Write(...) //nolint:errcheck
```

Do you have a preference on one of those styles?

----

There are a few cases where it seems like the error simply needs to be returned:

https://github.com/ServiceWeaver/weaver/blob/2ae2abb650843ff5c6dbc76c9934a715fa98f2d0/runtime/logging/query.go#L370-L371

https://github.com/ServiceWeaver/weaver/blob/2ae2abb650843ff5c6dbc76c9934a715fa98f2d0/internal/tool/ssh/impl/babysitter.go#L103-L104

https://github.com/ServiceWeaver/weaver/blob/2ae2abb650843ff5c6dbc76c9934a715fa98f2d0/internal/net/call/call.go#L283

or logged:

https://github.com/ServiceWeaver/weaver/blob/2ae2abb650843ff5c6dbc76c9934a715fa98f2d0/internal/tool/ssh/deploy.go#L107

https://github.com/ServiceWeaver/weaver/blob/2ae2abb650843ff5c6dbc76c9934a715fa98f2d0/internal/status/dashboard.go#L177

A log here to inform the user that the port is in use would probably help avoid confusion when perfetto isn't able to display the trace.

https://github.com/ServiceWeaver/weaver/blob/2ae2abb650843ff5c6dbc76c9934a715fa98f2d0/runtime/perfetto/db.go#L526

----

The remaining cases of unhandled errors seems to be mostly in these two categories:
1. starting goroutines with a function that returns an error - in a few of these cases the only possible return value is `ctx.Err`. I wonder if the error return value could be removed from the function signature in those cases.
[...]

PR #128, opened 2023-03-07
Handle or ignore many unhandled errors

dnephin:
Related to #90 

This PR addresses many of the unhandled errors, there are two common cases not handled here that will be done in a separate PR. This gets us closer to enabling the `errcheck` linter.

I've split up the changes into multiple commits. Each commit groups related changes by how the error is handled (ignored, removed, logged, or returned). Changes may be best viewed by individual commit. I'll leave some comments and questions inline. The last commit has some TODOs. I can remove that commit if we don't have any easy way of addressing those TODOs.

My IDE has sorted some of the import blocks alphabetically. I attempted to turn off that feature, but it was still sorting them. I hope that's not a problem, but I can revert those changes if necessary.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -40,11 +40,11 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/go/bin/golangci-lint
-          key: golangci-lint-v1.51.0
+          key: golangci-lint-v1.51.2
         if: ${{ matrix.command == 'lint' }}
 
       - name: Install linter
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.0
+        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2
         if: ${{ matrix.command == 'lint' }}
 
       - name: ${{ matrix.command }}
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -15,6 +15,8 @@
 linters:
   disable:
     - errcheck # Currently too many places are ignoring errors
+  enable:
+    - nolintlint
 linters-settings:
   staticcheck:
     checks:
--- a/internal/status/dashboard.go
+++ b/internal/status/dashboard.go
@@ -134,12 +134,12 @@ Flags:
 
 			traceDB, err := perfetto.Open(ctx)
 			if err != nil {
-				fmt.Fprintln(os.Stderr, "cannot open Perfetto database: %w", err)
+				fmt.Fprintf(os.Stderr, "cannot open Perfetto database: %v\n", err)
 			}
 			go traceDB.Serve(ctx)
 
 			fmt.Fprintln(os.Stderr, "Dashboard available at:", url)
-			go browser.OpenURL(url)
+			go browser.OpenURL(url) //nolint:errcheck // browser open is optional
 			return http.Serve(lis, nil)
 		},
 	}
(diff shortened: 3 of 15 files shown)


.golangci.yaml
```diff
@@ -15,6 +15,8 @@
 linters:
   disable:
     - errcheck # Currently too many places are ignoring errors
+  enable:
+    - nolintlint
```
dnephin, line 19: By enabling this linter `golangci-lint` will warn if any `nolint` comments become stale and are no longer required.

It looks like it is complaining about some of the existing `nolint` comments.

internal/status/dashboard.go
```diff
@@ -174,6 +174,7 @@ func (d *dashboard) handleIndex(w http.ResponseWriter, r *http.Request) {
 		Tool:     d.spec.Tool,
 		Statuses: statuses,
 	}
+	// TODO: unhandled error, would it be reasonable to panic here?
 	indexTemplate.Execute(w, content)
```
dnephin, line 178: I believe the only way for this `template.Execute` to error would be a programming error, the data being passed to the template does not match the values expected by the template, or the template itself has a syntax error.

Would you be ok with a panic here? I believe this category of error would generally be caught during development.

Another option would be to log the error, but I'm not sure how to get the appropriate logger in this HTTP handler.

internal/status/registry.go
```diff
@@ -201,7 +201,8 @@ func (r *Registry) List(ctx context.Context) ([]Registration, error) {
 		// the server is dead, we consider the deployment dead, and we garbage
 		// collect its registration.
 		if r.dead(ctx, reg) {
-			r.Unregister(ctx, reg.DeploymentId)
+			// TODO: log this error instead of ignoring it
+			r.Unregister(ctx, reg.DeploymentId) //nolint:errcheck
```
dnephin, line 205: I'm not sure how important it is to log this error. I believe this operation is some form of cleanup when a previous deployment is terminated forcefully. I suspect it would be helpful to log the error if the deployment can't be cleaned up.

How would I get access to the logger here? Is there currently any way to get a logger from `ctx` ? I did not find one.

.golangci.yaml
dnephin: Updating `golangci-lint` to the latest patch version has resolved the problem.

.github/workflows/go.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         if: ${{ matrix.command == 'lint' }}
 
       - name: Install linter
-        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.0
+        run: go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.51.2
```
mwhittaker, line 47: Could you update the `golangci-lint-v1.51.0` key on line 43 as well?

.golangci.yaml
mwhittaker: Very nice! And thanks for explaining!

internal/status/registry.go
mwhittaker: Maybe we should just write the error to `os.Stderr`? I don't think there's a nice way to get a logger here. What do you think?

mwhittaker (approved):
Very very nice change @dnephin! It's very much appreciated!

internal/status/dashboard.go
mwhittaker: Panic seems good to me!

internal/status/registry.go
dnephin: I looked at the callers to this function, and I think that should work well. I see it being called primarily from CLI commands, and from one HTTP handler that is started by a CLI command. I've made that change in the latest commit!

Storing and retrieving a `Logger` from `ctx` has some nice properties, but maybe is not appropriate in this case. Either way I imagine it's not a priority right now, so maybe something to explore in the future.

dnephin (commented):
Thanks for the review! I've made those changes and removed the TODO

mwhittaker:
GitHub says this branch is out-of-date with the base branch. Mind rebasing and squashing on top of main?

dnephin:
Rebase and squash done!

mwhittaker (approved):

PR #133, opened 2023-03-08
Encoding/decoding benchmarks report approximate throughput.

ghemawat:


rgrandl (approved):
I am trying to wrap up my head around it. If I understand correctly, this reflects the average number of bytes encoded per payload? 

Somehow in my mind when I think about throughput I think about something over time, but maybe the b.Run(...) takes care of that normalization. I forgot a little bit how these benchmarks work.

PR #135, opened 2023-03-09
Fix usage of Roboto font

loganrosen:
I noticed that this website uses the Roboto font without importing it as a web font. This causes it to show up as the default serif font on my system. To remedy this, I've added the code from https://fonts.google.com/specimen/Roboto to import the web font, and I've also modified the CSS to have a sans-serif fallback in case the font fails to load.

### DIFF
--- a/website/assets/css/common.css
+++ b/website/assets/css/common.css
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 body {
-  font-family: "Roboto";
+  font-family: "Roboto", sans-serif;
   padding: 0;
   margin: 0;
 }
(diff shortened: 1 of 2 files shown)


mwhittaker (approved):
Thanks so much @loganrosen! And thanks for linking to https://fonts.google.com/specimen/Roboto explaining where you got the code. I'm not super familiar with HTML, so that was helpful for me :) 

By the way, if you're looking for more contributions, I think it would be nice if we used Roboto Mono for the code snippets on the site. If you know how to get that working, that would be awesome 🔥

PR #136, opened 2023-03-09
Use Roboto Mono for code blocks

loganrosen:
Per @mwhittaker in #135, it would be better for the code blocks to use Roboto Mono as opposed to the default monospace font chosen by the browser. I've changed the code to use Roboto Mono at an appropriate size accordingly (with a fallback to the default monotspace font if needed).

### DIFF
--- a/website/assets/css/common.css
+++ b/website/assets/css/common.css
@@ -39,6 +39,11 @@ pre {
   border-radius: 6px;
 }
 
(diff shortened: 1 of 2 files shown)


mwhittaker (approved):
Wow, that was fast! Site looks much better now. Thanks @loganrosen!

PR #137, opened 2023-03-09
Remove bool return from runWeavelet

dnephin:
While browsing the code I noticed that the `bool` return from `Envelope.runWeavelet` was always `false`. Removing the return value simplifies the caller slightly. I'm opening this PR to see if you are interested in this small simplification. Thanks for considering it!

I often try to avoid `(bool, error)` returns, and opt to use a sentinel error value when possible. That may be an option in the future if `runWeavelet` needs to indicate a speicific error case.

### DIFF
--- a/runtime/envelope/envelope.go
+++ b/runtime/envelope/envelope.go
@@ -213,14 +213,14 @@ func (e *Envelope) Run(ctx context.Context) error {
 	}
 
 	for r := retry.BeginWithOptions(e.opts.Retry); r.Continue(ctx); {
-		badArg, err := e.runWeavelet(ctx)
+		err := e.runWeavelet(ctx)
 		if e.isStopped() {
-			// When an envelope is stopped, it kills its subprocesses. Theses
+			// When an envelope is stopped, it kills its subprocesses. These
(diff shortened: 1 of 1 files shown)


runtime/envelope/envelope.go
```diff
@@ -263,7 +263,7 @@ func (e *Envelope) RunProfiling(_ context.Context, req *protos.RunProfiling) (*p
 // runWeavelet with the provided environment and wait for it to terminate.
 //
 // It returns true,_ if some incoming argument is bad any retries should be stopped.
```
mwhittaker, line 265: Delete this line now that it's obsolete?

mwhittaker (approved):
Nice find! Much better now! :shipit:

runtime/envelope/envelope.go
dnephin: Ah, yes! I've pushed another commit that removes this line

mwhittaker:
GitHub says this branch is out-of-date with the base branch. Mind rebasing and squashing on top of main?

dnephin:
Rebase and squash done!

PR #138, opened 2023-03-09
Move CreateAndRunEnvelopeForMain to weavetest

dnephin:
While looking at handling the last few remaining unhandled errors for #90 I noticed that two of those unhandled errors are in `Babysitter.CreateAndRunEnvelopeForMain`. The errors are returned by `Run` functions that run in goroutines (`Envelope.Run`, `EnvelopeConn.Run`). In the production code path these errors are handled (at least logged), but in `weavertest` they are dropped. I'm opening this PR to get your thoughts on one option for handling these errors.

The only caller of `Babysitter.CreateAndRunEnvelopeForMain` was `weavertest.initMultiProcess`, so I started by inlining `CreateAndRunEnvelopeForMain` into the caller. This removed the need for the `dontGC` package-level variable. The references to the unused `os.File` are held by a closure passed to `t.Cleanup`. This should prevent them from being GC'ed (the tests in `weavertest/internal` appear to confirm). 

I'll add some comments inline about how this change might make it easier to manage the lifecycle of goroutine from `weavertest.Init`.

### DIFF
--- a/weavertest/multi.go
+++ b/weavertest/multi.go
@@ -26,11 +26,14 @@ import (
 
 	"github.com/ServiceWeaver/weaver"
 	"github.com/ServiceWeaver/weaver/internal/babysitter"
+	"github.com/ServiceWeaver/weaver/internal/envelope/conn"
 	"github.com/ServiceWeaver/weaver/runtime"
 	"github.com/ServiceWeaver/weaver/runtime/codegen"
 	"github.com/ServiceWeaver/weaver/runtime/colors"
+	"github.com/ServiceWeaver/weaver/runtime/envelope"
 	"github.com/ServiceWeaver/weaver/runtime/logging"
 	"github.com/ServiceWeaver/weaver/runtime/protos"
+	"github.com/ServiceWeaver/weaver/runtime/retry"
 	"github.com/google/uuid"
 )
 
@@ -71,6 +74,7 @@ func initMultiProcess(ctx context.Context, t testing.TB, config string) weaver.I
 	stopLog := false
 	ctx, cancelFunc := context.WithCancel(ctx)
 	logSaver := func(e *protos.LogEntry) {
+		t.Helper()
 		// TODO(mwhittaker): Capture the main process' log output and report it
 		// using t.Log as well.
 		mu.Lock()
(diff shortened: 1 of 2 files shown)


weavertest/multi.go
```diff
@@ -86,19 +90,81 @@ func initMultiProcess(ctx context.Context, t testing.TB, config string) weaver.I
...
+		Retry:           retry.DefaultOptions,
+		GetEnvelopeConn: func() *conn.EnvelopeConn { return econn },
+	}
+	e, err := envelope.NewEnvelope(weaveletInfo, dep.App, b, options)
+	if err != nil {
+		t.Fatal(fmt.Errorf("cannot create envelope: %v", err))
+	}
+
+	go econn.Run()
+	go e.Run(ctx)
```
dnephin, line 130: Now that this logic is in a function with `t.Cleanup`, these errors could be  handled by starting the goroutines with https://pkg.go.dev/golang.org/x/sync/errgroup.

The `t.Cleanup` function can call `group.Wait` and receive the errors. More about why that doesn't work yet below.

Another option would be to simply send the error to a channel, and check that the channel is empty in `t.Cleanup`. That avoids the need to wait on the goroutine exit.

weavertest/multi.go
```diff
@@ -86,19 +90,81 @@ func initMultiProcess(ctx context.Context, t testing.TB, config string) weaver.I
...
+		//
+		// To avoid this, we have to ensure that the pipes are not garbage
+		// collected prematurely. This is done by holding a reference to the
+		// os.File in the closure passed to t.Cleanup
+		//
+		// [1]: https://pkg.go.dev/os#File.Fd
+		// [2]: https://go.dev/play/p/9JG7voL2oHP
+		_ = fromWeaveletWriter
+		_ = toWeaveletReader
+
```
dnephin, line 167: When `weavertest.Init` is called multple times by the same test package it may leak goroutines and keep resources around after the test has exited. I've found that can cause tests to flake, but I'm not sure if that will be a problem for serviceweaver applications.

I was thinking that maybe this would be a good place to close the `toWeaveletReader` and `fromWeavletReader`, so that the goroutines could shutdown gracefully. And then call `group.Wait` before returning so that all of the resources are stopped/cleaned up before the next test starts.

Do you think that will be important for testing serviceweaver applications?

---

I attempted that change, but it resulted in all kinds of problems (errors from the goroutines failing to `proto.Read`, and sometimes tests hanging until the timeout). I'll need to understand serviceweaver better before I'm able to fix those problems.

weavertest/multi.go
dnephin: After posting the initial change I realized there is a 3rd option that is much easier. I can use `t.Error` here to capture the error now that this has access to `testing.T`. I've made that change in the latest commit and marked the PR ready for review.

weavertest/multi.go
mwhittaker: > When weavertest.Init is called multple times by the same test package it may leak goroutines and keep resources around after the test has exited. I've found that can cause tests to flake, but I'm not sure if it will be a problem for serviceweaver applications.

Great point! I think it is a problem. As you've seen, the code right now does a bad job cleaning everything up when a test ends.  The `GetEnvelopeConn` option and the way it interacts with a babysitter is also confusing. I think cleaning all this code up would be amazing (but definitely tricky 😢). 

> I was thinking that maybe this would be a good place to close the toWeaveletReader and fromWeavletReader, so that the goroutines could shutdown gracefully. And then call errgroup.Group.Wait (or sync.WaitGroup.Wait`) before returning so that all of the resources are stopped/cleaned up before the next test starts. 

Woah, I've never seen `errgroup` before. Looks super useful!

> I attempted that change, but it resulted in all kinds of problems (errors from the goroutines failing to proto.Read, and sometimes tests hanging until the timeout). I'll need to understand serviceweaver better before I'm able to fix those problems.

I'm definitely not sure what's going wrong. This code is nasty and brittle. If I had to guess, I think it's related to the comment above about the "very subtle bug we have to avoid". I think the test ends, but a subprocess is still running with access to the `fromWeaveletWriter` and `toWeaveletReader` pipes. The subprocess never expects these to close; it closes them itself. So, if we do close them and the subprocess is running, it produces all kinds of nasty errors. I think leaving things as they are now sounds good to me.

weavertest/multi.go
```diff
@@ -86,26 +90,95 @@ func initMultiProcess(ctx context.Context, t testing.TB, config string) weaver.I
...
+		// locally to reproduce this behavior. If the program doesn't crash on your
+		// machine, try increasing the number of iterations.
+		//
+		// To avoid this, we have to ensure that the pipes are not garbage
+		// collected prematurely. This is done by holding a reference to the
+		// os.File in the closure passed to t.Cleanup
+		//
+		// [1]: https://pkg.go.dev/os#File.Fd
+		// [2]: https://go.dev/play/p/9JG7voL2oHP
+		_ = fromWeaveletWriter
```
mwhittaker, line 167: Hm, I'm worried it's still possible for a test to end, these variables get garbage collected, the pipes get closed, and a subprocess weavelet with access to these pipes crashes? I think we have to wait for all subprocesses to die (which we don't have a good way of doing), not wait for the test to end?

What do you think about sticking with the super janky package-scoped `dontGC` for now, and then at some point in the future, when we clean up the code to make sure subprocesses are dying when they're supposed to, we can get rid of it?

mwhittaker (changes requested):
Very impressive @dnephin! You're understanding and fixing some of the nastiest code in the entire repo, and only days after we open-sourced 🤯. I wish I could grok things as fast as you! 

Code is definitely much cleaner now!

weavertest/multi.go
dnephin: I think the scenario you describe is very likely, but I wonder if it's a problem. As long as the test is finished, and nothing tries to write to `t.Log` after the test has finished (which seems to be handled by `logSaver` in this code), does it matter if a process crashes? I believe any logs and errors would be dropped (not sent to stderr). Having the subprocesses exit earlier seems preferable (even if they are exiting with an error). I ran the tests with `-count=100` a few times to see if I could get any tests to fail, but they passed consistently as they do on `main`.

I'm happy to switch back to the package-scoped variable if you prefer that approach. What do you think?

mwhittaker: Let's switch back to the package-scoped variable. I think what I wrote above was wrong. The main weavelet is actually running in this process, not a subprocess, and is the thing that can crash.

dnephin: Ok, sounds fair! I've restored the package-scoped `dontGC` slice in the latest commit. I added a package-scoped `dontGCLock sync.Mutex` variable as well, since we don't have access to the mutex that was used previously used. 

The previous lock was effectively scoped to the test I believe. It was a field on `Babysitter`, which is created for each call to `weavertest.Init`. I did not realize this until just now, but I think the previous implementation had a data race if two tests cases in the same package used `t.Parallel` and both of them called `weavertest.Init` at the start of the test. The package-scoped mutex should guard against that case.

mwhittaker: Welp, yeah the old code was completely busted! I told you the code was janky 🤣. Thanks for finding and fixing.

mwhittaker (approved):
🔥 🔥 🔥

dnephin:
No problem, all done!

Issue #131, opened 2023-03-08
running out of requesting address under performance testing


runvyang:
### Setup

os: macOS Catalina Version 10.15.7 
go version go1.20.1 darwin/amd64
commit: 1bd05fa3de1c2526a6944e9018be6d7753973314
deploy: examples/hello weaver multi deploy weaver.toml
benchmark command: ab -n 100000 -c 8 -k http://10.79.161.90:12345/hello?name=world

### Issue:
around finishing 40000 requests, it would throw errors like "http: proxy error: dial tcp 127.0.0.1:54178: connect: can't assign requested address"

### Investigation

* found the dial is from http proxy that launched by weaver, and default transport does not limit MaxConnsPerHost, connections were massively created when under pressure;

* tried to set transport args MaxConnsPerHost and MaxIdleConnsPerHost  to a high number, tried (100, 100), (150, 100), (100, 150), no longer seeing the same error above, but a new issue come out "panic: net/http: internal error: connCount underflow",  this one seems to come up intermittently,  see https://github.com/golang/go/issues/38172

* **tried the same on a Debian 5.4.143, however the issue didn't reproduce**,  it seemed to be macOS related.

### Questions
* what do you suggest to fix it?

runvyang:
Okay, it seems i can no longer reproduce "panic: net/http: internal error: connCount underflow", probably it was under critical connection resource saturation.

for those who encounters erorr "connect: can't assign requested address", set MaxConnsPerHost and MaxIdleConnsPerHost 
to values like (150, 100) the DefaultTransport  in file internal/proxy/proxy.go should help.

rgrandl:
@runvyang can you take another look at #144 ?

liufuyang:
I encountered the same problem and I seem need to set `MaxIdleConnsPerHost` as 100 to get over the problem.

I left `MaxIdleConns` as 100 as well as it is the default transport made for `httputil.ReverseProxy` is 100 see [here](https://github.com/golang/go/blob/ac85f2bedd405d64b54c3a248c9079b6cde57359/src/net/http/httputil/reverseproxy.go#L158) and [here](https://github.com/golang/go/blob/ac85f2bedd405d64b54c3a248c9079b6cde57359/src/net/http/transport.go#L50). 

@runvyang do you have time to update your PR? Otherwise, I can also come up with a similar fix.

rgrandl:
This is not actionable.

PR #144, opened 2023-03-10
Fix http proxy defaults to avoid running out of requesting address

runvyang:
see issue described in https://github.com/ServiceWeaver/weaver/issues/131

### DIFF
--- a/internal/proxy/proxy_darwin.go
+++ b/internal/proxy/proxy_darwin.go
@@ -12,33 +12,56 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//go:build !darwin
-//+build !darwin
+//go:build darwin
+//+build darwin
 
 package proxy
 
 import (
 	"errors"
 	"math/rand"
+	"net"
 	"net/http"
 	"net/http/httputil"
 	"sync"
+	"time"
 
-	"golang.org/x/exp/slog"
+	"github.com/ServiceWeaver/weaver/internal/logtype"
 )
 
 // Proxy is an HTTP proxy that forwards traffic to a set of backends.
 type Proxy struct {
-	logger   *slog.Logger          // logger
+	logger   logtype.Logger        // logger
 	reverse  httputil.ReverseProxy // underlying proxy
 	mu       sync.Mutex            // guards backends
 	backends []string              // backend addresses
 }
 
 // NewProxy returns a new proxy.
-func NewProxy(logger *slog.Logger) *Proxy {
+func NewProxy(logger logtype.Logger) *Proxy {
 	p := &Proxy{logger: logger}
-	p.reverse = httputil.ReverseProxy{Director: p.director}
+	p.reverse = httputil.ReverseProxy{
+		Director: p.director,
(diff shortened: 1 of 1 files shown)


rgrandl:
Thanks for looking into this @runvyang. 

I tried to reproduce the issue "http: proxy error: dial tcp 127.0.0.1:54178: connect: can't assign requested address" on my Linux machine, but I couldn't reproduce it even for 10 millions requests. 

My guess is that the macOS TCP stack is not able to recycle local port numbers as quickly as Linux. 

Looks like declaring a custom transport and tweaking the `MaxConnsPerHost` and `MaxIdleConnsPerHost` is what people do for benchmarking though. 

However, the default values for `MaxIdleConnsPerHost` is 2 and for for `MaxConnsPertHost` is unlimited. I am wondering which values we should set by default to not degrade the performance but also don't get into these issues. Seems like there is no exact answer. 

Would you mind doing some further benchmarking? Ideally we should modify maybe only one of these 2 values, and the values should be as close as possible to the default ones. For me (100, 150) vs (2, inf) are just magic that might work well for your setup.

rgrandl:
@runvyang, do you mind taking another look at this?

runvyang:
Update:
1. 5 is the minimum value for MaxIdleConnsPerHost
2. also added conditional build so the updated values only apply to darwin 


BTW, sorry for the late response, was able to grab sometime today;

internal/proxy/proxy_darwin.go
```diff
@@ -0,0 +1,93 @@
...
+			MaxIdleConns:          100,
+			IdleConnTimeout:       90 * time.Second,
+			TLSHandshakeTimeout:   10 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+			// for non HTTP/2 cases, set a reasonable limit on the
+			// setup reasonable defaults so this proxy wound not run out
+			// of requesting adddress under high load pressure, e.g examples/hello
+			// default value of MaxIdelConnsPerHost is 2, 5 is the minimum value to
+			// make requesting address issue disappear
+			MaxIdleConnsPerHost: 5,
```
liufuyang, line 61: In my case, I need to set this up to 100, otherwise still see the issue.

internal/proxy/proxy_darwin.go
```diff
@@ -0,0 +1,93 @@
...
+	p.reverse = httputil.ReverseProxy{
+		Director: p.director,
+		Transport: &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
+			DialContext: (&net.Dialer{
+				Timeout:   30 * time.Second,
+				KeepAlive: 30 * time.Second,
+			}).DialContext,
+			ForceAttemptHTTP2:     true,
+			MaxIdleConns:          100,
```
liufuyang, line 52: This looks correct to me as those are all just default parameters.
https://github.com/golang/go/blob/ac85f2bedd405d64b54c3a248c9079b6cde57359/src/net/http/transport.go#L50

rgrandl:
No follow-up from the users on this. Closing it for now.

PR #151, opened 2023-03-11
Enable errcheck linter on production files

### LINKED ISSUE #90: Enabling the errcheck linter

dnephin:
Closes #90

This PR completes the work to enable the `errcheck` linter on non-test files. There are two changes:

1. handle and print errors related to serving the perfetto backend
2. ignore errors from async calls to `conn.send`. `conn.send` captures the error as `conn.failure`, and any subsequent calls to `send` will return the first error. Ignoring the `send` error is these cases should be safe because  because the error will be related a bit later, not completely lost.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -13,10 +13,12 @@
 # limitations under the License.
 
 linters:
-  disable:
-    - errcheck # Currently too many places are ignoring errors
   enable:
     - nolintlint
+issues:
+  exclude-rules:
+    - path: '_test\.go'
+      linters: [errcheck] # Currently too many places are ignoring errors
 linters-settings:
   staticcheck:
     checks:
--- a/runtime/perfetto/db.go
+++ b/runtime/perfetto/db.go
@@ -27,8 +27,10 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/ServiceWeaver/weaver/internal/files"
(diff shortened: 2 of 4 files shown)


runtime/perfetto/db.go
```diff
@@ -513,16 +515,28 @@ func (d *DB) Serve(ctx context.Context) error {
...
 		select {
 		case <-ticker.C:
 			lis, err := net.Listen("tcp", "localhost:9001")
-			if err == nil {
-				ticker.Stop()
-				return server.Serve(lis)
+			if err != nil {
+				warnOnce.Do(func() {
+					fmt.Fprintf(os.Stderr, "Perfetto server failed to listen on port 9001: %v\n", err)
+					fmt.Fprintf(os.Stderr, "Perfetto server will retry until port 9001 is available\n")
```
dnephin, line 526: I'm not sure if "Perfetto server" is the right term, or if this should be "Perfetto backend", or something else. I'm happy to adjust the wording here if anyone has suggestions.

.golangci.yaml
```diff
@@ -13,10 +13,12 @@
 # limitations under the License.
 
 linters:
-  disable:
```
mwhittaker, line 16: :heart_eyes:

runtime/perfetto/db.go
```diff
@@ -513,16 +515,28 @@ func (d *DB) Serve(ctx context.Context) error {
...
+				})
+				continue
 			}
+
+			ticker.Stop()
+			err = server.Serve(lis)
+			if !errors.Is(err, http.ErrServerClosed) {
```
mwhittaker, line 533: Nice, never knew to check this.

runtime/perfetto/db.go
mwhittaker: This sounds great to me.

mwhittaker (approved):
Looks great! :checkered_flag: :heavy_check_mark: :white_check_mark: :ballot_box_with_check:

PR #152, opened 2023-03-11
Remove from runtime/envelope

dnephin:
This PR is a small collection of cleanups in the `runtime/envelope` package for your consideration. There may be good reasons not to make these changes. I'm happy to remove any of the commits in this PR.  

I've made each change in a separate commit, and provided more detail in the commit messages, so may be best viewed by individual commit. Two of the changes are very small, the the larger changes are:

1. Use `type EnvelopeHandler conn.EnvelopeHandler` to remove the duplicate definition of `EnvelopeHandler`.
2. Replace `Options.GetEnvelopeConn` with an exported `SetConn`.

### DIFF
--- a/runtime/envelope/envelope.go
+++ b/runtime/envelope/envelope.go
@@ -388,10 +388,6 @@ func (e *Envelope) copyLines(component string, src io.Reader) error {
 	}
 }
 
-func (e *Envelope) Weavelet() *protos.WeaveletInfo {
-	return e.weavelet
-}
-
 func dropNewline(line []byte) []byte {
 	if len(line) > 0 && line[len(line)-1] == '\n' {
 		line = line[:len(line)-1]
--- a/runtime/envelope/envelope.go
+++ b/runtime/envelope/envelope.go
@@ -112,20 +112,6 @@ type Options struct {
 	// Retry configures the exponential backoff performed when restarting the
 	// weavelet. Defaults to retry.DefaultOptions.
 	Retry retry.Options
-
-	// GetEnvelopeConn returns a connection between a pre-started weavelet and
-	// its Envelope.
-	//
-	// Note that setting this option implies that the weavelet has already been
-	// started, in which case the Envelope will not restart it (i.e., the
-	// Restart policy is assumed to be Never).
-	//
-	// TODO: conn.EnvelopeConn is internal. This is only used right now for
-	// weavertest. If we remove the weavertest use case, we can remove this
-	// option.
-	//
-	// TODO: Set this only in weavertest.
-	GetEnvelopeConn func() *conn.EnvelopeConn
 }
 
 // Envelope starts and manages a weavelet, i.e., an OS process running inside a
@@ -199,16 +185,9 @@ func (e *Envelope) toggleProfiling(expected bool) bool {
 	return true
 }
 
-// Run runs the application, restarting it when necessary.
+// Run runs the application, restarting it according to the policy specified by
+// opts.
 func (e *Envelope) Run(ctx context.Context) error {
-	if e.opts.GetEnvelopeConn != nil {
-		// The weavelet already started. Set the connection between the
-		// weavelet and the envelope.
-		conn := e.opts.GetEnvelopeConn()
-		e.setConn(conn)
-		return ctx.Err()
-	}
-
 	for r := retry.BeginWithOptions(e.opts.Retry); r.Continue(ctx); {
 		err := e.runWeavelet(ctx)
 		if e.isStopped() {
--- a/runtime/envelope/envelope.go
+++ b/runtime/envelope/envelope.go
@@ -374,7 +374,7 @@ func (e *Envelope) ReadMetrics() ([]*metrics.MetricSnapshot, error) {
 	if conn == nil {
 		return nil, fmt.Errorf("cannot read metrics: weavelet pipe is down")
 	}
-	return e.conn.GetMetricsRPC()
+	return conn.GetMetricsRPC()
 }
 
 func (e *Envelope) isStopped() bool {
--- a/weavertest/multi.go
+++ b/weavertest/multi.go
@@ -30,10 +30,8 @@ import (
 	"github.com/ServiceWeaver/weaver/runtime"
 	"github.com/ServiceWeaver/weaver/runtime/codegen"
 	"github.com/ServiceWeaver/weaver/runtime/colors"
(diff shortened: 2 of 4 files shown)


runtime/envelope/envelope.go
```diff
@@ -202,25 +142,18 @@ func (e *Envelope) toggleProfiling(expected bool) bool {
...
-		// weavelet and the envelope.
-		conn := e.opts.GetEnvelopeConn()
-		e.setConn(conn)
-		return ctx.Err()
-	}
-
-	for r := retry.BeginWithOptions(e.opts.Retry); r.Continue(ctx); {
+// Run runs the application, restarting it according to the policy specified by
+// opts.
+func (e *Envelope) Run(ctx context.Context, opts Options) error {
```
dnephin, line 147: The change to move `Options` from a field to an argument of this method isn't strictly necessary. It seemed nice because options were only used by `Run`, and `weavertest` no longer needs to call `Run`, so it is able to exclude options.

If `Options` are likely to be used by other `Envelope` methods, I can restore it as a field.

runtime/envelope/envelope.go
```diff
@@ -412,10 +347,6 @@ func (e *Envelope) copyLines(component string, src io.Reader) error {
 	}
 }
 
-func (e *Envelope) Weavelet() *protos.WeaveletInfo {
-	return e.weavelet
-}
```
dnephin, line 417: This change will require a patch to weaver-gke, which I've prepared in a branch: https://github.com/ServiceWeaver/weaver-gke/compare/main...dnephin:weaver-gke:remove-envelope-weavlet-method

I can open a PR for that change if this change is accepted.

dnephin: Ah, maybe not! https://github.com/ServiceWeaver/weaver-gke/pull/31 removes this call already.

runtime/envelope/envelope.go
```diff
@@ -35,55 +35,11 @@ import (
...
-	// RecvTraceSpans enables the envelope to receive a sequence of trace spans.
-	RecvTraceSpans(spans []trace.ReadOnlySpan) error
-}
-
-// Ensure that EnvelopeHandler remains in-sync with conn.EnvelopeHandler.
-var (
-	_ EnvelopeHandler      = conn.EnvelopeHandler(nil)
-	_ conn.EnvelopeHandler = EnvelopeHandler(nil)
-)
+type EnvelopeHandler conn.EnvelopeHandler
```
mwhittaker, line 42: I think the reason we copied the interface here is to avoid exposing `conn.EnvelopeHandler` in the public API since it's an internal package. And doing the opposite, putting `type EnvelopeHandler = envelope.EnvelopeHandler` in the `conn` package didn't work because of cyclic dependencies. @spetrovic77 what do you think?

runtime/envelope/envelope.go
```diff
@@ -377,7 +312,7 @@ func (e *Envelope) ReadMetrics() ([]*metrics.MetricSnapshot, error) {
 	if conn == nil {
 		return nil, fmt.Errorf("cannot read metrics: weavelet pipe is down")
 	}
-	return e.conn.GetMetricsRPC()
+	return conn.GetMetricsRPC()
```
mwhittaker, line 315: Great find! :detective:

runtime/envelope/envelope.go
```diff
@@ -412,10 +347,6 @@ func (e *Envelope) copyLines(component string, src io.Reader) error {
 	}
 }
 
-func (e *Envelope) Weavelet() *protos.WeaveletInfo {
```
mwhittaker, line 415: Nice!

weavertest/multi.go
```diff
@@ -116,26 +115,17 @@ func initMultiProcess(ctx context.Context, t testing.TB, config string) weaver.I
 	}
 
 	// Create and run an envelope for the main process.
-	options := envelope.Options{
-		Restart:         envelope.Never,
-		Retry:           retry.DefaultOptions,
-		GetEnvelopeConn: func() *conn.EnvelopeConn { return econn },
-	}
-	e, err := envelope.NewEnvelope(weaveletInfo, dep.App, b, options)
+	e, err := envelope.NewEnvelope(weaveletInfo, dep.App, b, envelope.Options{})
```
mwhittaker, line 118: Hm, maybe I'm missing something, but what if we don't make an envelope here at all? Then, we can unexport the `SetConn` method again while keeping the `GetEnvelopeConn` option deleted?

runtime/envelope/envelope.go
mwhittaker: SGTM! :+1:

mwhittaker (commented):
Lovely change @dnephin!

runtime/envelope/envelope.go
dnephin: I thought that might be the reason. I started to explore another approach, moving `EnvelopeConn` into the `runtime/envelope` package. That would address the problem because both consumers of this interface would be in the same package. That can be seen here:  https://github.com/dnephin/weaver/compare/envelope-handler-interface...dnephin:weaver:envelope-handler-interface-2

With a bit more work I think it would be possible to unexport `EnvelopeConn` and solve the `weavertest` case by accepting the pipes to `Envelope`. I realized that would almost reserve the other change I had made here to remove `GetEnvelopeConn`, so I stopped that approach, but it's another option to consider.

I can revert this part of the change if it's not a good time to try and address the duplicate interfaces.

weavertest/multi.go
dnephin: haha, oh ya! I missed that after this change `Envelope` is doing nothing! Removing it entirely seems to work, I'll push a commit that makes that change now.

runtime/envelope/envelope.go
mwhittaker: Yeah, maybe let's revert for now so we can get the other awesome changes in this PR merged. With the type assertions below, I feel like the duplication is ugly but at least not bug-prone. 

```go
var (
	_ EnvelopeHandler      = conn.EnvelopeHandler(nil)
	_ conn.EnvelopeHandler = EnvelopeHandler(nil)
)
```

dnephin: Sounds good! I've removed the commit with this change, rebased on latest main, and squashed some of the intermediate commits.

mwhittaker (approved):

PR #155, opened 2023-03-13
Remove the circular reference between ssh/impl.babysitter and Envelope

dnephin:
Thank you for all the prompt reviews on my pull requests! This one is a few small changes I noticed while reading over the code.

The first commit simplifies the relationship between `internal/tool/ssh/impl.babysitter` and `runtime/envelope.Envelope`. Currently `babysitter` has a reference to `Envelope` and is passed to Envelope as the implementation of `EnvelopeHandler`. In this commit I extracted the metrics collection to a new `metricsCollector` type. The result is `babysitter` becomes only the implementation of `EnvelopeHandler`, and `Envelope` is passed to the `metricsCollector` so that it can perform operations.

I've found that often code can be easier to read when there aren't circular references between types. My hope is that this change makes the `ssh` deployer implementation a bit easier to read.

**Before**

```mermaid
flowchart LR
  Run --> babysitter --> Envelope --> babysitter
  Run --> Envelope
```

**After**
```mermaid
flowchart LR
  Run --> Envelope --> babysitter
  Run --> metricsCollector --> Envelope
```

The second commit fixes some typos and removes some unused code in `generate`.

### DIFF
--- a/internal/tool/ssh/impl/babysitter.go
+++ b/internal/tool/ssh/impl/babysitter.go
@@ -40,17 +40,15 @@ type babysitter struct {
 	ctx           context.Context
 	opts          envelope.Options
 	dep           *protos.Deployment
-	group         *protos.ColocationGroup
-	replicaId     int32
 	mgrAddr       string
 	logger        logtype.Logger
 	traceExporter *traceio.Writer // to export traces to the manager
-	envelope      *envelope.Envelope
 }
 
 var _ envelope.EnvelopeHandler = &babysitter{}
 
-// RunBabysitter creates and runs a babysitter that was deployed using SSH.
+// RunBabysitter creates and runs an envelope.Envelope and a metrics collector for a
+// weavelet deployed with SSH.
 func RunBabysitter(ctx context.Context) error {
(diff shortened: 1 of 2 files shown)


mwhittaker (approved):
Nice change! Also your mermaid figure is blowing my mind :exploding_head:. I had no idea you could do something like that.

internal/tool/ssh/impl/babysitter.go
```diff
@@ -108,12 +104,18 @@ func RunBabysitter(ctx context.Context) error {
...
 }
 
-func (b *babysitter) collectMetrics() {
+type metricsCollector struct {
+	logger   logtype.Logger
+	envelope *envelope.Envelope
+	info     *BabysitterInfo
+}
+
+func (b *metricsCollector) run(ctx context.Context) {
```
mwhittaker, line 118: Oops, forgot to mention we should probably rename this `b` to an `m`.

dnephin: Sounds good, I'll open a PR with this change.

PR #156, opened 2023-03-13
Fixed UseLocalhost bug introduced in PR #115.

### LINKED ISSUE #154: tutorial wont multi deploy - cant resolve local hostname
this feels like an env variable problem, but its not clear which var or how to override.

when following the tutorial, i get to the `multi deploy` phase and i get this output when trying to hit the endpoint for the first time:

> S0313 11:24:35.338221 stderr  c17304c8           ] error initializing Service Weaver: error creating internal listener: listen tcp: lookup HOSTNAME.COMPANY.com: no such host

(hostname PII redacted) 

it seems like it is trying to resolve my FQDN, but should really just be using localhost or the loopback IP.

FWIW, single deploy is working fine:

```
hello listener available on 127.0.0.1:12345
╭───────────────────────────────────────────────────╮
│ app        : hello                                │
│ deployment : ddde8609-00e3-4074-a41c-60f6fa3e4275 │
╰───────────────────────────────────────────────────╯
```

```
curl "localhost:12345/hello?name=Weaver"
Hello, revaeW!
```

mwhittaker:
Before PR #115, the `WeaveletInfo` proto had a `UseLocalhost` field. When true, weavelets would listen on localhost for external listeners (the kind returned by the `Instance.Listener` method). In PR #115, I removed the `UseLocalhost` field and replaced it with a `GetAddress` API.

However, I didn't actually remove the field, and the field was being used for more than I initially thought. The `UseLocalhost` field was still in `WeaveletInfo` and was being used for internal listeners (the kind that a weavelet uses to execute RPCs). The PR changed 'weaver multi' to not set the `UseLocalhost` field which led to weavelets listening on `$HOSTNAME` rather than localhost.

This PR fixes the bug. I replace `UseLocalhost` with `SingleMachine` since it parallels the existing `SingleProcess` field better and I think it's a bit clearer (use localhost for what?).

I also removed some related unused fields from the `WeaveletInfo` proto.

Fixes #154.


internal/tool/ssh/impl/babysitter.go
```diff
@@ -99,6 +99,7 @@ func RunBabysitter(ctx context.Context) error {
 		SameProcess:   b.dep.App.SameProcess,
 		Sections:      b.dep.App.Sections,
 		SingleProcess: b.dep.SingleProcess,
+		SingleMachine: false,
```
rgrandl, line 102: this should be set by default to false, maybe no need to set it explicitly.

rgrandl (approved):
Thanks Michael. 

Just a note: It would be nice if we can somehow get rid of `SingleProcess`. We have now things like `SingleProcess`, `SameProcess` and `SingleMachine` and different combinations of these params for different deployers, which can make it challenging for a deployer developer to set them as expected.

internal/tool/ssh/impl/babysitter.go
mwhittaker: Done.

mwhittaker:
> Thanks Michael.
> 
> Just a note: It would be nice if we can somehow get rid of `SingleProcess`. We have now things like `SingleProcess`, `SameProcess` and `SingleMachine` and different combinations of these params for different deployers, which can make it challenging for a deployer developer to set them as expected.

Great point. I'll work on this in a future CL. I think we may be able to remove `SingleProcess`. And `SameProcess` should maybe be renamed `colocated` or something like that.

PR #157, opened 2023-03-13
Rename method receiver and small improvements to routelet

dnephin:
The first commit removes the `env` field from `routelet`, and updates `routelet.watchRoutingInfo` to use the `env` argument. Previously it was using the argument for one call, and the field for two others. Now it uses the arg consistently for all 3.

The second commit removes an unused `error` return from `weavelet.onNewRoutingInfo` (found using the `unused` linter), which simpifies the caller by removing the need to log an error.

The third commit addresses this comment (https://github.com/ServiceWeaver/weaver/pull/155#discussion_r1134600757) by fixing the method receiver name.


routing.go
```diff
@@ -51,7 +49,7 @@ type routelet struct {
 // lifetime of the routelet is bound to the provided context. When the context is
 // cancelled, the routelet stops tracking the latest routing information.
 func newRoutelet(ctx context.Context, env env, group string) *routelet {
-	r := &routelet{env: env, balancers: map[string]*routingBalancer{}}
+	r := &routelet{balancers: map[string]*routingBalancer{}}
 	go r.watchRoutingInfo(ctx, env, group)
```
dnephin, line 53: I've noticed there are a couple places where goroutines are started in the "constructor

dnephin, line 53: I've noticed there are a couple places where goroutines are started in the `new` function for the type. I usually try to make concurrency as explicit as possible and have the caller start the goroutine by calling a `run` function (ex `go routelet.Run(ctx)`). That pattern requires the caller to do more work, but I think generally makes it easier to wait for goroutines to shutdown, because the caller can manage them using `sync.WaitGroup` or` errgroup.Group`.

I'm wondering if that pattern might help us solve the TODO below, or if it's not worth the change.

routing.go
```diff
@@ -114,13 +112,13 @@ func (r *routelet) watchRoutingInfo(ctx context.Context, env env, group string)
 		routingInfo, newVersion, err := env.GetRoutingInfo(ctx, group, version)
 		if err != nil {
 			// TODO(mwhittaker): Handle errors more gracefully.
```
dnephin, line 114: I was considering a fix for this, but I'm not sure how the errors should be handled. Should an error here cause `watchRoutingInfo` to exit? Or is there some other way to handle these?

routing.go
mwhittaker: Sounds like a great idea to me! Feel free to do in this PR or in a separate PR, whatever you prefer.

routing.go
mwhittaker: Hmm, yeah this is tricky. I don't think an error should cause `watchRoutingInfo` to exit. The RPC to get routing info may experience a network error, for example, that should be retried. But, I also don't know what to do besides just logging and retrying. We should probably---here and everywhere else in the code---be more careful about which errors are retriable and which aren't. Maybe just leave it is as is for now since I think it's a more complicated fix.

routing.go
```diff
@@ -38,8 +38,6 @@ import (
 // routelet maintains the latest routing information for a process by
 // communicating with the manager.
 type routelet struct {
-	env env
```
mwhittaker, line 41: Nice!

weavelet.go
```diff
@@ -235,11 +235,7 @@ func (d *weavelet) start() (Instance, error) {
 
 		// Monitor our routing assignment.
 		routelet := newRoutelet(d.ctx, d.env, d.info.Group.Name)
-		routelet.onChange(func(info *protos.RoutingInfo) {
-			if err := d.onNewRoutingInfo(info); err != nil {
-				fmt.Fprintln(os.Stderr, err)
-			}
-		})
+		routelet.onChange(d.onNewRoutingInfo)
```
mwhittaker, line 238: Niiiice!

mwhittaker (approved):
Great change!

PR #160, opened 2023-03-14
Refactor example/collatz/colocated.toml right colocate

kiripeng214:


mwhittaker (approved):
Nice fix! Thanks so much.

PR #166, opened 2023-03-15
Update the website with the new GKE instructions.

spetrovic77:
These new instructions work on other platforms, in addition to Linux.


mwhittaker (approved):
:shipit:

PR #167, opened 2023-03-15
Website copy button only copies commands, if present.

spetrovic77:
Otherwise, it copies the entire text.

This fixes the reported issue where the commands were copied with "$" prefix, and as a result can't be just pasted into the terminal.

### DIFF
--- a/website/assets/js/copy.js
+++ b/website/assets/js/copy.js
@@ -30,18 +30,39 @@ function addCopyButtonsToPres() {
     button.innerText = '📋';
     button.classList.add('pre-copy-button');
     button.addEventListener('click', function() {
-      if (navigator.clipboard) {
-        // Due to how the markdown renderer converts markdown to HTML, every
-        // newline in innerText is duplicated. The replaceAll removes the
-        // redundant newlines.
-        navigator.clipboard.writeText(code.innerText.replaceAll('\n\n', '\n'));
-      }
+        if (navigator.clipboard) {
+            // Due to how the markdown renderer converts markdown to HTML, every
+            // newline in innerText is duplicated. The replaceAll removes the
+            // redundant newlines.
+            let text = code.innerText.replaceAll('\n\n', '\n');
+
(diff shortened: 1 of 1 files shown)


website/assets/js/copy.js
```diff
@@ -44,4 +52,18 @@ function addCopyButtonsToPres() {
...
 
+// getCommands splits the provided text into lines and returns the subset
+// of lines that represent commands, i.e., the lines that start with "$ ".
+function getCommands(text) {
+    let lines = text.split("\n");
+    let commands = Array()
+    for (let i = 0; i < lines.length; i++) {
+        let line = lines[i]
+        if (line.startsWith("$ ")) {
+            commands.push(line.substr(2))
```
mwhittaker, line 63: I think `substr` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr).  Replace with `substring`?

website/assets/js/copy.js
```diff
@@ -44,4 +52,18 @@ function addCopyButtonsToPres() {
   }
 }
 
+// getCommands splits the provided text into lines and returns the subset
+// of lines that represent commands, i.e., the lines that start with "$ ".
+function getCommands(text) {
+    let lines = text.split("\n");
+    let commands = Array()
```
mwhittaker, line 59: `let commands = [];`?

website/assets/js/copy.js
```diff
@@ -44,4 +52,18 @@ function addCopyButtonsToPres() {
...
+    for (let i = 0; i < lines.length; i++) {
```
mwhittaker, line 60: `for (let line of lines) {`?

website/assets/js/copy.js
mwhittaker, line 1: Mind adding semicolons to the ends of lines? My understanding is that they're optional but highly recommended.

mwhittaker (approved):
Nice fix! Thanks Srdjan!

PR #168, opened 2023-03-16
docs: add homebrew installation note

chenrui333:
- Add a homebrew installation note.


mwhittaker:
Hi @chenrui333, what are the advantages of installing Service Weaver with homebrew instead of installing it with `go install`?

mwhittaker:
I'm closing this PR because it's quite stale, but feel free to re-open.

Issue #127, opened 2023-03-07
multi deploy: not supported by windows


lemon-1997:
when I executed `weaver multi deploy weaver.toml`, it doesn't work.
config
```
[serviceweaver]
name = "weaver"
binary = "./weaver.exe"
```
temp/serviceweaver/logs/weaver-multiweaver..f1de7a14-6cfb-49fb-9f63-1fec84875a3b.error.log
```
,  
weaver
babysitter"$f1de7a14-6cfb-49fb-9f63-1fec84875a3b)#9N? 2error:dC:/Users/lemon/go/pkg/mod/github.com/!service!weaver/weaver@v0.1.1/internal/babysitter/babysitter.go@?Je.RunRserviceweaver/systemR RerrRSfork/exec C:\Users\lemon\GolandProjects\weaver\weaver.exe: not supported by windows,  
weaver
babysitter"$f1de7a14-6cfb-49fb-9f63-1fec84875a3b)#9N? 2error:dC:/Users/lemon/go/pkg/mod/github.com/!service!weaver/weaver@v0.1.1/internal/babysitter/babysitter.go@?Je.RunRserviceweaver/systemR RerrRSfork/exec C:\Users\lemon\GolandProjects\weaver\weaver.exe: not supported by windows
```

is weaver multiprocess does not support windows?

charles-chenzz:
it's a bug and the team fixing it I guess. In the mean time you can change the name and delete files on ``/tmp/serviceweaver/logs/weaver-multi`` to try again to see if it help

lemon-1997:
> it's a bug and the team fixing it I guess. In the mean time you can change the name and delete files on `/tmp/serviceweaver/logs/weaver-multi` to try again to see if it help

delete files doesn't work. i think  the problem is ExtraFiles is not supported on Windows.
```go
// ExtraFiles specifies additional open files to be inherited by the
// new process. It does not include standard input, standard output, or
// standard error. If non-nil, entry i becomes file descriptor 3+i.
//
// ExtraFiles is not supported on Windows.
ExtraFiles []*[os](https://pkg.go.dev/os).[File](https://pkg.go.dev/os#File)
```

rgrandl:
Duplicate of #81

PR #169, opened 2023-03-16
fix multi deploy on Windows

lemon-1997:
This PR fixes a problem with multi-process deployment on Windows. The windows inherit pipe uses handles, cmd.ExtraFiles is not supported on Windows. See issue https://github.com/ServiceWeaver/weaver/issues/127

### DIFF
--- a/internal/pipe/pipe.go
+++ b/internal/pipe/pipe.go
@@ -48,7 +48,7 @@ func CommandContext(ctx context.Context, name string, arg ...string) *Cmd {
 // idiomatic usage.
 //
 // [1]: https://pkg.go.dev/os/exec#example-Cmd.StdoutPipe
-func (c *Cmd) RPipe() (int, io.ReadCloser, error) {
+func (c *Cmd) RPipe() (uintptr, io.ReadCloser, error) {
 	if c.Process != nil {
 		return 0, nil, fmt.Errorf("capture: RPipe after process started")
 	}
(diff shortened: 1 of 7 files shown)


internal/pipe/pipe.go
```diff
@@ -86,11 +88,21 @@ func (c *Cmd) WPipe() (int, io.WriteCloser, error) {
...
-	c.Cmd.ExtraFiles = append(c.Cmd.ExtraFiles, remote)
+func (c *Cmd) registerPipe(local, remote *os.File) uintptr {
+	var fd uintptr
+	// Cmd.ExtraFiles is not supported on Windows.
+	if runtime.GOOS == "windows" {
+		fd = remote.Fd()
+		if c.Cmd.SysProcAttr == nil {
+			c.Cmd.SysProcAttr = &syscall.SysProcAttr{}
+		}
+		c.Cmd.SysProcAttr.AdditionalInheritedHandles = append(c.Cmd.SysProcAttr.AdditionalInheritedHandles, syscall.Handle(fd))
```
spetrovic77, line 99: It looks like this doesn't compile, at least with the latest Go versions and on Linux. Could you confirm that it does compile on Windows?

If it does compile on Windows, would adding a separate pipe_windows.go help here?

spetrovic77 (commented):
Thanks so much for this change. Overall, it looks great, modulo the compilation issue. PTAL

lemon-1997:
Hi @spetrovic77 ,now it compiles on windows and Linux

PR #162, opened 2023-03-14
Wrote "How to Implement a Deployer" blog.

mwhittaker:
In this PR, I write a "How to Implement a Deployer" blog post that describes what deployers are and how to write one. The blog post includes a fully working multiprocess deployer, a fully working single process deployer, and a trivial deployer that interacts with pipes directly. Writing this blog post also induced us to make [a number of improvements to the deployer API](https://github.com/ServiceWeaver/weaver/pulls?q=is%3Apr+is%3Aclosed+label%3Adeployer-api). Hopefully, this blog post will encourage community contributed deployer implementations!

PR #172, opened 2023-03-16
Cleaned up the multi/weavertest babysitter.

mwhittaker:
This PR cleans up the babysitter used by weaver multi and weavertest. The changes don't impact the behavior of the babysitter; they simply clean up the code. I made these changes in preparation for some bigger changes I'm making in an effort to simplify the deployer API.

I made the following changes:

- I replaced the babysitter's use of `versioned.Map` with a new `versioned.Versioned` abstraction (first introduced in #162). Previously, a babysitter was storing a single piece of state in a map, which was a bit confusing. The new `Versioned` abstraction allows for the next change.

- I introduced a `group` struct to store all the information about a group. Previously, the state was spread across a couple different fields, and in some `versioned.Map`s.

- I removed `babysitter.proto`. I realized that there was no need to be using protos. Now, everything is vanilla Go.

- I moved the code in `profile.go` to `babysitter.go`, so the babysitter is now a single file.

Ironically, a lot of the changes introduced in this PR will be deleted in future PRs as we transition to the new deployer API, but these changes will help that transition.


spetrovic77 (approved):
Approving for now, though as we discussed offline, the sync properties around versioned.Versioned seem extremely subtle. Since this versioning may be going away, maybe it's okay in the interim.

mwhittaker:
> Approving for now, though as we discussed offline, the sync properties around versioned.Versioned seem extremely subtle. Since this versioning may be going away, maybe it's okay in the interim.

Talked more offline, and Srdjan is going to follow up this CL with an improvement on the `Versioned` API.

PR #173, opened 2023-03-17
Implemented weavertest multiprocess deployer.

mwhittaker:
Before this PR, `weaver multi` and weavertest shared the same babysitter. Both `weaver multi` and weavertest deployed applications across multiple processes, so it made sense not to duplicate the code.

However, while we weren't duplicating code, we were also muddying the abstractions of the babysitter, as everything had to work for both `weaver multi` and weavertest. A good example of this is how we dealt with pipes. For a multiprocess weavertest, the main component runs in the same process as the babysitter. The babysitter doesn't expect this, so weavertest created its own `EnvelopeConn` and used a babysitter as an `EnvelopeHandler`. This worked but it wasn't easy to understand why. It really broke the abstraction of the babysitter.

This PR implements a new weavertest multiprocess deployer. The deployer does duplicate *some* code from the old babysitter, but much of the babysitter's implementation wasn't needed (e.g., code for handling metrics, routing, tracing). The new weavertest deployer has a number of improvements:

- Cleaner handling of pipes. Now, the deployer is very much aware that the main component is running in the same process. I also reworked the Bootstrap API to avoid a particularly nasty garbage collection bug we were facing with pipes.
- Better teardown. The `weaver multi` babysitter doesn't expect to get shut down, but a weavertest deployer does. The new deployer is more careful about killing all subprocesses and stopping all goroutines when a unit test ends.
- We're in the process of cleaning up the deployer API. As I was implementing some of these changes, I realized they made it very hard to share code between `weaver multi` and weavertest. This was again because the abstractions were being tortured to work across the two.
- Enables future testing innovations. We've been toying with the idea of implementing more sophisticated testing. For example, we may randomly fail a component. Now that weavertest has its own deployer, these changes will be much easier to make.

### DIFF
--- a/weavertest/deployer.go
+++ b/weavertest/deployer.go
@@ -1,0 +1,367 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package weavertest
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/ServiceWeaver/weaver"
+	"github.com/ServiceWeaver/weaver/internal/envelope/conn"
+	"github.com/ServiceWeaver/weaver/internal/logtype"
+	"github.com/ServiceWeaver/weaver/internal/versioned"
+	"github.com/ServiceWeaver/weaver/runtime"
+	"github.com/ServiceWeaver/weaver/runtime/colors"
+	"github.com/ServiceWeaver/weaver/runtime/envelope"
+	"github.com/ServiceWeaver/weaver/runtime/logging"
+	"github.com/ServiceWeaver/weaver/runtime/protomsg"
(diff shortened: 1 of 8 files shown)


mwhittaker:
@spetrovic77 I'm fixing some bugs I found. I'll add you back as a reviewer when I fix them!

weavertest/deployer.go
```diff
@@ -310,9 +310,11 @@ func (d *deployer) startGroup(group *group) error {
 		group.envelopes = append(group.envelopes, e)
 
 		// Run the envelope.
-		d.stopped.Add(1)
+		//
+		// TODO(mwhittaker): We should add 'd.stopped.Add(1)' and 'defer
```
mwhittaker, line 314: @spetrovic77, I ran into this bug. The changes you're making to remove restarts from `Envelope` might make it easier to fix this bug. If you're interested in fixing, go ahead! Otherwise, I can pick it up :)

weavertest/deployer.go
```diff
@@ -0,0 +1,376 @@
...
+	return &protos.ExportListenerReply{}, nil
+}
+
+// RegisterReplica implements the envelope.EnvelopeHandler interface.
+func (d *deployer) RegisterReplica(req *protos.ReplicaToRegister) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	group := d.group(req.Group)
+	routing, _ := group.routing.Get("")
```
spetrovic77, line 254: This conflicts with my versioned.Versioned change, but I can merge it on top.

weavertest/deployer.go
spetrovic77: Ack.

spetrovic77 (approved):
Looks great!

weavertest/deployer.go
mwhittaker: We chatted offline, and I'm rebasing on top of the `versioned.Versioned` change.

PR #175, opened 2023-03-20
Fix some typos. When we did the big renaming, some identifiers were n…

rgrandl:
…ot updated as well.


mwhittaker (approved):
Wait, I thought it was spelled "dweavelet". Just kidding :laughing:. Thanks for the change Robert!

PR #174, opened 2023-03-19
Implement the versioned.Versioned type with locks.

spetrovic77:
Joint work with mwhittaker@.

Other changes:
  * Converted babysitter to finer-grained locking, to remove the possible lock inversion w.r.t. b.mu.

PR #176, opened 2023-03-20
Cleaned up SSH deployer's manager.

mwhittaker:
This PR cleans up the SSH deployer's manager. Specifically, I ported the changes we made to the `weaver multi` babysitter in #172 and #174. These cleanups will make it easier to make some of the deployer API changes we've been working on.

I noticed that the SSH babysitter re-uses a lot of the same code as the `weaver multi` babysitter. I don't think we should merge the two into one mega babysitter. In fact, I recently split the `weaver multi` babysitter from the weavertest babysitter, so I think there's value in having things separate. But, maybe we can pull out some of the common code without having a frankenstein babysitter API.


internal/tool/ssh/impl/manager.go
```diff
@@ -448,57 +468,69 @@ func (m *manager) exportListener(_ context.Context, req *protos.ExportListenerRe
...
-	return m.startColocationGroup(ctx, &protos.ColocationGroup{Name: req.ColocationGroup})
+	// Start the colocation group, if it hasn't already started.
+	return m.startColocationGroup(g)
 }
 
-func (m *manager) startColocationGroup(_ context.Context, group *protos.ColocationGroup) error {
-	// If the group is already started, ignore.
-	if _, found := m.started[group.Name]; found {
+func (m *manager) startColocationGroup(g *group) error {
+	g.mu.Lock()
```
spetrovic77, line 508: maybe no need to hold the lock through starting below, i.e.,:

```
shouldStart := func() {
  g.mu.Lock()
  defer g.mu.Unlock()
  if g.started {
    return false
  }
  g.started = true
  return true
}
if !shouldStart() {
  return nil
}
```

It is a bit more verbose, so totallyh up to you.

spetrovic77 (approved):

PR #177, opened 2023-03-20
Terminate the multi deployment when the weavelet fails with error.

spetrovic77:
Part of a series of changes where the envelope no longer restarts the weavelet automatically.


mwhittaker (approved):
LGTM

PR #178, opened 2023-03-20
Don't restart weavelets in the envelope.

spetrovic77:
Related code cleanup.


mwhittaker (approved):
Awesome, thanks Srdjan!

PR #181, opened 2023-03-21
Removed co-location groups from bulk of pipe API.

mwhittaker:
Recall that, before this PR, the deployer API had four methods related to weavelets and components: `RegisterReplica`, `StartComponent`, `GetComponentsToStart`, and `GetRoutingInfo`. The notion of a co-location group was baked into these APIs. A weavelet would ask for a group's routing info, not a component's, for example.

This PR removes the idea of a co-location group from the APIs and makes them component-centric instead. This is desirable because co-location groups are implementation details and should not be forced into the API. Deployers are still free to use the notion of a co-location group internally---and right now, all our deployers do---but they do not have to.

Note that this change does not eliminate co-location groups from the API entirely. There are still some message that reference groups, and the weavelet still uses the idea of a group. Future changes to the pipe API should eliminate these entirely.


internal/babysitter/babysitter.go
```diff
@@ -232,85 +292,87 @@ func (b *Babysitter) StartComponent(req *protos.ComponentToStart) error {
...
-	for i, assignment := range g.routing.Val.Assignments {
-		g.routing.Val.Assignments[i] = routingAlgo(assignment, g.routing.Val.Replicas)
-	}
-
-	// Add pid.
 	g.mu.Lock()
 	defer g.mu.Unlock()
+
+	// Update addresses and pids.
+	if g.addresses[req.Address] {
```
spetrovic77, line 322: Ditto. Let's try avoid double-locking, and update the addresses/pids here, release the lock, and then update routing.

weavelet.go
```diff
@@ -60,8 +60,15 @@ type weavelet struct {
 	componentsByName map[string]*component       // component name -> component
 	componentsByType map[reflect.Type]*component // component type -> component
 
+	// TODO(mwhittaker): We have one client for every component. Every client
```
spetrovic77, line 63: True that!

spetrovic77 (commented):
Looks great, modulo the two-locking issue in the babysitter.

(Note that this problem really started after we changed the versioned.Versioned to acquire locks!!)

internal/babysitter/babysitter.go
```diff
@@ -148,22 +179,41 @@ func (b *Babysitter) RegisterStatusPages(mux *http.ServeMux) {
...
+			routings:   map[string]*versioned.Versioned[*protos.RoutingInfo]{},
 		}
 		b.groups[name] = g
 	}
 	return g
 }
 
+// routing returns the RoutingInfo for the provided component.
+//
+// REQUIRES: g.mu is held.
```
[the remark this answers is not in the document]
mwhittaker: Done.

weavelet.go
mwhittaker: I think I have a not-too-ugly way of fixing this that I'll do in a future PR.

internal/babysitter/babysitter.go
mwhittaker: Done.

internal/babysitter/babysitter.go
```diff
@@ -346,14 +408,18 @@ func (b *Babysitter) ExportListener(req *protos.ExportListenerRequest) (*protos.
...
+// GetRoutingInfo implements the envelope.EnvelopeHandler interface.
 func (b *Babysitter) GetRoutingInfo(req *protos.GetRoutingInfo) (*protos.RoutingInfo, error) {
-	g := b.group(req.Group)
-	version := g.routing.RLock(req.Version)
-	defer g.routing.RUnlock()
-	routing := protomsg.Clone(g.routing.Val)
-	routing.Version = version
-	return routing, nil
+	g := b.group(req.Component)
+	g.mu.Lock()
```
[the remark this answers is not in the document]
mwhittaker: Done.

mwhittaker:
Thanks for the thorough review Srdjan! I made the suggested fixes in a separate commit, if you want to look at the changes only: c68d63365719a99e85a6b826954a74cdd8cfd490.

spetrovic77 (approved):
Looks great, thanks.

PR #182, opened 2023-03-21
Change load report from push to pull

rgrandl:


mwhittaker (approved):
Nice, code looks much simpler now!

PR #183, opened 2023-03-22
Fixed subtle potential EnvelopeConn deadlock.

mwhittaker:
###  The Bug

Recall that an envelope and a weavelet issue RPCs to one another via a pair of pipes. Before this PR, the envelope's main loop looked something like this:

```go
for {
    msg := &protos.WeaveletMsg{}
    e.conn.recv(msg)
    e.handleMessage(msg)
}
```

An envelope receives both RPC requests and RPC responses from the weavelet. `conn.recv` handles the RPC responses internally but returns all RPC requests, which are then handled by `handleMessage`.

This code is susceptible to a subtle deadlock. Consider what happens when, in the process of handling an RPC *from* the weavelet, `handleMessage` issues an RPC *to* the weavelet. `handleMessage` sends an RPC request to the weavelet and then blocks waiting for a response. The weavelet receives and responds to the request. Because `handleMessage` is blocking, `conn.recv` is never called, but `conn.recv` needs to be called to wake up `handleMessage`.

Currently, our code does not deadlock because none of our deployers issue RPCs to a weavelet when processing an RPC from a weavelet. However, [our pending deployer API changes][1] will make this communication pattern very common. When an envelope receives a request to start a component for example, the envelope will issue `UpdateComponent` RPCs to all the weavelets that should start the component.

### A Broken Fix

My first thought to fix the bug was to run every invocation of `handleMessage` in its own goroutine. This is similar to how our RPC server is implemented (see [`internal/net/call/call.go`][2]). However, the envelope is required to process RPCs from the weavelet in order. Among many reasons, this strict ordering guarantee ensures that logs are reported chronologically.

### This PR's Fix

This PR fixes the bug by instead running one goroutine to receive messages and another goroutine to handle them. Messages are exchanged between the two goroutines using a channel to preserve their order.

### A Proposed Fix

This PR works (I think), but I think the code is now a bit complicated. I think we may be able to simplify things if we used four pipes instead of two. The four pipes will be divided into two pairs. One pair will be used for RPCs from the envelope to the weavelet, and the other pair will be used for RPCs from the weavelet to the envelope.

[1]: https://github.com/ServiceWeaver/weaver/pull/162
[2]: https://github.com/ServiceWeaver/weaver/blob/dbe6df4ba8c063a18d41c5f51f9b70f47af67abc/internal/net/call/call.go#L716

### DIFF
--- a/internal/envelope/conn/envelope_conn.go
+++ b/internal/envelope/conn/envelope_conn.go
@@ -23,6 +23,7 @@ import (
 	"github.com/ServiceWeaver/weaver/runtime/protomsg"
 	"github.com/ServiceWeaver/weaver/runtime/protos"
 	"go.opentelemetry.io/otel/sdk/trace"
+	"golang.org/x/sync/errgroup"
 )
 
 // EnvelopeHandler implements the envelope side processing of messages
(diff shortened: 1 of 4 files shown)


internal/envelope/conn/envelope_conn.go
```diff
@@ -87,15 +88,65 @@ func NewEnvelopeConn(r io.ReadCloser, w io.WriteCloser, h EnvelopeHandler, weave
...
-	for {
-		if err := e.conn.recv(msg); err != nil {
-			return err
+	var group errgroup.Group
+	msgs := make(chan *protos.WeaveletMsg, 100)
+
+	// Spawn a goroutine that repeatedly reads messages from the pipe. A
+	// received message is either an RPC response or an RPC request. conn.recv
+	// handles RPC responses internally but returns all RPC requests. This
+	// goroutine forwards these requests to the goroutine spawned below for
```
spetrovic77, line 97: `The reading goroutine forwards those RPC requests to to goroutine ...`

spetrovic77 (approved):

internal/envelope/conn/envelope_conn.go
mwhittaker: Done.

PR #186, opened 2023-03-23
Fix the babysitter deadlock caused by double-locking.

spetrovic77:
This breakage was introduced in my PR:
  https://github.com/ServiceWeaver/weaver/pull/184

### DIFF
--- a/internal/babysitter/babysitter.go
+++ b/internal/babysitter/babysitter.go
@@ -105,6 +105,7 @@ type group struct {
 
 	// guards the following data structures, but not their contents.
 	mu        sync.Mutex
+	started   bool                                                 // Has the group been started?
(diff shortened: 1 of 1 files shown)


internal/babysitter/babysitter.go
```diff
@@ -169,6 +171,7 @@ func NewBabysitter(ctx context.Context, dep *protos.Deployment, logSaver func(*p
 		proxies:        map[string]*proxyInfo{},
 	}
 	go b.statsProcessor.CollectMetrics(b.ctx, b.readMetrics)
+	fmt.Fprintln(os.Stderr, "Created babysitter")
```
mwhittaker, line 174: Delete?

mwhittaker (approved):
:lock:L:lock:G:lock:T:lock:M:lock:

PR #190, opened 2023-03-24
Hello Example return on error

amammay:
Noticed in the error handling block that the func did not return when an error would occur from the reverser.


spetrovic77 (approved):
Nice catch, thanks!

PR #191, opened 2023-03-24
Made envelopes push components and routing info.

mwhittaker:
Before this PR, a weavelet issued [blocking gets][1] to the envelope to learn which components to run and to get updated routing information. This version-based long polling API was complicated to explain. People had to understand what long polling was and learn the details of our version semantics. I also felt that it was a bit tricky to implement a deployer using this API.

This PR replaces the long polling with envelope-initiated RPCs: `UpdateComponents` and `UpdateRoutingInfo`. Now, whenever the set of components changes or a component's routing info changes, the deployer pushes an update to the weavelet.

I think this API is simpler, but a deployer implementer does have to be careful not to forget to call `UpdateComponents` or `UpdateRoutingInfo`. This can be tricky when events happen in unexpected orders. For example, a weavelet may subscribe to a component before the component exists.

With this API, our weavelet also has to be careful about how it handles `UpdateComponents` and `UpdateRoutingInfo`. In general, a weavelet handler should never block or touch the pipe. Doing so makes it easy for a deployer to deadlock. For example, imagine an envelope receives a `StartComponent` request. It grabs a lock and then issues an `UpdateComponents` request. If the weavelet then issues another RPC over the pipe, and the envelope handles this RPC by grabbing the same lock, there is deadlock.

### Future Work

Note that there are still lingering artifacts in the pipe API. There are still co-location groups here and there despite the work in PR #181. The weavelet still assumes placement based on groups. I'll continue to fix these issues in future PRs.

### Details

I removed versioned.Versioned from the weavertest and weaver multi deployers. The SSH deployer still uses it, but we can remove that in the future.

[1]: https://javascript.info/long-polling


spetrovic77 (approved):

weavelet.go
```diff
@@ -39,22 +39,14 @@ import (
...
-	// CollectLoad returns the latest load information at the weavelet.
-	CollectLoad() (*protos.WeaveletLoadReport, error)
-}
-
-// A weavelet is responsible for running and managing Service Weaver components. As the
-// name suggests, a weavelet is analogous to a kubelet or borglet. Every
-// weavelet executes components for a single Service Weaver process, but a process may be
-// implemented by multiple weavelets.
+// A weavelet runs and manages components. As the name suggests, a weavelet is
+// analogous to a kubelet or borglet.
```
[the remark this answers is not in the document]
mwhittaker: Done.

weavelet.go
```diff
@@ -110,6 +97,9 @@ func newWeavelet(ctx context.Context, componentInfos []*codegen.Registration) (*
 		tcpClients:       map[string]*client{},
 		loads:            map[string]*loadCollector{},
 	}
+
+	// TODO(mwhittaker): getEnv starts the WeaveletConn handler which calls
```
mwhittaker, line 101: @spetrovic77 I realized that the `WeaveletConn` has the same annoying set up issues as the `EnvelopeConn`. Here, I want to set up the connection to the envelope, but I don't want to start receiving requests until `w` is fully set up.

dnephin: Would the solution (at least in the short-term) be to move the `WeaveletConn.Serve` call from `newRemoteEnv`, to `weavelet.start` ? That seems to be where other goroutines are started for the weavelet, and at that point `w` would be fully constructed.

That would require adding a new method to the `env` interface that is only implemented by `remoteEnv`. There's already one method that is only used in remote mode (`WeaveletListener`).

If that sounds right, maybe the `WeaveletListener` method on `env` could be replaced by:

```go
type env interface {
    ...
    // remote returns nil in single process mode, and returns the
    // remoteEnv in multi process mode.
   remote() *remoteEnv
```

Or maybe avoid the `remote()` method and instead use a type assertion?

`weavelet.start` could do something like this, while avoiding no-op implementations of these methods on `singleProcessEnv`:

```go
if !w.info.SingleProcess {
    remote := env.(*remoteEnv)
    lis := remote.conn.Listener()
    ...
    go remote.conn.Serve()
```

If this approach sounds interesting, I can prepare a PR based on this one with those changes.

mwhittaker: This sounds great to me! Definitely feel free to open a PR. 

One thing that's bad about our `Conn` APIs right now is that you cannot issue RPCs until you call `Serve`. Er well, you can, but they block because `Serve` needs to run to process the RPC replies. This means that if you delay calling `Serve`, you have to be very careful not to call any RPCs or else things deadlock. Srdjan is working on improving the API to avoid this.

spetrovic77: Looks like this is really flakey. @dnephin, we're going to have to fix it ASAP, hope you're not halfway to fixing! :)

PR #192, opened 2023-03-24
Removed groups entirely from deployer API!

mwhittaker:
Recently, we've been doing a lot of refactoring to remove co-location groups from the deployer API: #184, #182, #181. This PR caps this effort off by removing them entirely. Now, the weavelet is completely unaware of groups. It starts whatever components the deployer tells it to start.

To enable this change, I added a boolean `Local` field to the `RoutingInfo` proto. When this field is set, a weavelet routes requests locally. Otherwise, it routes requests over the network. Note that this is orthogonal to whether or not a component is running locally. A weavelet may want to call a remote instance of a component even if it has a local instance running. Routed objects, for example, do this to respect sharding.

This also changes the flow of weaver.Get. Before, when a weavelet called `weaver.Get[T]`, if `T` was local and unrouted, the weavelet would create a local instance and route to it immediately. Now, `Get[T]` blocks until it receives a `RoutingInfo` from the deployer to learn whether it should be routing locally or remote.

Right now, the decision to route locally is rigid. For a given component, you either always route locally or always don't. In the future, we could make this more flexible.


internal/register/writeonce.go
```diff
@@ -0,0 +1,80 @@
...
+// Package register implements a write-once register.
+package register
+
+import (
+	"fmt"
+	"sync"
+)
+
+// WriteOnce is a concurrent write-once register.
+type WriteOnce[T any] struct {
```
spetrovic77, line 24: Pretty cool!

spetrovic77 (approved):
Nice work!

PR #193, opened 2023-03-26
Reduce overlap between Envelope and EnvelopeConn

dnephin:
Hello service weaver team! This PR explores another option for the relationship between `internal/envelope/conn.Conn`, `internal/envelope/conn.EnvelopeConn` and `runtime/envelope.Envelope`. Currently `internal/envelope` contains a `conn.Conn` used for communicating over pipes, and both an envelope and weavlet specific type for handling messages on either side of that connection. This approach has some nice properties, but also seems to result in a lot of method mirroring between `Envelope` and `EnvelopeConn` and similarly between `remoteEnv` and `WeaveletConn`.

The approach explored in this PR is to first export `Conn` and a few of its methods. The exported type has an interface that looks like this:

```go
type conn interface {
  Send(msg proto.Message) error
  Recv(msg Resp) error
  RPC(request proto.Message) (Resp, error)
  Cleanup(err error)
}
```

By exporting `Conn` the `EnvelopeConn` type can be moved into `runtime/envelope` next to `Envelope`. Moving `EnvelopeConn` allows for the removal of the duplicate `EnvelopeHandler` type, and also some of the mirrored methods (`HealthStatusRPC`, `GetLoadInfoRPC`, etc).

Along the way I noticed two things:
1. The `EnvelopeConn.send` and `WeavletConn.send` methods can be removed (6c78a73a2e85dfc3253948bb186590ea4db51b3a). These methods were only adding a call to `Conn.Cleanup` on error, but the underlying `Conn.Send` already did this, so the extra `Cleanup` was a no-op.
2. The `EnvelopeConn.RPC` and `WeaveletConn.RPC` methods can be removed (2a15611ad92e4bd1aac583281ce2faab9682b0d7) by adding a type parameter to `Conn`.

These two changes could be implemented without moving `EnvelopeConn`. I could cherry-pick those changes onto main if the rest of this change is not accepted.

This PR only moves `EnvelopeConn`, but I think a similar change could be made with `WeaveletConn`. That's probably better off after #191 is merged.

I think the biggest disadvantage to this approach is that now the `runtime/envelope` package has an `EnvelopeConn` type exported with a couple methods on it. Un-exporting this type is difficult because `weavertest` needs to access these internals. I believe the `WeaveletConn` doesn't have this problem, so maybe it would be better to try this on `WeaveletConn` first.

I've made this PR a draft initially to gather feedback, and wait for #191 to merge.


dnephin:
Thanks for the review! I agree about the type parameter on `Conn`, it felt a bit strange. I think renaming the method on `EnvelopeConn` and `WeaveletConn` instead of the type parameter would be good. Embedding `EnvelopeConn` into `Envelope` could work too, I'll explore that option a bit. I expect that should work to remove the extra methods, but not the duplicate interface. I suspect the only way to remove the extra interface is to move EnvelopeConn into the `runtime/envelope` package.

I've just rebased with PR with the latest `main`. During that rebase I noticed I was able to remove the `connection` type in `weavertest`. With the change to use `Envelope` directly in `weavertest` it no longer needs to switch between `EnvelopeConn` and `Envelope`. 

I agree it's not ideal that Envelope can start a process sometimes and not others. I was able to hide that distinction from the public interface, but that's only part of the problem.

dnephin:
While thinking about embedding `EnvelopeConn` into `Envelope` I realized that most likely re-introduces the problem that I think the duplicate `EnvelopeHandler` interfaces are meant to solve. The godoc for `Envelope` will not show those methods as methods on `Envelope`. Someone reading the docs would have to look at the internal `EnvelopeConn` type.

I'll open a separate PR for the cleanup related to `Conn.Send` and `Conn.RPC`. I'll leave this PR as a draft for now, and see what this change looks like after rebasing with #194.

dnephin:
The only way to remove the second copy of the `EnvelopeHandler` interface would be to move `EnvelopeConn.Serve` into `runtime/envelope` , which I think effectively requires moving all of `EnvelopeConn`.  That introduces the problem of `EnvelopeConn` being an exported type with a bunch of exported methods that consumers of `runtime/envelope` should never use.  To address that secondary problem we'd need either a separate `Envelope` constructor for `weavertest`, or to pass the Reader and Writer via context (as was done in this PR).  Neither of those seem worth it.

`EnvelopeHandler` has fewer methods now, and the internal copy no longer has the comments, so it should be easier to maintain already.

I'm going to close this PR. Some of these changes were merged as part of #200, and the rest can't be merged because of the problems described above.

PR #194, opened 2023-03-27
Graceful cleanup of babysitter, envelope, and envelope conn.

spetrovic77:
In particular, we allow the caller of the above APIs to:
  * Stop the babysitter/envelope/envelope-conn early by closing the context.
  * Wait for the babysitter/envelope/envelope-conn to terminate.

By the time the babysitter/envelope/envelope-conn is terminated, we guarantee that all of the internal state has been cleaned up.

Other changes:
  * After discussion with mwhittaker@, changed the envelope/envelope-conn APIs to pass EnvelopeHandler to a call to Serve(), instead of the constructor.


internal/envelope/conn/envelope_conn.go
```diff
@@ -131,20 +140,41 @@ func (e *EnvelopeConn) Serve() error {
...
+				}
+			case <-e.ctx.Done():
+				e.stop(e.ctx.Err())
+				return e.ctx.Err()
 			}
 		}
 	})
+}
+
+// Wait waits for the connection to terminate. It returns an error that
```
mwhittaker, line 159: We chatted offline and might merge `Wait` into `Serve`. I think that will make `EnvelopeConn` consistent with `WeaveletConn` as well.

internal/envelope/conn/envelope_conn.go
```diff
@@ -15,8 +15,10 @@
 package conn
 
 import (
+	"context"
 	"fmt"
 	"io"
+	sync "sync"
```
mwhittaker, line 21: Just `"sync"`?

internal/envelope/conn/envelope_conn.go
```diff
@@ -47,36 +49,44 @@ type EnvelopeHandler interface {
...
-		handler: h,
-		conn:    conn{name: "envelope", reader: r, writer: w},
+		ctx:       ctx,
+		ctxCancel: cancel,
+		conn:      conn{name: "envelope", reader: r, writer: w},
 	}
 	// Send the setup information to the weavelet, and receive the weavelet
 	// information in return.
 	if err := e.send(&protos.EnvelopeMsg{WeaveletSetupInfo: wlet}); err != nil {
+		e.stop(err)
```
mwhittaker, line 78: Do we need to `stop` if we haven't called `Serve` yet?

internal/envelope/conn/conn_test.go
```diff
@@ -106,45 +107,35 @@ func makeConnections(t *testing.T, handler conn.EnvelopeHandler) (*conn.Envelope
...
+	e.Serve(handler)
 
-	// Stop goroutines when test has finished.
+	// Stop the goroutine when test has finished.
 	t.Cleanup(func() {
-		wReader.Close()
-		eReader.Close()
-		// NOTE(mwhittaker): wWriter and eWriter are closed by the conns.
-		done.Wait()
+		cancel()
```
mwhittaker, line 137: Should we call `e.Wait()`?

internal/envelope/conn/conn_test.go
```diff
@@ -106,45 +107,35 @@ func makeConnections(t *testing.T, handler conn.EnvelopeHandler) (*conn.Envelope
...
+		done <- struct{}{}
 	}()
 
-	// Wait for both conns to start.
-	started.Wait()
+	e, err := conn.NewEnvelopeConn(ctx, eReader, eWriter, wlet)
+	if err != nil {
+		t.Fatal(err)
+	}
+	<-created
```
mwhittaker, line 132: Why do we need `created`? If `conn.NewEnvelopeConn` returns, then it successfully completed the handshake with the weavelet conn?

weavertest/deployer.go
```diff
@@ -147,10 +138,8 @@ func newDeployer(ctx context.Context, t testing.TB, wlet *protos.WeaveletSetupIn
 	}
 
 	t.Cleanup(func() {
-		// When the unit test ends, kill all subprocs and stop all goroutines.
-		cancel()
-		if err := d.Stop(); err != nil {
-			d.logger.Error("Stop", err)
+		if err := d.cleanup(); err != nil {
+			d.logger.Error("Wait", err)
```
mwhittaker, line 142: Should we rename `cleanup` to `wait`?

weavertest/deployer.go
```diff
@@ -192,68 +180,69 @@ func (d *deployer) Init(config string) weaver.Instance {
...
+	d.running.Go(func() error {
+		err := e.Wait()
+		d.stop(err)
+		return err
+	})
+
+	if err := d.registerReplica(g, e.WeaveletInfo()); err != nil {
+		panic(fmt.Errorf(`cannot register the replica for "main"`))
+	}
+	// TODO(mwhittaker): Issue an UpdateComponents call to the main group.
```
mwhittaker, line 223: You can get rid of this todo. I added a `RunMain` field to the `WeaveletSetupInfo`.

runtime/envelope/envelope_test.go
```diff
@@ -59,15 +61,14 @@ func TestMain(m *testing.M) {
 			fmt.Fprintf(os.Stderr, "unable to create weavelet conn for subprocess: %v\n", err)
 			os.Exit(1)
 		}
+		fmt.Print("weavelet started")
```
mwhittaker, line 64: Delete?

runtime/envelope/envelope_test.go
```diff
@@ -90,6 +91,7 @@ func TestMain(m *testing.M) {
 			fmt.Fprintf(os.Stderr, "subprocess: %v\n", err)
 			os.Exit(1)
 		}
+		<-make(chan struct{}) // block forever as weavelet cannot exit.
```
mwhittaker, line 94: I think `select{}` also works.

internal/babysitter/babysitter.go
```diff
@@ -237,19 +285,20 @@ func (b *Babysitter) startColocationGroup(g *group) error {
...
 		h.envelope = e
-		go func() {
-			if err := e.Serve(b.ctx); err != nil {
-				b.done <- err
-			}
-		}()
 		if err := b.registerReplica(g, e.WeaveletInfo()); err != nil {
 			return err
 		}
+		e.Serve(h)
```
mwhittaker, line 296: I think we need to call `Serve` before `registerReplica`, as `registerReplica` may issue RPCs.

runtime/envelope/envelope.go
```diff
@@ -141,33 +125,82 @@ func (e *Envelope) init(ctx context.Context) error {
...
 	e.conn = conn
-	return nil
+	e.stdoutPipe = outpipe
+	e.stderrPipe = errpipe
+	return e, nil
+}
+
+// Serve accepts incoming messages from the weavelet. Messages that are received
+// are handled as an ordered sequence.
+func (e *Envelope) Serve(h EnvelopeHandler) {
```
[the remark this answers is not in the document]
spetrovic77: Good point, done.

internal/envelope/conn/conn_test.go
spetrovic77: You're right. Done.

internal/envelope/conn/envelope_conn.go
spetrovic77: It's a bit of an overkill. But at the minimum, we should be calling e.conn.cleanup() here.

Reverted to that.

weavertest/deployer.go
spetrovic77: In this case, the functionality is of actual cleanup, since the test has completed.
Leaving as-is, unless you have a strong preference.

mwhittaker: Nope, sounds good.

PR #198, opened 2023-03-27
Remove unused parameters and remote-only methods from env interface

dnephin:
For your consideration, this PR removes a few things from the `env` interface.

The `CreateLogSaver` args were not used. Both callers use `newAttrLogger` to add attributes to the logger. This seems like a nice way to handle attributes. The error return from `CreateTraceExporter` was also not used.

The primary change is to remove the two remote-only methods from the interface. These methods are called from a block that is only used when not single process.

### DIFF
--- a/weavelet.go
+++ b/weavelet.go
@@ -212,10 +212,10 @@ func (w *weavelet) start() (Instance, error) {
 	// For a singleprocess deployment, no server is launched because all
 	// method invocations are process-local and executed as regular go function
 	// calls.
-	if !w.info.SingleProcess {
-		startWork(w.ctx, "serve weavelet conn", w.env.ServeWeaveletConn)
+	if remote, ok := w.env.(*remoteEnv); ok {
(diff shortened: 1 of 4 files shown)


weavelet.go
```diff
@@ -108,7 +108,7 @@ func newWeavelet(ctx context.Context, componentInfos []*codegen.Registration) (*
 	}
 	w.info = wletInfo
 
-	exporter, err := env.CreateTraceExporter()
+	exporter := env.CreateTraceExporter()
 	if err != nil {
```
mwhittaker, line 112: This `if` is stale now?

mwhittaker (approved):
Very clean :soap: :soap: :soap:

weavelet.go
dnephin: Yes, good catch! I moved the call into `sdktrace.WithBatcher()` since that was the only place the `exporter` variable was used.

PR #200, opened 2023-03-27
Remove conn.send and use conn.rpc consistently

dnephin:
This PR applies a couple of the cleanups found while working on #193

The `EnvelopeConn.send` and `WeaveletConn.send` methods can be removed. These methods were only adding a call to `conn.cleanup` on error, but the underlying `conn.send` already did this, so the extra cleanup was a no-op.

A couple new RPC methods were calling `conn.rpc` instead of `EnvelopeConn.rpc` which required them to repeat the error handling already available in `EnvelopeConn.rpc`. A commit in this PR corrects that usage, and renames `conn.rpc` to `conn.doBlockingRPC` to make the method names more distinct. The shorter `{EnvelopeConn, WeavletConn}.rpc` should be more convenient to use than the longer `conn.doBlockingRPC`.

Also moves some response checks from  `Envelope.RunProfiling` to `EnvelopeConn.DoProfilingRPC` so that all the checks happen in the same place.


mwhittaker (approved):
Awesome! :shipit:

PR #202, opened 2023-03-28
Don't call os.Cmd.Wait() before os.Cmd pipe operations are done.

spetrovic77:
This is part of the os.Cmd spec, and also fixes a bunch of races in envelope_test. Suggested by mwhittaker@.


mwhittaker (approved):
Nice!

PR #203, opened 2023-03-28
Replace logger.Logger with slog

rgrandl:


runtime/logging/logger.go
```diff
@@ -38,100 +38,98 @@ type Options struct {
...
 	entry := protos.LogEntry{
-		App:        opts.App,
-		Version:    opts.Deployment,
-		Component:  opts.Component,
-		Node:       opts.Weavelet,
+		App:        h.Opts.App,
+		Version:    h.Opts.Deployment,
+		Component:  h.Opts.Component,
+		Node:       h.Opts.Weavelet,
 		TimeMicros: time.Now().UnixMicro(),
```
mwhittaker, line 96: Should we use `rec.Time`?

runtime/logging/logger.go
```diff
@@ -38,100 +38,98 @@ type Options struct {
...
+}
+
+// Enabled implements the slog.Handler interface.
+func (h *LogHandler) Enabled(context.Context, slog.Level) bool {
+	// Support all logging levels.
+	return true
+}
+
+// WithAttrs implements the slog.Handler interface.
+func (h *LogHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
```
mwhittaker, line 65: Is `WithAttrs` supposed to return a copy of the handler and leave the original handler unmodified? For example, consider the following code: 

```go
logger := ...
logger.WithAttrs([]slog.Attry{{"foo", slog.StringValue("bar")}})
logger.Debug("hi")
```

With this code, it looks like the `Debug` log would have the `"foo"="bar"` attribute, but I would expect it not to? I could be misunderstanding the `Handler` interface though.

runtime/logging/logger.go
```diff
@@ -38,100 +38,98 @@ type Options struct {
...
+	for _, attr := range attrs {
+		h.Opts.Attrs = append(h.Opts.Attrs, attr.Key)
+		h.Opts.Attrs = append(h.Opts.Attrs, attr.Value.String())
+	}
+	return h
+}
+
+// WithGroup implements the slog.Handler interface.
+func (h *LogHandler) WithGroup(string) slog.Handler {
+	return h
```
mwhittaker, line 77: Add a TODO to implement this?

mwhittaker (changes requested):
Awesome! I feel like this change simplifies things a lot. :wood::wood::wood:

runtime/logging/logger.go
```diff
@@ -38,100 +38,105 @@ type Options struct {
...
+		Level:      rec.Level.String(),
+		File:       "",
+		Line:       -1,
+		Msg:        rec.Message,
+		Attrs:      appendAttrs(h.Opts.Attrs, attrs),
 	}
-	l.Write(e)
+
+	// Get the file and line information.
+	details := runtime.FuncForPC(rec.PC)
```
dnephin, line 112: I was just reading over the docs for `slog`, and I noticed this about `rec.PC` (https://pkg.go.dev/golang.org/x/exp/slog#Record):

	// The only valid use for this value is as an argument to
	// [runtime.CallersFrames]. In particular, it must not be passed to
	// [runtime.FuncForPC].
	PC [uintptr](https://pkg.go.dev/builtin#uintptr)
	
It doesn't say why, but do you think this call is safe?

In `slog`, the `rec.PC` is used with `runtime.CallersFrames`, which I think can get you the same info:

https://github.com/golang/exp/blob/10a507213a2916eb81243cb95bcd7f5e70a771e1/slog/handler.go#L288

https://github.com/golang/exp/blob/10a507213a2916eb81243cb95bcd7f5e70a771e1/slog/record.go#L70-L74

dnephin (commented):
Nice change! I noticed one thing, sorry I was too slow to leave the comment before merge!

runtime/logging/logger.go
rgrandl: Thanks Daniel. Great point.

PR #207, opened 2023-03-28
Allow envelope RPCs to proceed even before Serve() is called.

spetrovic77:
This prevents a deadlock in the following situation.

```
  e, err := NewEnvelope(ctx, wlet, config)
  e.UpdateComponents([]string{"foo", "bar"})
  err = e.Serve()
```

Based on discussions with mwhittaker@.

### DIFF
--- a/internal/queue/queue.go
+++ b/internal/queue/queue.go
@@ -1,0 +1,70 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


internal/queue/queue.go
```diff
@@ -0,0 +1,70 @@
...
+// Unlike a Go channel, Queue doesn't have any constraints on how many
+// elements can be in the queue.
+type Queue[T any] struct {
+	mu    sync.Mutex
+	elems []T
+	wait  *cond.Cond
+}
+
+// Enqueue places elem at the back of the queue.
+func (q *Queue[T]) Enqueue(elem T) {
```
mwhittaker, line 35: I like `Push` and `Pop`, but happy with `Enqueue` and `Dequeue` too!

mwhittaker (approved):
Nice!

PR #213, opened 2023-03-31
move internal babysitter under the multi deployer tool

rgrandl:
The internal babysitter is used only by the multi deployer. It makes sense to move it in the tool/multi directory.

Also, renamed babysitter to deployer, because really there is a single babysitter managing everything which is the actual deployer. Some methods can be private now.


mwhittaker (approved):
Amazing :baby: :baby_bottle: :baby_chick: :baby_symbol:

PR #219, opened 2023-04-03
Added prototype versioning system.

mwhittaker:
## Version Issues

Service Weaver is currently at v0, which means we're making a lot of breaking changes to the code. This has the problem that a deployer and the app its trying to deploy may be at incompatible versions.

For example, imagine someone is trying to learn Service Weaver. They run `go install github.com/ServiceWeaver/weaver@latest`. A couple weeks later, they start writing a new Service Weaver application and run `go mod tidy`. `go mod tidy` picks up the latest version of Service Weaver which is likely incompatible with the stale version of the `weaver` tool they downloaded a couple weeks back. As a result, the `weaver` tool will likely fail to deploy the application, and it won't produce any meaningful error messages explaining why it failed.

## Proposed Solution

This PR introduces a proposed solution. It's not perfect, but hopefully it kickstarts some ideas on how to solve this annoying versioning issue. With this PR, we version the deployer API. A weavelet reports to the
envelope the deployer API version it was built with. The envelope can then check whether it supports the reported version, aborting if not. See `runtime/version.go` for more details.


internal/version/version.go
```diff
@@ -0,0 +1,43 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package version includes constants specifying the semantic version of the
+// core Service Weaver module.
+package version
```
spetrovic77, line 17: Notes on what we discussed offline. We have two different versions:

1. SW module version, which includes *all* of the public API changes (not just the pipe) and is updated according to the Go module versioning strategy (e.g., on every public backward-compatible API change, bump up the minor version, on public backward-incompatible API change, bump up the major version).
2. Pipe version, which is updated only when the pipe protocol changes.

Deployers only care about pipe versions. We decided therefore to change the version in this file to represent the pipe version, with the following properties:

* Pipe version is based on the SW module version that introduced the version change.
* Pipe version follows the same major/minor/patch properties as the SW module version.
* Changes to the SW module's major version trigger changes to the pipe version, and vice-versa.
* Changes to the SW module's minor version don't trigger changes to the pipe version.
* Changes to the pipe minor version trigger changes to the SW module's minor version.

I think all of the above means that, when pipe versions are incompatible, the deployer can print out the weavelet's pipe version, which matches some version of SW module, and ask the user to install a deployer that supports that particular SW module version. Deployers can keep a list of supported SW module versions. That way, the user doesn't have to deal or know what a "pipe version" really is.

internal/version/version.go
```diff
@@ -0,0 +1,43 @@
...
+// [go_generate]: https://developers.redhat.com/articles/2022/11/14/3-ways-embed-commit-hash-go-programs#2__using_go_generate
+// [go_version]: https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/runtime/extern.go;l=279-287;drc=e4435cb8448514d2413f9d9aa3ee40738d26fd67
+
+const (
+	// The semantic version (Major.Minor.Patch) of the core Service Weaver
+	// module. When an envelope spawns a weavelet, the weavelet reports this
+	// version to the envelope. The envelope then errors out if it is not
+	// compatible with the reported version.
+	Major = 0
+	Minor = 16
```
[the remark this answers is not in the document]
mwhittaker: Done.

internal/version/version.go
mwhittaker: Thanks for the offline discussion Srdjan. It helped clarify and solidify a lot of my thoughts. I updated all the comments in this PR to reflect the new model. I also moved the version into the `runtime` package because I realized that other deployers will likely need to use it as well. I'll follow up with a doc describing everything we talked about in more detail, and I'll be sure to include your comment.

mwhittaker (commented):
Thanks Srdjan!

runtime/version.go
```diff
@@ -0,0 +1,43 @@
...
+	// Every time we make a change to deployer API, we assign it a new version.
+	// When an envelope spawns a weavelet, the weavelet reports this version to
+	// the envelope. The envelope then errors out if it is not compatible with
+	// the reported version.
+	//
+	// We could assign the deployer API versions v1, v2, v3, and so on.
+	// However, this makes it hard to understand the relationship between the
+	// deployer API version and the version of the Service Weaver module. (What
+	// version of Service Weaver do I need to install to get version 7 of the
+	// pipe?)
```
spetrovic77, line 30: Nice explanation.

spetrovic77 (approved):
Looks great!

Issue #179, opened 2023-03-21
weaver multi deploy is failure


veterancj:
step1:   weaver multi deploy weaver.toml
╭───────────────────────────────────────────────────╮
│ app        : hello                                │
│ deployment : cb92a3e5-a956-46ff-86ac-269d7b09b0c9 │
╰───────────────────────────────────────────────────╯
S0321 15:16:50.243458 stdout  567250c1           ] hello listener available on 127.0.0.1:12345
S0321 15:16:50.243765 stdout  020d2067           ] hello listener available on 127.0.0.1:12345

step2: curl "localhost:12345/hello?name=Weaver"
keep blocking

step3: weaver multi status
╭──────────────────────────────────────────────────────╮
│ DEPLOYMENTS                                          │
├───────┬──────────────────────────────────────┬───────┤
│ APP   │ DEPLOYMENT                           │ AGE   │
├───────┼──────────────────────────────────────┼───────┤
│ hello │ cb92a3e5-a956-46ff-86ac-269d7b09b0c9 │ 3m56s │
╰───────┴──────────────────────────────────────┴───────╯
╭───────────────────────────────────────────────╮
│ COMPONENTS                                    │
├───────┬────────────┬───────────┬──────────────┤
│ APP   │ DEPLOYMENT │ COMPONENT │ REPLICA PIDS │
├───────┼────────────┼───────────┼──────────────┤
│ hello │ cb92a3e5   │           │              │
│ hello │ cb92a3e5   │ main      │ 15208, 15209 │
╰───────┴────────────┴───────────┴──────────────╯
╭─────────────────────────────────────────────────╮
│ LISTENERS                                       │
├───────┬────────────┬──────────┬─────────────────┤
│ APP   │ DEPLOYMENT │ LISTENER │ ADDRESS         │
├───────┼────────────┼──────────┼─────────────────┤
│ hello │ cb92a3e5   │ hello    │ 127.0.0.1:12345 │
╰───────┴────────────┴──────────┴─────────────────╯

component only one , expect two.

mwhittaker:
Could you also try the following steps. First, delete the following directories. Note that doing so will delete any history of any Service Weaver apps you're running locally.

```console
rm -r ~/.local/share/serviceweaver/
rm -r /tmp/serviceweaver/
```

[...]

PR #220, opened 2023-04-03
Added "weaver version" command.

mwhittaker:
## Example

```console
$ weaver version
weaver version 0.16.0 254d063fcc5a3c27f95869215dfa6bb36bbfbac6 linux/amd64
```

## Overview

This PR introduces a `weaver version` command which shows the version of the `weaver` tool. The output also includes the git commit at which the tool was built as well as the target OS and architecture.

## Motivation

As described in #219, a Service Weaver deployer and the app it is trying to deploy need to have compatible versions. We're seeing some issues (e.g., #179, #154) that I think may be cause by version issues. `weaver version` makes it easier to debug these kinds of issues, as people can include the output of `weaver version` in their bug reports.


cmd/weaver/main.go
```diff
@@ -79,6 +83,23 @@ func main() {
 		}
 		return
 
+	case "version":
+		semver := fmt.Sprintf("%d.%d.%d", version.Major, version.Minor, version.Patch)
+		commit := "?"
+		if info, ok := debug.ReadBuildInfo(); ok {
```
rgrandl, line 89: nice, didn't know about it.

rgrandl (approved):

PR #221, opened 2023-04-03
Added "weaver purge" command.

mwhittaker:
## Example

```console
$ weaver multi purge
WARNING: You are about to delete the following directories used to store logs
and data for local Service Weaver applications. This data will be deleted
immediately and irrevocably. Are you sure you want to proceed?

    - /tmp/serviceweaver/logs/weaver-multi
    - /usr/local/google/home/mwhittaker/.local/share/serviceweaver/multi_registry

Enter (y)es to continue: y

Deleting /tmp/serviceweaver/logs/weaver-multi... ✅
Deleting /usr/local/google/home/mwhittaker/.local/share/serviceweaver/multi_registry... ✅
```

## Overview

This PR introduces `weaver single purge` and `weaver multi purge` commands that delete all logs and data produced by locally run Service Weaver applications. In the future, we may also want to kill lingering processes. For example, we could kill the gke-local controller and distributors.

## Motivation

The motivation for this PR is the same as #220. We're seeing some issues (e.g., #179, #154) that I think may be cause by version issues. Users run `weaver multi logs` on logs produced by a different version of Service Weaver, for example. `weaver purge` makes it easier to debug these kinds of issues, as we can tell people to run `weaver purge` before reproducing their reported bug.

### DIFF
--- a/internal/tool/multi/multi.go
+++ b/internal/tool/multi/multi.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"path/filepath"
 
+	"github.com/ServiceWeaver/weaver/internal/must"
 	"github.com/ServiceWeaver/weaver/internal/status"
 	"github.com/ServiceWeaver/weaver/runtime/logging"
 	"github.com/ServiceWeaver/weaver/runtime/tool"
@@ -41,6 +42,14 @@ var (
 		},
 	}
 
+	purgeSpec = &tool.PurgeSpec{
+		Tool: "weaver multi",
+		Paths: []string{
+			logdir,
+			must.Must(defaultRegistryDir()),
+		},
+	}
+
 	Commands = map[string]*tool.Command{
 		"deploy": &deployCmd,
 		"logs": tool.LogsCmd(&tool.LogsSpec{
@@ -53,5 +62,6 @@ var (
 		"status":    status.StatusCommand("weaver multi", defaultRegistry),
 		"metrics":   status.MetricsCommand("weaver multi", defaultRegistry),
 		"profile":   status.ProfileCommand("weaver multi", defaultRegistry),
+		"purge":     tool.PurgeCmd(purgeSpec),
 	}
 )
--- a/internal/tool/single/single.go
+++ b/internal/tool/single/single.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 
 	"github.com/ServiceWeaver/weaver/internal/files"
+	"github.com/ServiceWeaver/weaver/internal/must"
 	"github.com/ServiceWeaver/weaver/internal/status"
 	"github.com/ServiceWeaver/weaver/runtime/tool"
 )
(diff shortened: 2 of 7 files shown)


internal/tool/multi/multi.go
```diff
@@ -53,5 +61,13 @@ var (
 		"status":    status.StatusCommand("weaver multi", defaultRegistry),
 		"metrics":   status.MetricsCommand("weaver multi", defaultRegistry),
 		"profile":   status.ProfileCommand("weaver multi", defaultRegistry),
+		"purge":     tool.PurgeCmd(purgeSpec),
 	}
 )
+
+func must[T any](x T, err error) T {
```
rgrandl, line 68: can you add a comment?

rgrandl (commented):
The PR looks great, I have some thoughts though. Why do we need a command to delete only the directories? What if someone has things running, and by mistake runs the purge command?

I was wondering if we should have a kill/stop/clear/reset command instead. Or rename purge to one of these names, and later on we can also stop processes.

internal/tool/multi/multi.go
mwhittaker: Done.

mwhittaker:
> The PR looks great, I have some thoughts though. Why do we need a command to delete only the directories? What if someone has things running, and by mistake runs the purge command?
> 
> I was wondering if we should have a kill/stop/clear/reset command instead. Or rename purge to one of these names, and later on we can also stop processes.

Thanks Robert! We chatted offline. We'll stick with the name "purge" and make it kill processes as well. I also strengthened the warning message.

internal/tool/multi/multi.go
```diff
@@ -53,5 +61,18 @@ var (
 		"status":    status.StatusCommand("weaver multi", defaultRegistry),
 		"metrics":   status.MetricsCommand("weaver multi", defaultRegistry),
 		"profile":   status.ProfileCommand("weaver multi", defaultRegistry),
+		"purge":     tool.PurgeCmd(purgeSpec),
 	}
 )
+
+// must is a generic implementation of the Go must idiom [1, 2]. It panics if
```
rgrandl, line 68: very nice. Thanks for the comment, I learned something new too.

internal/tool/single/single.go
```diff
@@ -36,14 +36,32 @@ var (
...
 	Commands = map[string]*tool.Command{
 		"status":    status.StatusCommand("weaver single", defaultRegistry),
 		"dashboard": status.DashboardCommand(dashboardSpec),
 		"metrics":   status.MetricsCommand("weaver single", defaultRegistry),
 		"profile":   status.ProfileCommand("weaver single", defaultRegistry),
+		"purge":     tool.PurgeCmd(purgeSpec),
 	}
 )
 
+// must is a generic implementation of the Go must idiom [1, 2]. It panics if
```
rgrandl, line 53: can we put this method somewhere in a shared directory? Right now is duplicated among single, multi.

runtime/tool/purge.go
```diff
@@ -0,0 +1,113 @@
...
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+	"text/template"
+)
+
+// PurgeSpec configures the command returned by PurgeCmd.
+type PurgeSpec struct {
```
rgrandl, line 28: Maybe add a TODO here; probably we will also specify some reqular expression or smthg in the `PurgeSpec` based on which we will kill all processes/subprocesses for each deployer?

rgrandl (approved):
Nice!

internal/tool/single/single.go
mwhittaker: Done.

runtime/tool/purge.go
mwhittaker: Done. Good idea.

mwhittaker (commented):
Thanks Robert!

PR #223, opened 2023-04-04
Performed miscellaneous multi deployer cleanups.

mwhittaker:
This PR cleans up the weaver multi deployer.

- I removed the use of `protos.Deployment`. I'm hoping we can remove this proto eventually. It should be obsolete now.
- The deployer now instantiates its own logs and trace databases.
- I renamed some variables and updates some stale comments.


internal/tool/multi/deployer.go
```diff
@@ -405,16 +394,13 @@ func (d *deployer) registerReplica(g *group, info *protos.WeaveletInfo) error {
...
 }
 
 // HandleTraceSpans implements the envelope.EnvelopeHandler interface.
-func (d *deployer) HandleTraceSpans(_ context.Context, spans []trace.ReadOnlySpan) error {
-	if d.traceSaver == nil {
-		return nil
-	}
-	return d.traceSaver(spans)
+func (d *deployer) HandleTraceSpans(ctx context.Context, spans []trace.ReadOnlySpan) error {
+	return d.traceDB.Store(ctx, d.config.Name, d.deploymentId, spans)
```
rgrandl, line 403: ditto

internal/tool/multi/deployer.go
```diff
@@ -405,16 +394,13 @@ func (d *deployer) registerReplica(g *group, info *protos.WeaveletInfo) error {
 
 // HandleLogEntry implements the envelope.EnvelopeHandler interface.
 func (d *deployer) HandleLogEntry(_ context.Context, entry *protos.LogEntry) error {
-	d.logSaver(entry)
+	d.logsDB.Add(entry)
```
rgrandl, line 397: wouldn't this panic if logsDB is nil?

Somehow it assumes that someone created a deployer with newDeployer.

internal/tool/multi/deployer.go
```diff
@@ -48,27 +48,15 @@ const defaultReplication = 2
...
 // A deployer manages an application deployment.
 type deployer struct {
-	ctx       context.Context
-	ctxCancel context.CancelFunc
-	dep       *protos.Deployment
-	started   time.Time
-	logger    *slog.Logger
-	running   errgroup.Group
-
-	// logSaver processes log entries generated by the weavelet. The entries
```
rgrandl, line 58: I like that we could delete these functions.

rgrandl (approved):
Nice cleanups Michael!

internal/tool/multi/deployer.go
mwhittaker: Me too! I think when the multiprocess deployer was used for weavertests, we had to pass these in, but now we don't :)

internal/tool/multi/deployer.go
mwhittaker: That's true! But, I don't think any of a deployer's methods would work if you don't construct it with `newDeployer`, so I think it's okay?

mwhittaker (commented):
Thanks Robert!

PR #230, opened 2023-04-06
internal/net/call: close listener before Serve exits

dnephin:
This PR makes a few small improvements to the `internal/net/call` package.

Primarily it moves the close of `net.Listener` from the caller into `Serve`. This matches the behaviour of https://pkg.go.dev/net/http#Server.Serve, which says "Serve always returns a non-nil error and closes l". There are 3 other related `Serve` functions in the `net/http` packakge which don't document this behaviour of closing the listener, but from what I can tell they all behave the same (because they all call `Server.Serve`).

This change allows me to remove the `listener.Close` call from a number of tests. Notably `TestCancelServe` now behaves correctly. Previously cancelling the context would not cause `Serve` to exit (it was working in practice because the caller was closing the listener).

The second change was to the initialization of `ClientOptions.Balancer`. Originally I wanted to move the initialization of `opts.Balancer` into `withDefaults` so that all of the initialization happened in the same place, however that caused some tests to fail because the client had a second reference to the balancer. This second commit removes the extra reference. The lock continues to guard calls to `balancer.Pick`. The `balancer` reference itself is never modified after the type is initialized.


internal/net/call/call.go
```diff
@@ -843,7 +846,6 @@ func logError(logger *slog.Logger, details string, err error) {
 	if errors.Is(err, context.Canceled) ||
 		errors.Is(err, io.EOF) ||
 		errors.Is(err, io.ErrUnexpectedEOF) ||
-		errors.Is(err, io.ErrClosedPipe) ||
 		errors.Is(err, io.ErrClosedPipe) {
```
dnephin, line 849: This appeared to be checked twice, so I assume it's safe to remove one of them.

internal/net/call/call_test.go
```diff
@@ -1343,7 +1343,7 @@ func TestCancelServe(t *testing.T) {
 	go func() {
 		opts := call.ServerOptions{Logger: logging.NewTestLogger(t)}
 		err := call.Serve(ctx, lis, handlers, opts)
-		if err != nil && ctx.Err() == nil {
+		if err != ctx.Err() {
```
dnephin, line 1346: This function can never exit with `nil`, so I think what we want to check here is that the returned error is the `ctx.Err`, and not anything else.

weavelet.go
```diff
@@ -234,24 +234,9 @@ func (w *weavelet) start() (Instance, error) {
 			}
 		}
 
-		// Start serving the transport.
-		serve := func(lis net.Listener, transport *transport) {
-			if lis == nil || transport == nil {
-				// TODO(spetrovic): Can this ever happen?
-				return
-			}
```
dnephin, line 242: I looked into this TODO. `lis` is initialized in the constructor to `WeaveletConn`

https://github.com/ServiceWeaver/weaver/blob/65037e069c1fbce40887617eec7a167f463efda6/internal/envelope/conn/weavelet_conn.go#L94

And `transport` is initialized in `newWeavelet`

https://github.com/ServiceWeaver/weaver/blob/65037e069c1fbce40887617eec7a167f463efda6/weavelet.go#L150

Which I think means neither of these can be nil in practice, so I removed the TODO and these checks for nil.

dnephin (commented):
I had a few questions while reading over this package.

First I noticed that `MethodKey` is 16 bytes

https://github.com/ServiceWeaver/weaver/blob/65037e069c1fbce40887617eec7a167f463efda6/internal/net/call/handlers.go#L24

but it's populated from the first 16 bytes of a 32 byte SHA256 checksum. I guess this is ok because we expect a small enough number of functions that this shouldn't collide, is that right? And increasing this size would add 16 bytes to the message header.

Second I notice that both

https://github.com/ServiceWeaver/weaver/blob/65037e069c1fbce40887617eec7a167f463efda6/internal/net/call/call.go#L149 

and `serverConnection.version` are only ever written to (never read). I guess this version is the RPC protocol version, and since the only implemented version is 0 there's no reason to read it. Is that right?

Finally, there are some great comments in `msg.go` and `call.go`. Now that godoc supports headers I think it would be great to get that into the godoc for this package. What do you think about moving both the `call.go` comment (https://github.com/ServiceWeaver/weaver/blob/main/internal/net/call/call.go#L18-L67) and the `msg.go` comment that describes the message format into the package godoc?

internal/net/call/call.go
```diff
@@ -188,19 +187,29 @@ type serverState struct {
...
+	}()
+
+	for {
 		conn, err := l.Accept()
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
 		if err != nil {
 			return fmt.Errorf("call server error listening on %s: %w", l.Addr(), err)
 		}
```
dnephin, line 210: I removed the condition in the `for` loop and the return at the end of the function and replaced it with this check. I believe this is roughly the same behaviour as before, but is maybe slightly more clear that the function can only return a non-nil error. 

The one difference is that if the context is cancelled while in `serveConnection` there will be one extra call to `l.Accept`, but that should error immediately because `l.Close` would have been called already. I believe it's much more likely that the context is cancelled while in `Accept` because `serveConnection` only blocks for a very short time on a lock to update the set of connections, where as `Accept` could block for much longer waiting on a new connection.

When there are multiple error conditions I generally write them using a switch:

```suggestion
		switch {
		case ctx.Err() != nil:
			return ctx.Err()
		case err != nil:
			return fmt.Errorf("call server error listening on %s: %w", l.Addr(), err)
		}
```

Do you have any preference on the style (one switch vs two if) ?

internal/net/call/msg.go
```diff
@@ -49,31 +49,31 @@ const currentVersion = initialVersion
...
 //
 // All messages have the following format:
-//	id	[8]byte
-//	type	[1]byte		-- messageType
-//	length	[7]byte		-- length of the remainder of the message
-//	payload	[length]byte	-- message-type-specific data
+//    id        [8]byte       -- identifier used to track the message
+//    type      [1]byte       -- messageType
+//    length    [7]byte       -- length of the remainder of the message
+//    payload   [length]byte  -- message-type-specific data
```
dnephin, line 55: There were some mixes of tabs and spaces in this comment that caused it to format poorly when tabs were set to 4 spaces. I switched them to spaces so that it formats correctly regardless of tab width.

internal/net/call/call.go
```diff
@@ -188,19 +187,29 @@ type serverState struct {
...
+		<-ctx.Done()
+		l.Close()
+	}()
+
+	for {
 		conn, err := l.Accept()
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
 		if err != nil {
```
mwhittaker, line 208: If `l.Accept` fails for non-context related reasons and we get here, do we also have to close `l`? And then if we do, is it safe to race against the goroutine above that might also close `l`? We could use a `sync.Once` to close `l` if we're paranoid.

internal/net/call/call.go
mwhittaker: Good catch.

internal/net/call/call_test.go
mwhittaker: Yeah, this is much better.

mwhittaker (approved):
Excellent change @dnephin! And your comments explaining what you did and why you did it are super helfpul. 

> First I noticed that MethodKey is 16 bytes but it's populated from the first 16 bytes of a 32 byte SHA256 checksum. I guess this is ok because we expect a small enough number of functions that this shouldn't collide, is that right? And increasing this size would add 16 bytes to the message header.

I think that's right. But @ghemawat, who designed this protocol, knows more.

> Second I notice that both and serverConnection.version are only ever written to (never read). I guess this version is the RPC protocol version, and since the only implemented version is 0 there's no reason to read it. Is that right?

Yup, I think we wanted to add it to the protocol, but it's not used yet. 

> Finally, there are some great comments in msg.go and call.go. Now that godoc supports headers I think it would be great to get that into the godoc for this package. What do you think about moving both the call.go comment (https://github.com/ServiceWeaver/weaver/blob/main/internal/net/call/call.go#L18-L67) and the msg.go comment that describes the message format into the package godoc?

That sounds great to me!

internal/net/call/call.go
dnephin: Ah yes, great question! I think we do need to close `l` if we exit from some other error. It looks like `http.Server.Serve` has both `defer l.Close()` and `closeListenersLocked()` to close this listener, and uses `onceCloseListener` to wrap the close with a `sync.Once`.

Another option here might be to create a child context, cancel it when we receive a different error, and allow the goroutine to close `l`. In that case I think we need to wait on the goroutine to exit before exiting from `Serve`. 

I'll explore both of those options.

dnephin: I went with sync.Once, and copied the onceCloseListener from `net/http`. That seems more clear than adding a second context and a channel to wait on goroutine exit.

internal/net/call/call.go
```diff
@@ -188,19 +187,47 @@ type serverState struct {
...
+			l.Close()
 			return fmt.Errorf("call server error listening on %s: %w", l.Addr(), err)
 		}
 		ss.serveConnection(ctx, conn, hmap)
 	}
-	return ctx.Err()
+}
+
+// onceCloseListener wraps a net.Listener, protecting it from multiple Close calls.
+// TODO: replace with sync.OnceValues which should be available in go1.21
```
mwhittaker, line 219: `sync.OnceValues` seems cool!

mwhittaker (approved):
Awesome, thanks Daniel!

PR #231, opened 2023-04-06
Remove call.NetworkAddress

dnephin:
I noticed that `call.NetworkAddress` was never referenced from the `internal/net/call` package (where it lives), and that it's only use was to parse a `call.NetEndpoint`.  

This PR removes the type, and replaces it with a `ParseNetEndpoint` function. This seems to simplify the code a bit. I find that fewer types in a package can make it easier to understand.

In a couple places I changed a field from `NetworkAddress` to either `string` or `NetEndpoint`.


weavelet.go
```diff
@@ -225,12 +225,12 @@ func (w *weavelet) start() (Instance, error) {
 		startWork(w.ctx, "serve weavelet conn", remote.conn.Serve)
 
 		lis := remote.conn.Listener()
-		addr := call.NetworkAddress(fmt.Sprintf("tcp://%s", lis.Addr().String()))
+		addr := call.NetEndpoint{Net: "tcp", Addr: lis.Addr().String()}
```
mwhittaker, line 228: Maybe change `addr` to `string` instead of `call.NetEndpoint` since we never dial it and only ever call the `String` method on it? Up to you.

mwhittaker (approved):
Niiiice! `NetworkAddress` was a bit janky. Code looks much cleaner now. :soap:

weavelet.go
dnephin: I like that! Done

Issue #113, opened 2023-03-06
deploy the application error


y26s4824k264:
![image](https://user-images.githubusercontent.com/54879886/223173107-734ef2d6-c501-4bc5-8c0a-8a331389a529.png)

y26s4824k264:
![image](https://user-images.githubusercontent.com/54879886/223173272-e555aa7c-1ee8-40f3-8620-aa70128a1021.png)

alejandrojnm:
Hi I got the same error but on mac 

```bash
filename "...2f6add44-e6ff-4bff-8d89-8ebe81a69af3.error.log" must have format "<app>.<deployment>.<weavelet>.<level>.log"
``` 

Any idea ??

alejandrojnm:
@mwhittaker I put the name and I got the same error this is the toml file
```toml
[serviceweaver]
name = "api"
binary = "./api"
```

mwhittaker:
Do you mind trying to delete all the log files in `/tmp/serviceweaver/logs/weaver-multi` and then trying again? The ill-formatted log file might still exist and could be messing things up.

alejandrojnm:
I don't have that in my `tmp` folder

aranw:
> I don't have that in my `tmp` folder 

Most likely if you are on a Mac it'll be under the `/var/folders` temporary directory

geange:
On win11, Win+R and input `%temp%`  to open the temp dir. delete all the log files is ok !

comyxy:
> On win11, Win+R and input `%temp%` to open the temp dir. delete all the log files is ok !

%temp%\serviceweaver\logs\weaver-multi

mwhittaker:
I'm closing this issue because it seems resolved? Feel free to re-open if not.

myyrakle:
I think this should also be added to the official documentation(https://serviceweaver.dev/docs.html#multiprocess).
Or, would it be difficult to apply branch handling only for windows?

Currently, it is difficult to solve without looking at this issue.

myyrakle:
https://github.com/ServiceWeaver/weaver/pull/232

What do you think of processing in this form?
Are there side effects or other issues?

PR #232, opened 2023-04-06
Handling . in the app name

myyrakle:
https://github.com/ServiceWeaver/weaver/issues/113

### DIFF
--- a/runtime/logging/files.go
+++ b/runtime/logging/files.go
@@ -147,7 +147,7 @@ func (fs *FileStore) Add(e *protos.LogEntry) {
 // level redundantly in every log entry. Omit them from the log entries and
 // infer them from the log name.
 func filename(app, deployment, weavelet, level string) string {
-	return fmt.Sprintf("%s.%s.%s.%s.log", app, deployment, weavelet, level)
+	return fmt.Sprintf("%s#%s#%s#%s.log", app, deployment, weavelet, level)
 }
 
 // logfile represents a log file for a specific (app, deployment, weavelet,
@@ -164,11 +164,25 @@ func parseLogfile(filename string) (logfile, error) {
 	// TODO(mwhittaker): Ensure that apps, deployments, weavelet ids, levels
 	// don't contain a ".". Or, switch to some other delimiter that doesn't
 	// show up.
-	parts := strings.SplitN(filename, ".", 5)
-	if len(parts) < 5 || parts[4] != "log" {
-		want := "<app>.<deployment>.<weavelet>.<level>.log"
+
+	dotParts := strings.Split(filename, ".")
+	dotPartsLen := len(dotParts)
+
+	if len(dotParts) < 2 || dotParts[dotPartsLen-1] != "log" {
+		want := "<app>#<deployment>#<weavelet>#<level>.log"
 		return logfile{}, fmt.Errorf("filename %q must have format %q", filename, want)
 	}
+
+	noExtension := strings.Join(dotParts[0:dotPartsLen-1], ".")
+
+	parts := strings.Split(noExtension, "#")
+	partsLen := len(parts)
+
+	if partsLen != 4 {
+		want := "<app>#<deployment>#<weavelet>#<level>.log"
+		return logfile{}, fmt.Errorf("filename %q must have format %q", filename, want)
+	}
+
 	return logfile{
 		app:        parts[0],
 		deployment: parts[1],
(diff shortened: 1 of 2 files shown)


mwhittaker (changes requested):
Thanks for PR @myyrakle! 

I think the code&mdash;both before and after this PR&mdash;is buggy if app, deployment, or weavelet have a period in them. For example, if app is `"foo.bar"` and deployment is `"moo.cow"`, this code will parse app as `"foo.bar.moo"` and deployment as `"cow"` which is not what we want. 

I think we should either (1) restrict the set of app, deployment, and weavelet names. For example, disallow periods. Or (2), use a more sophisticated way of picking filenames. For example, serialize a proto with app, deployment, weavelet, and level and base64 encode it. 

I think (2) is overkill. We should probably do (1). The team will chat about what set of names to allow. There are some things to take into account. For example, we've recently made weavelet ids pod names, so we don't want to disallow that. 

I like your idea of fixing things in the short term though, as people are running into issues.  What do you think about revising this PR to simply replace the period separator with a hash `#`? It doesn't fundamentally solve anything, but it will help the common case until we implement a full solution. Doing nothing for now is also totally fine too :)

mwhittaker:
Ah, sorry, maybe I was unclear. I mean that instead of storing files like `app.deployment.weavelet.level.log`, we store them as `app#deployment#weavelet#level.log`.

myyrakle:
Ah, I misunderstood. I think that's fine too.

runtime/logging/files.go
mwhittaker, line 1: What do you think about the following simplifications?

```diff
diff --git a/runtime/logging/files.go b/runtime/logging/files.go
index 42bbf80..02ff6a5 100644
--- a/runtime/logging/files.go
+++ b/runtime/logging/files.go
@@ -162,27 +162,17 @@ type logfile struct {
 // parseLogfile parses a logfile filename.
 func parseLogfile(filename string) (logfile, error) {
 	// TODO(mwhittaker): Ensure that apps, deployments, weavelet ids, levels
-	// don't contain a ".". Or, switch to some other delimiter that doesn't
+	// don't contain a "#". Or, switch to some other delimiter that doesn't
 	// show up.
-
-	dotParts := strings.Split(filename, ".")
-	dotPartsLen := len(dotParts)
-
-	if len(dotParts) < 2 || dotParts[dotPartsLen-1] != "log" {
-		want := "<app>#<deployment>#<weavelet>#<level>.log"
+	const want = "<app>#<deployment>#<weavelet>#<level>.log"
+	prefix, hasLogSuffix := strings.CutSuffix(filename, ".log")
+	if !hasLogSuffix {
 		return logfile{}, fmt.Errorf("filename %q must have format %q", filename, want)
 	}
-
-	noExtension := strings.Join(dotParts[0:dotPartsLen-1], ".")
-
-	parts := strings.Split(noExtension, "#")
-	partsLen := len(parts)
-
-	if partsLen != 4 {
-		want := "<app>#<deployment>#<weavelet>#<level>.log"
+	parts := strings.SplitN(prefix, "#", 4)
+	if len(parts) < 4 {
 		return logfile{}, fmt.Errorf("filename %q must have format %q", filename, want)
 	}
-
 	return logfile{
 		app:        parts[0],
 		deployment: parts[1],
```

And do you mind adding the following unit test to `files_test.go`?

```go
func TestParseLogfile(t *testing.T) {
        for _, want := range []logfile{
                // Simple strings.
                {"a", "b", "c", "d"},

                // Some empty strings.
                {"", "b", "c", "d"},
                {"a", "", "c", "d"},
                {"a", "b", "", "d"},
                {"a", "b", "c", ""},
                {"", "", "", ""},

                // Periods.
                {"a", "b.b", "c.c.c", "d.d.d.d"},

                // File suffix.
                {"log", "log", "log", "log"},
        } {
                name := filename(want.app, want.deployment, want.weavelet, want.level)
                t.Run(name, func(t *testing.T) {
                        got, err := parseLogfile(name)
                        if err != nil {
                                t.Fatalf("parseLogFile(name): %v", err)
                        }
                        if got != want {
                                t.Errorf("parseLogFile(name): got %v, want %v", got, want)
                        }
                })
        }
}
```

mwhittaker (changes requested):
Thanks @myyrakle!

runtime/logging/files.go
myyrakle: Simple and looks good to me :)

I didn't use `strings.CutSuffix` because it supports the latest version, but the minimum version of weaver is 1.20, so I knew there was no problem!

myyrakle: After reflecting the feedback again, I committed it.
Thank you for your meticulous feedback!

mwhittaker (commented):
Awesome, thank you!

mwhittaker (approved):

PR #234, opened 2023-04-06
Improve the error message about missing WithConfig[T] embedding

nipun-sehrawat:
While we suggest the user to embed an appropriate `WithConfig[T]`, it's easy for users to make that change and retry building their app immediately.

However, without running `weaver generate` again, they'll see the same error message again, which is likely to confuse them.  Hence, this CL improves the error message to also remind the users to run `weaver generate` again after making appropriate changes to their component.


mwhittaker (approved):
Awesome! I think these small quality of life things go a long way.

PR #235, opened 2023-04-07
Added example wrap application, which wraps text.

mwhittaker:
@rgrandl and I have been planning a Service Weaver demo. We wanted to show an app that was simple but interesting. Also, if we could host the app on GKE and let people play with it, that would be a plus. We settled on a text wrapping application with a web frontend.


examples/wrap/wrap.go
```diff
@@ -0,0 +1,101 @@
...
+		}
+	}
+	if len(part) > 0 {
+		parts = append(parts, string(part))
+	}
+	return parts
+}
+
+// firstRune returns the first rune of s, or panics if s is empty.
+func firstRune(s string) rune {
```
rgrandl, line 96: this is a little bit of magic for me. can we expand the commend here, or add this as a function inside the `wrapLine` + comment. 

Also, should we panic, or return an error with a meaningful message (based on the usage of this function) and propagate it all the way to `Wrap`? Right now we always return nil for err, but this might be a usecase for returning smthg else.

rgrandl (approved):
Looks great Michael. Thank you!

examples/wrap/wrap.go
mwhittaker: Moved to `wrapLine`. Go doesn't have a convenient way to get the first rune of a string. `s[0]` returns the first byte, but not the first rune, whereas iterating over a string returns its runes. The code will never panic since we never call it on empty strings.

mwhittaker (commented):
Thanks Robert!

PR #236, opened 2023-04-07
Allow the user to change line length in wrap app.

mwhittaker:
This PR adds functionality to the wrap app to allow a user to change the length to which lines are wrapped.

[Wrap Demo.webm](https://user-images.githubusercontent.com/3654277/230654946-430141db-2eb5-4477-bc42-00f8eafa1326.webm)

### DIFF
--- a/examples/wrap/index.html
+++ b/examples/wrap/index.html
@@ -30,7 +30,7 @@
           <textarea id="unwrapped" placeholder="Enter text here..."></textarea>
         </div>
         <div>
-          <h1>Wrapped to 80 Characters</h1>
+          <h1>Wrapped to <input id="linewidth" type="number" value="40" min="1"/> Characters</h1>
           <pre id="wrapped"></pre>
         </div>
       </div>
--- a/examples/wrap/main.go
+++ b/examples/wrap/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"strconv"
 
 	"github.com/ServiceWeaver/weaver"
 )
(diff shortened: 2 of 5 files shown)


examples/wrap/index.html
```diff
@@ -30,7 +30,7 @@ <h1>Unwrapped</h1>
           <textarea id="unwrapped" placeholder="Enter text here..."></textarea>
         </div>
         <div>
-          <h1>Wrapped to 80 Characters</h1>
+          <h1>Wrapped to <input id="n" type="number" value="80" min="1"/> Characters</h1>
```
rgrandl, line 33: Should we have a much smaller value by default? Like 40 characters? I am wondering if for a live demo, will spend some time texting to get to 80 characters.

rgrandl, line 33: what about 'width' or 'wrappedLen' or smthg? `n` is too generic.

examples/wrap/main.go
```diff
@@ -74,8 +75,30 @@ func main() {
...
+			}
+		}
+		if n <= 0 {
+			http.Error(w, fmt.Sprintf("invalid n: %d", n), http.StatusBadRequest)
+			return
+		}
+
+		// Parse s.
+		s := r.URL.Query().Get("s")
+		if len(s) > 10000 {
```
rgrandl, line 95: any particular reason we put a limit on the length of the string to wrap? If so, we should be clear in specs or in the error message about this limit.

examples/wrap/style.css
```diff
@@ -33,16 +32,22 @@
   margin: 0.5rem;
 }
 
+#n {
```
rgrandl, line 35: same as everywhere else, `n` should be named something less generic.

rgrandl (approved):
Awesome!

examples/wrap/index.html
mwhittaker: Good idea. Done.

mwhittaker: Done.

examples/wrap/main.go
mwhittaker: Added the limit to the error message. I just wanted to avoid someone sending a super big string and causing the service to stall.

examples/wrap/style.css
mwhittaker: Done.

mwhittaker (commented):
Thanks Robert!

PR #238, opened 2023-04-07
Replaced ErrRetriable with RemoteCallError.

mwhittaker:
## Retrying, The API

Before this PR, we had an error called `weaver.ErrRetriable`. The idea was that component method calls would sometimes return an error with an embedded `ErrRetriable`. You could detect this error with `errors.Is` and retry the failed method call:

```go
var val string
var err error
for i := 0; i < 5; i++ {
    val, err = cache.Get(ctx, "key")
    if errors.Is(err, weaver.ErrRetriable) {
        // Retriable system error! Retry.
        continue
    }
    break
}
```

In theory, this is great, but what exactly does it mean for an error to be retriable? Does it mean that the method call was guaranteed to never have started executing? If so, why don't we just retry the error for the user? Or does it mean that the method call was maybe partially or fully executed? But then, it might not be safe to retry the call. And what about system errors that are not retriable? How do I detect those?

## Retrying the API

This PR throws out `ErrRetriable` and replaces it with a much simpler and more straightforward `weaver.RemoteCallError` error. A `RemoteCallError` indicates that a component method call failed to execute properly (e.g., because of a failed machine or a network partition). That's it. It doesn't imply anything about the retry-ability of a call. Here's an illustrative example:

```go
// Call the foo.Foo method.
err := foo.Foo(ctx)
if errors.Is(err, weaver.RemoteCallError) {
    // foo.Foo did not execute properly.
} else if err != nil {
    // foo.Foo executed properly, but returned an error.
} else {
    // foo.Foo executed properly and did not return an error.
}
```

It is up to the user to retry failed calls carefully. Actually, this was always true, but I think the new API makes that clearer. Internally, if we detect that a method call failed and we are 100% sure it never started executing (e.g., because we failed to make a network connection in the first place), we can retry the call transparently for the user.

The new API is also easier to implement and allows us to remove some ugly `WrapError` related code that I had to introduce to avoid some cyclic dependencies.

In the future, we can extend the API with a set of canonical error codes if we want, similar to [this][1].

[1]: https://cloud.google.com/apis/design/errors


runtime/codegen/errors.go
```diff
@@ -31,3 +34,40 @@ func CatchPanics(r interface{}) error {
...
+}
+
+// Error implements the error interface.
+func (j joinedErrors) Error() string {
+	return fmt.Sprintf("%v: %v", j.err1, j.err2)
+}
+
+// Is makes joinedErrors compatible with errors.Is.
+func (j joinedErrors) Is(err error) bool {
+	return errors.Is(j.err1, err)
```
ghemawat, line 61: Is it intentional that errors.Is(j.err2, err) is not consulted? From what I can tell, the behavior here is currently incompatible with errors.Join().

runtime/codegen/errors.go
```diff
@@ -31,3 +34,40 @@ func CatchPanics(r interface{}) error {
...
+}
+
+// As makes joinedErrors compatible with errors.Ass.
+func (j joinedErrors) As(target any) bool {
+	return errors.As(j.err1, target)
```
ghemawat, line 66: Ditto.

runtime/codegen/errors_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+func TestJoinErrorsSupportsErrorsIs(t *testing.T) {
+	a := fmt.Errorf("a")
+	b := fmt.Errorf("b")
+	c := fmt.Errorf("c")
+	ab := JoinErrors(a, b)
+	if !errors.Is(ab, a) {
+		t.Error("JoinErrors(a, b) is not a.")
+	}
+	if !errors.Is(ab, b) {
```
ghemawat, line 60: I guess this passes because joinedErrors.Is checks against err1 and Unwrap returns true.

Seems round-about? or is this the recommended way before Go 1.20. (The doc from 1.20 says that the error model is a tree of errors and As/Is search through the entire tree.)

Should we just switch to 1.20?

weaver.go
```diff
@@ -42,21 +42,26 @@ import (
...
+//	    // foo.Foo executed properly and did not return an error.
 //	}
-var ErrRetriable = errors.New("retriable")
+//
+// Note that if a method call returns an error with an embedded SystemError, it
+// does NOT mean that the method never executed. The method may have executed
+// partially or fully. Thus, you must be careful retrying method calls that
+// result in a SystemError. Ensuring that all methods are either read-only or
+// idempotent is one way to ensure safe retries, for example.
+var SystemError = errors.New("Service Weaver system error")
```
ghemawat, line 64: Would CallError be a better name (it conveys that the error is scoped to a call). SystemError could perhaps overlap with some other future error where we are reporting a problem unrelated to any particular call.

Or perhaps weaver.RemoteCallError? Does this happen only for failed remote calls?

ghemawat (commented):
Thanks! I like the change. I suggest an alternative name in my review, but if you prefer SystemError, that is fine by me.

weaver.go
mwhittaker: Great suggestion. I went with `RemoteCallError`. And yup, these errors only happen for failed remote calls.

runtime/codegen/errors.go
mwhittaker: My understanding was `errors.Is` and `errors.As` call the `Is` and `As` method on every error returned by repeatedly calling `Unwrap`. I switched to go 1.20, as per your suggestion though, so I was able to delete this code :)

runtime/codegen/errors.go
mwhittaker: Deleted with the switch to go 1.20.

runtime/codegen/errors_test.go
mwhittaker: Switched to 1.20!

mwhittaker (commented):
Thanks Sanjay!

ghemawat (approved):
Great!

PR #242, opened 2023-04-09
website: health check

myyrakle:
When deploying GKE, I don't see the documentation saying that the health check API is required.

I guess I should add an explanation.

### DIFF
--- a/website/docs.md
+++ b/website/docs.md
@@ -1890,6 +1890,18 @@ By default, all listeners are **private**, i.e., accessible only from the cloud
 project's internal network. In our example, we declare that the `hello` listener
 is public.
 
+In order to perform rolling update, an API to be used for health check is required.
+Implement a "GET /" API that returns status code 200 as follows:
+
+```
+    ...
+    // Health Check
+    http.HandleFunc("/", func(writer http.ResponseWriter, request *http.Request) {
+        fmt.Fprintf(writer, "OK")
+    })
+    ...
+```
+
 Deploy the application using `weaver gke deploy`:
 
 ```console
--- a/website/docs.md
+++ b/website/docs.md
@@ -1857,11 +1857,6 @@ to it. Go to [this page][gcloud_billing] to create a new billing account, and
 [this page][gcloud_billing_projects] to associate a billing account with your
 cloud project.
 
-<div hidden class="todo">
-TODO(mwhittaker): Explain how to set up a `/healthz` endpoint once we
-finalize that design.
-</div>
-
 ## Getting Started
 
 Consider again the "Hello, World!" Service Weaver application from the [Step by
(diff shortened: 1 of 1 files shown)


website/docs.md
```diff
@@ -1890,6 +1890,18 @@ By default, all listeners are **private**, i.e., accessible only from the cloud
 project's internal network. In our example, we declare that the `hello` listener
 is public.
 
+In order to perform rolling update, an API to be used for health check is required.
```
spetrovic77, line 1893: Could you rewrite this leading paragraph as follows:

```
All listeners deployed to GKE are configured to be health-checked by GKE
load-balancers on the `/healthz` HTTP path. To make the `hello` application
compatible with these health checks, augment the application code to have
the HTTP server return status `200` on the `/healthz` endpoint:
```

website/docs.md
```diff
@@ -1890,6 +1890,18 @@ By default, all listeners are **private**, i.e., accessible only from the cloud
 project's internal network. In our example, we declare that the `hello` listener
 is public.
```
spetrovic77, line 1892: Thanks for bringing this up, we forgot to add a section on health-checks (see TODO above). Could you delete that TODO?

website/docs.md
```diff
@@ -1890,6 +1890,18 @@ By default, all listeners are **private**, i.e., accessible only from the cloud
...
 
+In order to perform rolling update, an API to be used for health check is required.
+Implement a "GET /" API that returns status code 200 as follows:
+
+```
+    ...
+    // Health Check
+    http.HandleFunc("/", func(writer http.ResponseWriter, request *http.Request) {
```
spetrovic77, line 1899: Can you replace "/" with "/healthz"?

(Note: this works because "/" will automatically handle all paths.)

website/docs.md
```diff
@@ -1890,6 +1890,18 @@ By default, all listeners are **private**, i.e., accessible only from the cloud
...
+        fmt.Fprintf(writer, "OK")
+    })
+    ...
```
spetrovic77, line 1902: Could you please paste the beginning of the existing "/hello" handler here, i.e.:

```
// Serve the /hello endpoint.
http.HandleFunc("/hello", func(w http.ResponseWriter, r *http.Request) {
...
```
```

spetrovic77 (commented):
Thanks for bringing this up and fixing!!

website/docs.md
myyrakle: done

website/docs.md
myyrakle: deleted

website/docs.md
myyrakle: I took the code from the "Listener" section and merged it.

PR #246, opened 2023-04-10
Updated metrics package documentation.

mwhittaker:
This PR adds documentation to the metrics package. This text is borrowed heavily from the docs on the website. I also fixed some broken code in the docs that were written before the metrics package was moved out of the core weaver package.

### DIFF
--- a/metrics/doc.go
+++ b/metrics/doc.go
@@ -1,0 +1,143 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


metrics/doc.go
```diff
@@ -0,0 +1,129 @@
...
+//			"An example gauge",
+//		)
+//		exampleHistogram = metrics.NewHistogram(
+//			"example_histogram",
+//			"An example histogram",
+//			[]float64{1, 10, 100, 1000, 10000},
+//		)
+//	)
+//
+// [NonNegativeBuckets] returns a default set of non-negative buckets. It's
```
spetrovic77, line 39: nit: s/it's/it is/

spetrovic77 (approved):
Very nice!

metrics/doc.go
mwhittaker: Done.

PR #247, opened 2023-04-10
Fixed perfetto stale colocation group bug.

mwhittaker:
Before this PR, we arranged for Perfetto to group traces according to colocation group and colocation group replica. However, we no longer have the notion of colocation groups or replicas. As a result, the Perfetto code was silently treating every event as if it had come from group "" replica "".

This PR fixes the bug by using weavelet ids instead of colocation groups. This uglifies the UI a bit, but in a future CL, we can pick something prettier.

9gvJBeDhkaKmbBc

I also simplified the Perfetto DB implementation by removing an unneeded table from the SQLite database. Before, the code had two tables:

```sql
CREATE TABLE IF NOT EXISTS replica_num (
    app TEXT NOT NULL,
    version TEXT NOT NULL,
    cgroup TEXT NOT NULL,
    cgroup_replica_id TEXT NOT NULL,
    num INTEGER NOT NULL,
    PRIMARY KEY(app,version,cgroup,cgroup_replica_id)
);

--- Next replica number for a colocation group.
CREATE TABLE IF NOT EXISTS next_replica_num (
    app TEXT NOT NULL,
    version TEXT NOT NULL,
    cgroup TEXT NOT NULL,
    next INTEGER NOT NULL,
    PRIMARY KEY(app,version,cgroup)
);
```

`replica_num` stored the replica number of every replica. `next_replica_num` was used to compute the next replica number. The `next_replica_num` table is unneeded, as we can compute the next replica number from the `replica_num table` directly.


spetrovic77 (approved):
Nice fix!

PR #250, opened 2023-04-11
update slog dependency

rgrandl:


weavelet.go
```diff
@@ -455,19 +455,18 @@ func (w *weavelet) UpdateComponents(req *protos.UpdateComponentsRequest) (*proto
...
-	// TODO(rgrandl): After we switch to slog, call With here to avoid
-	// repeating the same attributes again and again.
-	attrs := []any{
+	logger := w.env.SystemLogger().With([]any{
 		"component", req.RoutingInfo.Component,
 		"local", req.RoutingInfo.Local,
 		"replicas", req.RoutingInfo.Replicas,
-	}
-	w.env.SystemLogger().Debug("Updating routing info...", attrs...)
+	}...)
```
mwhittaker, line 462: Can we remote the `[]any` part and just do the following?

```go
logger := w.env.SystemLogger().With(
	"component", req.RoutingInfo.Component,
	"local", req.RoutingInfo.Local,
	"replicas", req.RoutingInfo.Replicas,
)
```

mwhittaker (changes requested):
Since the `Error` method changed from 

```
func (l *Logger) Error(msg string, err error, args ...any)
```

to 

```
func (l *Logger) Error(msg string, args ...any)
```

Don't we have to update all our calls to `Error` to add an explicit `"err"` label before the error?

weavelet.go
rgrandl: Good point.

mwhittaker (commented):
Looks great! Thanks Robert!

PR #251, opened 2023-04-11
Added basic telemetry to wrap app.

mwhittaker:
This PR adds two histograms to the wrap app and enables tracing. We'll show off this telemetry in the demo.

### DIFF
--- a/examples/wrap/main.go
+++ b/examples/wrap/main.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 
 	"github.com/ServiceWeaver/weaver"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 )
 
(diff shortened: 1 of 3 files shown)


examples/wrap/main.go
```diff
@@ -107,5 +108,6 @@ func main() {
 		fmt.Fprintf(w, wrapped)
 	})
 	handler := weaver.InstrumentHandler("hello", &mux)
+	handler = otelhttp.NewHandler(handler, "http")
```
rgrandl, line 111: i find it a bit weird to define `handler` and then override it with a different type of handler. 

Maybne either inline line 110 in line 111, or have different names for handlers.

rgrandl (approved):

examples/wrap/main.go
mwhittaker: Done.

PR #252, opened 2023-04-11
Added reverser example app, derived from wrap.

mwhittaker:
## Overview

In PRs #235, #236, and #237, I added an example wrap app that wrapped text. Our plan was to use this app in a demo, explaining how to write and deploy applications. Additionally, we would host an instance of the app on GKE and let people play with it.

After implementing the app, we realized that it was a bit confusing. The text you wrote was getting wrapped naturally by the browser. The text returned by the Wrapper component was also getting wrapped, but in a different way that was imperceptible unless you understood what was going on. Plus, the implementation got a little too complicated.

This PR introduces a much simpler reverser app. It borrows the same UI from the wrap app but reverses text instead of wrapping it, which is easier to understand. The implementation of reversing is also much simpler than wrapping.

We'll continue to iterate on the example app. If we like this one, we can delete the wrap app in the future.

## Demo

[Reverser Demo.webm](https://user-images.githubusercontent.com/3654277/231269714-3dc23b05-974a-4bee-b441-0041a03020ac.webm)


examples/reverser/main.go
```diff
@@ -0,0 +1,72 @@
+// Copyright 2022 Google LLC
```
rgrandl, line 1: nit: 2023

rgrandl (approved):
Looks nice. Quick question: we also have the reverser app under examples/hello. Should we remove the duplicate code? Maybe make the tutorial in the docs point to this reverser app instead?

examples/reverser/main.go
mwhittaker: Good catch. Done.

mwhittaker:
> Looks nice. Quick question: we also have the reverser app under examples/hello. Should we remove the duplicate code? Maybe make the tutorial in the docs point to this reverser app instead?

It's a bit annoying because the tutorial app has a `Cache` component to demonstrate how a component can call another, but I didn't want to have a `Cache` component in this demo to keep it simple. I'll submit for now to get the app running on GKE and then we can figure out later how best to deduplicate.

PR #254, opened 2023-04-13
Update the reverser app for running on GKE.

spetrovic77:
Namely:
  * Add the /healthz endpoint.
  * Add the public listener to the config.


mwhittaker (approved):
Do we need to update `examples/reverser/main.go` as well?

PR #260, opened 2023-04-17
Delete perfetto.db files on weaver purge

nipun-sehrawat:
Changes in the schema of perfetto database can result in application crashes such as the following:

`Application bankofanthos error: SQL logic error: no such column: weavelet_id (1)`

These crashes happen because of a schema mismatch between an existing perfetto.db file and that of the running application.  This change thus expands `weaver purge` command to delete the perfetto.db files as well.

### DIFF
--- a/runtime/perfetto/db.go
+++ b/runtime/perfetto/db.go
@@ -96,18 +96,18 @@ type DB struct {
 }
 
 // DatabaseFilePath returns the path of the file that stores the trace database.
-func DatabaseFilePath() (string, error) {
+func DatabaseFilePath(mode string) (string, error) {
 	dataDir, err := files.DefaultDataDir()
 	if err != nil {
 		return "", err
 	}
-	// TODO(mwhittaker): Use a different db for every deployer.
-	return filepath.Join(dataDir, "perfetto.db"), nil
+	return filepath.Join(dataDir, fmt.Sprintf("perfetto_%s.db", mode)), nil
 }
 
-// Open opens the default trace database on the local machine.
-func Open(ctx context.Context) (*DB, error) {
-	fname, err := DatabaseFilePath()
+// Open opens the default trace database on the local machine. mode specifies
+// the weaver mode (e.g. single, multi, ssh) for which the database is being open.
+func Open(ctx context.Context, mode string) (*DB, error) {
+	fname, err := DatabaseFilePath(mode)
 	if err != nil {
 		return nil, err
 	}
--- a/runtime/perfetto/db.go
+++ b/runtime/perfetto/db.go
@@ -95,16 +95,22 @@ type DB struct {
 	replicaNumCache *lru.Cache[replicaCacheKey, int]
 }
 
+// DatabaseFilePath returns the path of the file that stores the trace database.
+func DatabaseFilePath() (string, error) {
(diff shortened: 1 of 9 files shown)


runtime/perfetto/db.go
```diff
@@ -95,16 +95,22 @@ type DB struct {
 	replicaNumCache *lru.Cache[replicaCacheKey, int]
 }
 
+// DatabaseFilePath returns the path of the file that stores the trace database.
+func DatabaseFilePath() (string, error) {
+	dataDir, err := files.DefaultDataDir()
+	if err != nil {
+		return "", err
+	}
+	// TODO(mwhittaker): Use a different db for every deployer.
```
mwhittaker, line 104: Let me know if you want to tackle this in a future PR! Right now, `weaver single purge` deletes the perfetto db used by `weaver multi` (and vice versa), which is a little odd. I'm happy to work on it too, but I'll give you first dibs :)

mwhittaker (approved):
Awesome, thanks Nipun!

mwhittaker:
Heads up you'll have to run `dev/build_and_test generate` and push the changes to make the `generate` GitHub action happy.

runtime/perfetto/db.go
nipun-sehrawat: Done.  Although I am not happy with the strings constants I sprinkled around.  It would've helped if there was a type for specifying _mode_ of weaver tool but I didn't want to make that change in context of this CL.  Let me know if you have any suggestions.

runtime/perfetto/db.go
```diff
@@ -95,16 +95,22 @@ type DB struct {
...
 	}
+	return filepath.Join(dataDir, fmt.Sprintf("perfetto_%s.db", mode)), nil
+}
 
-	// TODO(mwhittaker): Use a different db for every deployer. Have the purge
-	// commands delete the db.
-	fname := filepath.Join(dataDir, "perfetto.db")
+// Open opens the default trace database on the local machine. mode specifies
+// the weaver mode (e.g. single, multi, ssh) for which the database is being open.
+func Open(ctx context.Context, mode string) (*DB, error) {
```
mwhittaker, line 109: > Done. Although I am not happy with the strings constants I sprinkled around. It would've helped if there was a type for specifying mode of weaver tool but I didn't want to make that change in context of this CL. Let me know if you have any suggestions.

This looks great to me! An alternative would be to change `Open` to receive the filename of the database: 

```go
func Open(ctx context.Context, filename string) (*DB, error)  
```

and then delete `DatabaseFilePath`. Then, every deployer's package (i.e. single, multi, ssh) defines a var or function that returns the filename of their perfetto database. `DashboardSpec` can either take this filename or the `perfetto.DB` directly.

I like your approach and this approach, both, so don't feel compelled to make any changes. I think we'll eventually want to make a bigger change later anyway to avoid having to purge everything all the time.

mwhittaker (approved):
Looks great Nipun!

runtime/perfetto/db.go
nipun-sehrawat: Thanks for the suggestion. Let me stick with what I have currently and we can make appropriate changes as we improve our tool versioning story in the future.

PR #263, opened 2023-04-18
Suppress golangci-lint false positives with Go tip.

ghemawat:
golangci-lint currently gives a couple of false positives (no such field or method) errors when built with an upcoming Go release. Suppress these errors until the next Go release and the corresponding golangci-lint fix.


mwhittaker:
Thanks Sanjay! I think because the linter is working for Go 1.20, the nolintlint linter reports the `nolint:typecheck` annotations as unused, so I think we have to `nolint:nolintlint`?

mwhittaker (approved):

PR #266, opened 2023-04-18
Generator Cleanups

mwhittaker:
## `weaver generate` Refresher
 
`weaver generate` has two phases. In the first phase, it analyzes and validates the provided code. For example:
    
- It finds every component interface and implementation.
- It checks that the implementation implements the interface and that all method arguments and returns are serializable.
- It finds every router type and checks that the router's methods match the component's methods.
    
In the second phase, it generates code and places it in weaver_gen.go files.
    
## Changes
    
Before this PR, the `generator` type was responsible for both validation and generation. This PR changes validation to use pure functions and makes `generator` responsible solely for generation. I also made a bunch of miscellaneous cleanups.
    
Now that validation is performed as pure functions, the flow of errors is more explicit. Before, all errors were placed in `generator.errors`, which made it a bit confusing to what was failing. For example, code generation almost never fails. I don't think this was clear before.
    
A question to reviewers, would you prefer we place all the validation code in a separate file? Right now, generator.go is pretty long (2100+ lines).

### DIFF
--- a/cmd/weaver/main.go
+++ b/cmd/weaver/main.go
@@ -74,7 +74,7 @@ func main() {
 		}
 		generateFlags.Parse(flag.Args()[1:]) //nolint:errcheck // does os.Exit on error
 		if err := generate.Generate(".", flag.Args()[1:]); err != nil {
-			fmt.Fprint(os.Stderr, err)
+			fmt.Fprintln(os.Stderr, err)
 			os.Exit(1)
 		}
 		return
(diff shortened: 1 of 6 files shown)


spetrovic77 (approved):
Thanks, looks great. Few possible followup-changes:

1. Move the validation function to a separate file.
2. Create a type (e.g., validator) that hosts these functions.

Issue #270, opened 2023-04-19
Component initialization for subscriptions in Pub/Sub scenario


fkorotkov:
In my Pub/Sub scenario I'd like to have components that doesn't have any methods but they only subscribe to bunch of topics in `Init`. These components do call other components inside subscriptions thought.

As far as I understand this component won't be initialized until I manually get an instance of it. I probably can initialize it in `main` but I have concerns when I have multiple of such subscribe-only components.

Let's say I have 3 different components `Foo`, `Bar` and `Baz` that only subscribe to topics. If I init each one of them in `main` how does splitting will work? Does it mean that every instance of my app on different machines will init `Foo`, `Bar` and `Baz` components and subscribe to all the topics.

fkorotkov:
I guess my question is still unanswered. Seems with the current architecture, in order to support Pub/Sub pattern, one will need to register all the subscribers in the `root`, therefore making every weavelet to handle all the subscriptions for the entire app which seems a bit resource wasteful.

If there would be a way to tell that a particular component always need to have at least one replica even if no one called `Get` and there will be a `Deinit` method (#275), then similar Pub/Sub scenarios will be covered.

Just want to share a use case so you might consider supporting it. 🙌

mwhittaker:
> I guess my question is still unanswered. Seems with the current architecture, in order to support Pub/Sub pattern, one will need to register all the subscribers in the root, therefore making every weavelet to handle all the subscriptions for the entire app which seems a bit resource wasteful.

Ah, I think I see the confusion. Calling `weaver.Get[T]` on some component `T` does not start a local instance of `T`. It ensures that some replicas of `T` are running somewhere. The exact location of where they run is determined by the deployer. So, you are free to call `weaver.Get` for all of your subscriber components. This won't make every weavelet run an instance of the components. Does that make sense?

fkorotkov:
Interesting! I'll need to run some experiments but I think this answers my question. Thank you!

mwhittaker:
Awesome! Please re-open the issue if you have any more questions!

PR #272, opened 2023-04-20
Allow empty component interfaces.

mwhittaker:
This PR changes `weaver generate` to allow empty component interfaces. We're in the process of converting main to be an explicit component that implements an empty `weaver.Main` interface. This change will allow that. We've also received [an issue][issue] explaining a use case for an interface without any methods.

[issue]: https://github.com/ServiceWeaver/weaver/issues/270

### DIFF
--- a/internal/tool/generate/testdata/missing_methods.go
+++ b/internal/tool/generate/testdata/missing_methods.go
@@ -12,14 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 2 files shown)


internal/tool/generate/testdata/missing_methods.go
mwhittaker, line 1: Note that this file moved from `testdata/errors` to `testdata` because it no longer produces errors.

spetrovic77 (approved):

PR #273, opened 2023-04-20
Simplified weaver generate's component type.

mwhittaker:
Before this PR, `weaver generate` had the following type to represent components:

```go
type component struct {
    name          string           // component interface name
    pos           token.Pos        // Location of component implementation
    fullName      string           // package-prefixed component interface name
    implName      string           // name of the component implementation type
    intf          *types.Interface // component's interface type
    file          *ast.File        // file that contains component's implementation
    methods       []*types.Func    // exported component methods, in deterministic order
    router        *types.Named     // router type for the component, or nil if there is no router.
    hasConfig     bool             // True iff implementation contains a weaver.WithConfig field.
    routingKey    types.Type       // routing key, or nil if there is no router.
    routedMethods map[string]bool  // the set of methods with a routing function
}
```

Many of these fields were either easily derivable from other fields (e.g., `name`, `fullName`, `implName`) or just straight up unused (e.g., `pos`, `file`). This PR simplifies the type to the following.

```go
type component struct {
    intf          *types.Named    // component interface
    impl          *types.Named    // component implementation
    router        *types.Named    // router, or nil if there is no router
    routingKey    types.Type      // routing key, or nil if there is no router
    routedMethods map[string]bool // the set of methods with a routing function
    hasConfig     bool            // implementation embeds weaver.WithConfig?
}
```

In making this change, I also cleaned up a couple of small things here and there (e.g., imprecise `token.Pos`'s in errors, some returns that should have been continues).

Note that some `weaver_gen.go` files changed because I simplified (and made more deterministic I think) the order in which methods are processed. None of the generated code should be different, just shuffled around.

### DIFF
--- a/examples/chat/weaver_gen.go
+++ b/examples/chat/weaver_gen.go
@@ -53,7 +53,7 @@ func init() {
 			return sQLStore_local_stub{impl: impl.(SQLStore), tracer: tracer}
 		},
 		ClientStubFn: func(stub codegen.Stub, caller string) any {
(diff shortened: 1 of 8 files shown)


spetrovic77 (approved):
Nice simplification!

PR #277, opened 2023-04-20
Fixed duplicate healthz registration in hello app.

### LINKED ISSUE #276: Cannot run examples/hello out of the box
Essentially I did
```zsh
git clone git@github.com:ServiceWeaver/weaver.git
cd weaver/examples/hello
go mod init hello
go mod tidy
weaver generate .
go run .
```
and got output
```
hello listener available on 127.0.0.1:12345
panic: http: multiple registrations for /healthz
```
followed by a stack trace pointing [here](https://github.com/ServiceWeaver/weaver/blob/main/examples/hello/main.go#L57).

Have I missed a step? Is it not expected that I could run the example like that? Where is that prior call to register `/healthz` happening?

mwhittaker:
Fixes #276.

### DIFF
--- a/examples/hello/main.go
+++ b/examples/hello/main.go
@@ -52,11 +52,5 @@ func main() {
 			}
 			fmt.Fprintf(w, "Hello, %s!\n", reversed)
 		}))
-
-	// Serve the /healthz endpoint.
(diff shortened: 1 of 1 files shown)


spetrovic77 (approved):
Just for posterity: good to delete this and we will change the healthz registration in the near future.

PR #278, opened 2023-04-20
Experimenting with weaver generate templates.

mwhittaker:
This PR experiments with `weaver generate` using templates to generate code. If we think it's more readable, I can covert the whole generator. Otherwise, I'll abandon the change.


spetrovic77 (commented):
LGTM

I would ask @ghemawat as well, since he intentionally chose not to use templates.

ghemawat (commented):
It's plausible. Though the main thing to watch out for is how much literal text there in the template as a fraction of the template size. If literal fraction is low, we are not getting much mileage out of templates (and have to deal with template's idiosyncratic structures instead of regular Go).

And when generating code, if literal fraction is high, perhaps that is an indication that there are missing abstractions in the codegen package?

Try to translate some of the harder code (maybe decoder)? And focus on the diff in that part to see if it looks better or not.

mwhittaker:
Thanks for the explanation Sanjay! I played around with converting `generateEncDecMethods` to use templates, and I think it makes the code worse, so I'll abandon this change. 

Another metric I looked at, besides the fraction of literal text to template size, is the amount of repetition in the template. The code  I templatized in the PR has a lot of repetition&mdash;the component name is used again and again&mdash;which makes the template more useful. Most of other generated code has less repetition, making the templates less useful.

ghemawat:
I wonder if we can clean up the generated code a bit by using a generic
codegen.Register function.

mwhittaker:
Oh, interesting idea. Let me look into that.

PR #281, opened 2023-04-20
Export healthz URL path and default handler.

spetrovic77:
Also, change the URL path to "/debug/weaver/healthz", to avoid interference with application handlers.

Based on @ghemawat's suggestions.

### DIFF
--- a/weaver.go
+++ b/weaver.go
@@ -31,6 +31,7 @@ import (
 	"net/http"
 	"os"
 	"reflect"
+	"sync"
 
 	"github.com/ServiceWeaver/weaver/runtime/codegen"
 	"go.opentelemetry.io/otel/trace"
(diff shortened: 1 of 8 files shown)


weaver.go
```diff
@@ -85,10 +110,8 @@ func init() {
 		ServerStubFn: func(any, func(uint64, float64)) codegen.Server { return nil },
 	})
 
-	// Add a trivial /healthz handler to the default mux.
-	http.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
-		w.Write([]byte("ok")) //nolint:errcheck // response write error
-	})
+	// Register HealthzHandler in the default ServerMux.
+	http.HandleFunc(HealthzURL, HealthzHandler)
```
ghemawat, line 114: I think we should move this into weaver.Init so that no changes are made if the user just happened to link in this package.

But we will some protection to make sure we only do it once (at least with weavertest, weaver.Init can be called more than once.)

website/docs.md
```diff
@@ -1879,67 +1879,18 @@ public_listener = [
...
-
-    // Serve Health Check endpoint.
-    http.HandleFunc("/healthz", func(writer http.ResponseWriter, request *http.Request) {
-        fmt.Fprintf(writer, "OK")
-    })
-
-    http.Serve(lis, nil)
-}
-```
+load-balancers on the `weaver.HealthzURL` URL path. ServiceWeaver automatically
```
ghemawat, line 1891: I think we should write out the actual URL path literal here as well.

ghemawat (changes requested):
Thanks.

weaver.go
spetrovic77: Good point, done.

ghemawat (approved):

Issue #224, opened 2023-04-05
Generate files from another package


jpearll:
Hi, I am pretty new to this project but found it very interesting!

I am working on building a hexagonal architecture structure in Go, and one of the thing I am doing is separating the ports (interface) from their corresponding concrete implementations adapters.

I was looking to compose my adapter with the weaver.Implements[T], where T is the port (interface, abstraction) which reside on another package.

There is this limitation now, of requiring the interface in the same package as the concrete implementation (when using the weaver CLI). For me that goes against the coupling and segregation concerns I want to enforce with this type of architecture. 

I was wondering if that limitation be removed in the future.

Thanks

mwhittaker:
Hi @jpearll! Thanks so much for checking out the project. You're totally correct that right now we require a component interface and component implementation to be in the same package. We elaborate on this more in #98. I'll close this issue as a duplicate of #98, if that's okay with you. We'll definitely re-examine the limitation and see if it can be removed. Feel free to re-open if you have any other questions :)

Issue #98, opened 2023-03-04
Interface contract outside of the struct implementation


ghazimuharam:
Hi everyone, thank you for introducing the new way of service implementation.

I'm trying to implement a POC for the service weaver and I stumbled upon this error
`internal/domain/reverser/reverser.go:16:6: weaver.Implements argument contract.Reverser is a type outside the current package.`

The thing is, implementing the contract in a separate package could be a better solution, well at least for the implementation I'm tryna made.

![image](https://user-images.githubusercontent.com/22569688/222914421-bea3e61c-243c-4548-a888-56b908c853d4.png)

I'm just wondering, is there any specific case of this rule that prevented the generator to use the interface contract outside of the implementation package? 

```
// Implementation of the Reverser component.
type reverser struct {
	weaver.Implements[contract.Reverser]
}
```

spetrovic77:
Thanks for reporting. Yes, we made a conscious decision not to support implementations of out-of-package interfaces, at least for now. @mwhittaker will have more state on this.

mwhittaker:
Hi @ghazimuharam! Here's the code in `weaver generate` that disallows a struct from implementing a component interface outside its package:

https://github.com/ServiceWeaver/weaver/blob/35727ae307c392d27cadfbf5302d56a3e76977ab/internal/tool/generate/generator.go#L357-L360

As Srdjan said, this was a conscious decision. I don't remember exactly why we made this decision; it might have just been a simplifying assumption for our initial release. Let me think harder about whether there's any disadvantages to allowing a component implementation to be in a separate package from its component interface. If there's none, I can take a stab at relaxing the restriction!

ghazimuharam:
Thank you, @spetrovic77 and @mwhittaker, for responding to my issue. As this appears to be a result of conscious decisions, I understand that I may need to wait for further clarification. However, please do not hesitate to reach out if you require any help from me. In the meantime, I will try to familiarize myself with the entire repository.

jpearll:
Just to share that I've found the simple workaround to embed the external interface (from another package), to a local interface (same package) that is getting referenced by weaver.Implement. The code generation works fine.

mwhittaker:
We explored the idea of allowing component interfaces and implementations to be in separate packages in #282 but ultimately decided to disallow it. In #285, we're updating an error message to suggest @jpearll's workaround.

PR #282, opened 2023-04-21
Component interface, struct in different packages.

mwhittaker:
## Overview

Currently, we require that a component interface and its accompanying implementation be in the same package. We've received a number of requests on GitHub (#98, #224) and on our [Discord][discord] to allow the interface and implementation to be in different packages. For example, one user's application generates gRPC server interfaces in one package, and they want to implement these as components in a different package.

Here are four options for how to deal with this issue in increasing order of complexity. This PR implements the second option, but that doesn't mean I think it's the best option. It was quick to implement, and it helped me understand all the options.

## Option 1: Disallow It

This option is trivial. Don't change anything and continue requiring component interfaces and implementations to be in the same package.

## Option 2: Allow It (this PR)

Allow component interfaces and implementations to be in separate packages. This introduces two issues:

1. A developer may forget to link in a component implementation. Before, if you called `weaver.Get[T]`, you were sure to link in the package containing `T`, and therefore you would link in the corresponding implementation of `T`. With this option, `weaver.Get[T]` is not sufficient. A developer must also link in the package with the implementation of `T`.

2. A developer may link multiple implementations of a component. This is the opposite of the previous problem. Previously, we could detect duplicate implementations of the same interface at code generation time, but with this option, that becomes impossible. A user may generate code separately for two packages, each with an implementation of the same interface.

This option appears to have the advantage that it allows people to swap out different implementations of the same interface, which some users have requested. That's true, but it's not so straightforward. If the two components have different configs, for example, swapping is not so easy.

The benefit of this option is that it's simple, and if we encourage people to place their interfaces and implementations in the same package, they shouldn't run into any problems. Only advanced users, who opt to split their interfaces and implementations have the additional burden of remembering to link in one, and only one, implementation.

## Option 3: `weaver.Get[Impl]`

For this option, we change `weaver.Get`. Rather than writing `weaver.Get[Intf]` for some interface `Intf`, developers write `weaver.Get[Impl]` for some implementation type `Impl`. This makes it impossible for a user to forget to link in an implementation, though it still allows them to link multiple implementations.

Implementing this option is a bit tricky because `weaver.Get` still needs to return the interface type, even if it receives the implementation type. I think this is doable though. Here's a rough sketch: https://go.dev/play/p/2L7R6JibaNN. Note that this doesn't build on go 1.20, but it does build on go 1.21.

## Option 4: Impls Are Components

This option builds on Option 3. `weaver.Get` takes an implementation type. However, we take it a step further and make a radical shift in the way we view components. A component is no longer uniquely identified by its interface, but rather by its implementation. Even if two implementations implement the same interface, we view them as distinct components.

This option avoids forgetting to link a component and avoids double linking a component. It's also not so different from the way we had things way back in the day where we automatically generated interfaces from implementations.

[discord]: https://discord.gg/FzbQ3SM8R5


mwhittaker:
@spetrovic77, @ghemawat, feel free to leave feedback on the four options without digging into the details of this PR. If you want to see an example of this PR in action, it includes a new example app, `greet`, which demonstrates the key changes.

Personally, I prefer Option 1 or 2 because I'm not sure it's worth complicating things to support something that should be rare (interfaces and implementations being in separate packages).

ghemawat (commented):
Thanks for the carefully laid out options.

Option #1: the work-around is not hard: Somebody who wants to implement an interface in another package can just make a local type:

```
    type MyType otherpkg.Type
```

Since interfaces have structural typing, instances of MyType can be passed anywhere otherpkg.Type is expected. Would clear documentation in the error message caused by attempting to use a foreign interface fix the issue?

Option 2: makes it too easy to get conflicts at link time (e.g., two packages that work independently just fine, but if put into the same binary, things fail; no way to fix it without getting an author of one of the two packages to make a change, which might break compatibility for users of that package. 

Option 3: metrics, traces, config, etc. would be keyed by the implementation type name? Maybe that's fine. But I am not sure if this is better than #1.

Option 4: What would be the return type here? If we synthesize a component type, we run into old familiar issues:

1. Code generation was touchy; in particular code could not pass around the interface before code generator had run once to create the interface. And that's a chicken and egg problem.
2. We have to synthesize a name for the component type corresponding to the implementation.

mwhittaker:
Let's go with Option 1 then? I can open a new PR to update the error message. @spetrovic77, does this sound good to you as well?

spetrovic77:
> Let's go with Option 1 then? I can open a new PR to update the error message. @spetrovic77, does this sound good to you as well?

Yeah, this makes sense to me, and like Sanjay's said, the interface compatibility makes this an easy issue to fix.

PR #283, opened 2023-04-21
Improved status registry errors + input handling.

mwhittaker:
Recall that the `weaver single status` command shows the status of every active Service Weaver application run with `go run .`:

```console
$ weaver single status
╭──────────────────────────────────────────────────────╮
│ DEPLOYMENTS                                          │
├─────────┬──────────────────────────────────────┬─────┤
│ APP     │ DEPLOYMENT                           │ AGE │
├─────────┼──────────────────────────────────────┼─────┤
│ collatz │ 62824743-478a-455d-9563-ead17900301a │ 2s  │
╰─────────┴──────────────────────────────────────┴─────╯
...
```

To know the set of all active applications, `weaver single status` leverages a **registry**. A registry is a directory that contains JSON-formatted registration files, one per deployment. When a deployment is launched, a registration file is added to the directory. When the deployment is killed, the registration file is removed.

This PR makes a couple of small changes to the registry:

1. I expanded the error messages returned by a registry. I was getting some error messages like "unable to decode JSON" that were hard to understand because they didn't have enough context.

2. I changed the registry to ignore any files in the registry directory that aren't registration files. Sometimes, spurious files are created in the registry directory. For example, if you open a registration file in vim and it crashes, you might get a `.swp` file.

3. I updated some stale documentation.

4. I simplified the implementation of the `Get` method.

### DIFF
--- a/internal/status/registry.go
+++ b/internal/status/registry.go
@@ -37,7 +37,8 @@ import (
 // running applications.
 type Registry struct {
 	// A Registry stores registrations as files in a directory. Every
-	// registration r is stored in a JSON file called {r.DeploymentId}.json.
+	// registration r is stored in a JSON file called
+	// {r.DeploymentId}.registration.json.
 	//
 	// TODO(mwhittaker): Store as protos instead of JSON?
 	dir string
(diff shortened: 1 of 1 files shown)


internal/status/registry.go
```diff
@@ -105,59 +106,65 @@ func (r Registration) Rolodex() string {
...
 // Register adds a registration to the registry.
 func (r *Registry) Register(ctx context.Context, reg Registration) error {
 	bytes, err := json.Marshal(reg)
 	if err != nil {
-		return err
+		return fmt.Errorf("registry: encode %v: %w", reg, err)
 	}
-	filename := fmt.Sprintf("%s.json", reg.DeploymentId)
-	w := files.NewWriter(filepath.Join(r.dir, filename))
+	filename := fmt.Sprintf("%s.registration.json", reg.DeploymentId)
```
nipun-sehrawat, line 124: Does this change in filename require users to weaver purge the existing registrations?  Or will the existing registration files be simply ignored?

nipun-sehrawat (approved):
Nice!

internal/status/registry.go
mwhittaker: I think the existing registrations will be ignored, but they could purge too. I'm still thinking about how to improve the whole purge-on-every-update situation we have right now. This change doesn't fix that.

PR #290, opened 2023-04-25
Introduce main components.

ghemawat:
Previously, a weaver application was started by calling `weaver.Init` and placing the body of the application after the call. An application is now started by placing the body of the application in a main component, calling `weaver.Run` and passing it a function that accepts the main component as an argument.

```
// Old code
root := weaver.Init(ctx)
... application code ...

// New code
type app struct {
  weaver.Implements[weaver.Main]
  ...
}

weaver.Run(ctx, func(ctx context.Context, app *app) {
  ... application code ...
})
```

The motivations for this change:

1. Application setup code will now belong to a component. Previously non-trivial amounts of setup code could run in a non-component context. This code is now part of a component and can therefore use component features (e.g., configuration).
1. A future change will make the component graph (which component uses which component) explicit. Having a main component will allow analysis tools to find the root of the graph.

Details of changes:

1. Replace `weaver.Init` by `weaver.Run`.
1. Add a `weaver.Main` interface that a component implementation can implement (by embedding `weaver.Implements[weaver.Main]`.
1. Update examples to use `weaver.Run`.
1. Replace `weavertest.Init` by `weavertest.Run`.
1. Update tests to use `weavertest.Run` (diverge_test needed some significant changes).
1. Update documentation.
1. Update code generator to allow implementation of `weaver.Main` (which would previously have been disallowed because the interface type was in a separate package).
1. Also update code generator to allow components without methods (weaver.Main does not have any methods, and application components may also not have any methods if all of their functionality is triggered at initialization time.)
1. Add a private package used to expose some weaver internals to other Service Weaver packages.
1. Improve error message when multiple implementations are registered for the same component.


weavertest/multi.go
```diff
@@ -64,16 +68,16 @@ func initMultiProcess(ctx context.Context, t testing.TB, config string) weaver.I
...
 	exe, err := os.Executable()
 	if err != nil {
-		t.Fatalf("error fetching binary path: %v", err)
+		return nil, nil, fmt.Errorf("error fetching binary path: %v", err)
 	}
-	appConfig.Name = strings.ReplaceAll(t.Name(), "/", "_")
+	appConfig.Name = strings.ReplaceAll(name, "/", "_")
 	appConfig.Binary = exe
-	appConfig.Args = []string{"-test.run", regexp.QuoteMeta(t.Name())}
+	appConfig.Args = []string{"-test.run", "^" + regexp.QuoteMeta(name) + "$"}
```
mwhittaker, line 80: Good catch.

weavertest/deployer.go
```diff
@@ -182,52 +150,55 @@ func (d *deployer) Init(config string) weaver.Instance {
...
+			return err
+		})
+		if err := d.registerReplica(g, e.WeaveletInfo()); err != nil {
+			d.stopLocked(fmt.Errorf(`cannot register the replica for "main": %w`, err))
+		}
+	}()
+	return nil
 }
 
+// stoplocked stops the deployer.
```
mwhittaker, line 191: stop instead of stoplocked.

weavertest/internal/simple/simple_test.go
```diff
@@ -80,39 +71,32 @@ func TestTwoComponents(t *testing.T) {
...
-			}
+				if !reflect.DeepEqual(want, got) {
+					t.Fatalf("GetAll() = %v; expecting %v", got, want)
+				}
+			})
 		})
 	}
 }
 
+/*
```
mwhittaker, line 99: Should we delete this test? I'm not sure it's testing much.

mwhittaker (approved):
Very nice Sanjay! Really nice how you were able to leverage all of the existing code to make such a big change. I was expecting a lot more code churn. Looking forward to the implementation of `weaver.Reference` and the removal of `weaver.Get`!

weavertest/internal/simple/simple_test.go
ghemawat: It is nice to make sure that the Listener() method works. However I don't know how to get a handle to the underlying implementation, which is why it is commented out.

I will leave it commented out with an explanatory comment and maybe we can get back to it soon.

website/docs.md
```diff
@@ -151,30 +151,42 @@ import (
...
 }
 ```
 
-`weaver.Init(...)` initializes the Service Weaver application. It also returns a
-`weaver.Instance`, which we assign to `root`. We'll explain instances in more
-detail momentarily. `weaver.Get[Reverser](root)` returns an instance of the
-`Reverser` component. We invoke methods on the component like we would any
-regular interface. In this example, we call `reverser.Reverse`.
+`weaver.Run(...)` initializes and runs the Service Weaver application.  In
+particular, `weaver.Run` finds the main component creates it, and passes it to a
```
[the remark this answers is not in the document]
ghemawat: Done.

ghemawat (commented):
I really appreciate the quick and thorough review.

PR #291, opened 2023-04-25
Re-enable the listener/HTTP-server test in simple_test.go.

ghemawat:
Add a new Server component to simple.go that gets a listener and exports and HTTP server. Make a test create this component and test it by fetching from the HTTP server.


mwhittaker (approved):
Brilliant!

ghemawat:
Thanks. Added some linter fixes. Will submit when checks are done.

PR #292, opened 2023-04-25
Added `weaver generate` warning for mistyped Init.

mwhittaker:
A component's `Init` method is supposed to have type `func(context.Context) error`, but it's easy to forget to add a `context.Context` argument or `error` return. If a component's `Init` method has the wrong type, it is silently ignored. This can be very frustrating to debug.

This PR augments `weaver generate` to warn the user if they have a mistyped `Init` function.


spetrovic77 (approved):
Nice!

PR #293, opened 2023-04-25
Added reverser server weavertest.

mwhittaker:
I wanted to understand how a developer could test their HTTP servers using the new `weavertest.Run` API. Let me know if the approach in this PR looks reasonable or if there's a better way!

### DIFF
--- a/examples/reverser/server.go
+++ b/examples/reverser/server.go
@@ -1,0 +1,70 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+
+	"github.com/ServiceWeaver/weaver"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+type server struct {
+	weaver.Implements[weaver.Main]
+	reverser Reverser
+	lis      net.Listener
+}
+
+func (s *server) Init(context.Context) error {
+	// Get a client to the Reverser component.
+	var err error
+	s.reverser, err = weaver.Get[Reverser](s)
+	if err != nil {
+		return err
+	}
+
+	// Get a network listener.
+	opts := weaver.ListenerOptions{LocalAddress: *address}
+	s.lis, err = s.Listener("reverser", opts)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("hello listener available on %v\n", s.lis)
+	return nil
+}
+
+func serve(ctx context.Context, s *server) error {
+	// Serve HTTP traffic.
+	var mux http.ServeMux
+	mux.Handle("/", weaver.InstrumentHandlerFunc("reverser",
+		func(w http.ResponseWriter, r *http.Request) {
+			fmt.Fprint(w, indexHtml)
+		}))
+	mux.Handle("/reverse", weaver.InstrumentHandlerFunc("reverser",
+		func(w http.ResponseWriter, r *http.Request) {
+			reversed, err := s.reverser.Reverse(r.Context(), r.URL.Query().Get("s"))
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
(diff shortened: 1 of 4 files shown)


examples/reverser/server.go
```diff
@@ -0,0 +1,70 @@
...
+	// Get a client to the Reverser component.
+	var err error
+	s.reverser, err = weaver.Get[Reverser](s)
+	if err != nil {
+		return err
+	}
+
+	// Get a network listener.
+	opts := weaver.ListenerOptions{LocalAddress: *address}
+	s.lis, err = s.Listener("reverser", opts)
```
ghemawat, line 43: If listener is fetched in serve() below, the test will become simpler (it won't need to create a listener itself).

ghemawat (commented):
Looks plausible, but there are trade-offs.

Pro: simpler than my round-about approach in simple_test.go

Con: server is running as a component in non-test mode and as a non-component in test mode. Some issues that might cause:

1. If server needs other components, not sure how we will get them.
2. Logger may get a nil ptr dereferencing error?
3. When we inject component dependencies, such direct use of server will require application test code to fill in these dependencies. So we will have to provide a method on weaver.Reference[T] to do that, but that seems okay?
4. If we also inject listeners automatically, we will again need a way for the application test code to fill in the listener.

Ideally, we should see if we can get the server to run as a real component in weavertest and have weavertest give us a way to get at the listener address from the test body?

mwhittaker:
> Looks plausible, but there are trade-offs.
> 
> Pro: simpler than my round-about approach in simple_test.go
> 
> Con: server is running as a component in non-test mode and as a non-component in test mode. Some issues that might cause:
> 
> 1. If server needs other components, not sure how we will get them.
> 2. Logger may get a nil ptr dereferencing error?
> 3. When we inject component dependencies, such direct use of server will require application test code to fill in these dependencies. So we will have to provide a method on weaver.Reference[T] to do that, but that seems okay?
> 4. If we also inject listeners automatically, we will again need a way for the application test code to fill in the listener.
> 
> Ideally, we should see if we can get the server to run as a real component in weavertest and have weavertest give us a way to get at the listener address from the test body?

Great points. Here's a half-baked idea. What if `weavertest.Run` accepted either a component interface (like it does today) or a component implementation (like `weaver.Run`)? I'm imagining something like this: 

```go
func TestServer(t *testing.T) {
    // Get a component implementation. Can do white-box testing.
    weavertest.Run(t, weavertest.Options{}, func(server *server) {
        // ... 
    })
}

func TestReverser(t *testing.T) {
    // Get a component interface. Can do black-box testing.
    weavertest.Run(t, weavertest.Options{}, func(reverser Reverser) {
        // ... 
    })
}
```

ghemawat:
> Great points. Here's a half-baked idea. What if `weavertest.Run` accepted either a component interface (like it does today) or a component implementation (like `weaver.Run`)? I'm imagining something like this:
> 
> ```go
> func TestServer(t *testing.T) {
>     // Get a component implementation. Can do white-box testing.
>     weavertest.Run(t, weavertest.Options{}, func(server *server) {
>         // ... 
>     })
> }
> 
> func TestReverser(t *testing.T) {
>     // Get a component interface. Can do black-box testing.
>     weavertest.Run(t, weavertest.Options{}, func(reverser Reverser) {
>         // ... 
>     })
> }
> ```

Interesting idea. Makes it very easy to test both ways. One potential issue is that people who don't write weavertest.Run() calls every day may have a hard time remembering which one to use.

I was thinking that perhaps we could switch to always passing just the implementation object. But then users will not be able to test remote calls to it. But maybe that is fine (a user testing some type X is interested in testing X's implementation. not how the RPC system works)?  This will reduce the benefit of multi-process tests, but perhaps those in practice are only useful for integration tests and will be tested if X uses a different component Y.

mwhittaker:
> Interesting idea. Makes it very easy to test both ways. One potential issue is that people who don't write weavertest.Run() calls every day may have a hard time remembering which one to use.

Great point. And looking only at the signature of `weavertest.Run`, it's hard to remember. 

> I was thinking that perhaps we could switch to always passing just the implementation object. But then users will not be able to test remote calls to it. But maybe that is fine (a user testing some type X is interested in testing X's implementation. not how the RPC system works)? This will reduce the benefit of multi-process tests, but perhaps those in practice are only useful for integration tests and will be tested if X uses a different component Y.

Very interesting. It sounds promising to me. It's simple, it mirrors `weaver.Run`, it lets you do whitebox testing. One wrinkle we may have to solve is something like the following. The test includes both a client to the `B` component (`a.b`) and an implementation of the `B` component (`b`). I'm not sure if this is confusing or not. A call like `a.b.Foo()` may or may not be routed to `b`, which might surprise people? Not sure. I'll think more about it!

```go
type A interface { ... }
type a struct { 
    weaver.Implements[A] 
    b weaver.Reference[B]
}

type B interface { ... }
type b struct {
    weaver.Implements[B]
}

weavertest.Run(t, weavertest.Options{}, func(a *a, b *b) {
    a.b.Foo()
})
```

mwhittaker:
Let me also close this PR since the code has too many problems I think.

PR #296, opened 2023-04-26
Fixed `weaver multi` and `weaver ssh` main bug.

mwhittaker:
Before, `weaver multi` and `weaver ssh` assumed the name of the main component was "main". Recently, we switched it to
"github.com/ServiceWeaver/weaver/main". Both the meaver multi and weaver ssh deployers would instruct a weavelet to start "main", but no such component was registered, and the app would immediately crash.

### DIFF
--- a/internal/tool/multi/deployer.go
+++ b/internal/tool/multi/deployer.go
@@ -266,7 +266,7 @@ func (d *deployer) startColocationGroup(g *group) error {
 			Sections:      d.config.Sections,
 			SingleProcess: false,
 			SingleMachine: true,
-			RunMain:       g.components["main"],
+			RunMain:       g.components["github.com/ServiceWeaver/weaver/Main"],
 		}
 		e, err := envelope.NewEnvelope(d.ctx, info, d.config)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


internal/tool/multi/deployer.go
```diff
@@ -266,7 +266,7 @@ func (d *deployer) startColocationGroup(g *group) error {
 			Sections:      d.config.Sections,
 			SingleProcess: false,
 			SingleMachine: true,
-			RunMain:       g.components["main"],
+			RunMain:       g.components["github.com/ServiceWeaver/weaver/Main"],
```
ghemawat, line 269: Should we add a common constant somewhere?

ghemawat (commented):
Thanks for fixing!

I am a little worried that we don't have any tests that caught the problem. Could we somehow exercise the multi and ssh deployers in a test, maybe even a manually run recipe.

ghemawat (approved):

mwhittaker:
> Thanks for fixing!
> 
> I am a little worried that we don't have any tests that caught the problem. Could we somehow exercise the multi and ssh deployers in a test, maybe even a manually run recipe.

@dnephin is working on this in #287!

PR #301, opened 2023-04-27
Fix bug in metrics package's codegen.MethodLatencies.

ryan961:
Fix bug in metrics package's codegen.MethodLatencies, converting microseconds to milliseconds.

### DIFF
--- a/internal/metrics/stats.go
+++ b/internal/metrics/stats.go
@@ -183,7 +183,7 @@ func (s *StatsProcessor) getSnapshot(snapshot []*metrics.MetricSnapshot) {
 		case codegen.MethodBytesRequest.Name():
 			bucket.kbRecvd += m.Value / 1024 // B to KB
 		case codegen.MethodLatencies.Name():
-			bucket.latencyMs += m.Value / 1024 // us to ms
+			bucket.latencyMs += m.Value / 1000 // µs to ms
 
 			var count uint64
 			for idx := range m.Bounds {


internal/metrics/stats.go
```diff
@@ -183,7 +183,7 @@ func (s *StatsProcessor) getSnapshot(snapshot []*metrics.MetricSnapshot) {
 		case codegen.MethodBytesRequest.Name():
 			bucket.kbRecvd += m.Value / 1024 // B to KB
 		case codegen.MethodLatencies.Name():
-			bucket.latencyMs += m.Value / 1024 // us to ms
+			bucket.latencyMs += m.Value / 1000 // µs to ms
```
rgrandl, line 186: Nice, thanks for catching it.

.gitignore
```diff
@@ -5,3 +5,6 @@ examples/factors/factors
 examples/onlineboutique/onlineboutique
 examples/hello/hello
 website/public/
+
+.idea/
```
rgrandl, line 9: Can you please remove this? We don't allow adding things in .gitignore that only makes sense on your local machine.

rgrandl (changes requested):
Thanks for fixing this ryan.

.gitignore
dnephin: This is the approach I use to ignore editor files: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c (another reference: [github docs](https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer)).

It's a great approach because you don't have to ignore these files in every git repository. You set it up once for each machine, and it will ignore the files in all git repositories.

ryan961: Thank you for your suggestion, I will make the necessary changes and resubmit.

rgrandl (approved):

ryan961:
Hi, @rgrandl. I tried adding some unit test code and I'm considering whether or not to commit it. Also, I used the `github.com/stretchr/testify/assert` library in the package, and I'm not sure if there are any relevant standards or recommendations (such as importing third-party libraries...)

rgrandl:
Hi @ryan961, sorry I missed your comment because I closed the PR. 

We would prefer to not add any third party libraries unless they are really needed. Can you take a look at our tests to see how we compare the expected value and the output value? Usually we do cmp.Diff and check if the diff is as expected. I don't think you need to do assert. 

Otherwise, send in a separate PR the unit test you want to add, and we can review it.

ryan961:
> Hi @ryan961, sorry I missed your comment because I closed the PR.
> 
> We would prefer to not add any third party libraries unless they are really needed. Can you take a look at our tests to see how we compare the expected value and the output value? Usually we do cmp.Diff and check if the diff is as expected. I don't think you need to do assert.
> 
> Otherwise, send in a separate PR the unit test you want to add, and we can review it.

@rgrandl Thank you for your response. First of all, I really like the Weaver project and I would love to contribute to it (within my limited spare time). Before that, I have thoroughly read and studied the existing unit test code of the project. I understand that for a framework/library, third-party libraries are not introduced unless absolutely necessary (and even if they are introduced, it is done after careful consideration by the maintainers). The reason I raised the above question is that while attempting to write unit tests for https://github.com/ServiceWeaver/weaver/blob/main/internal/files/file.go, I used `assert.FileExists`, `assert.DirExists`, etc. to simplify the unit test code. Of course, I can directly implement the corresponding functions in the `file_test.go` file myself (it is possible that I am not aware of the existing implementation in the official library, so if there is one, I would appreciate being informed about it, thank you). I am concerned that other modules in the project may have similar functional requirements, which could lead to redundant code and increased workload in the future. Below is the unit test code I previously attempted, which used the assert library. In the future, I will modify it using the official standard library and attempt to submit a pull request.
```go
func TestWriter_Close(t *testing.T) {
	file := "test.txt"

	writer := NewWriter(file)
	err := writer.Close()
	require.NoError(t, err)
	assert.FileExists(t, file)
	assert.NoFileExists(t, writer.tmpName)

	os.Remove(file)
}

func TestDefaultDataDir(t *testing.T) {
	homeDir, _ := os.UserHomeDir()

	dir, err := DefaultDataDir()
	require.NoError(t, err)
	assert.Equal(t, dir, path.Join(homeDir, ".local/share/serviceweaver"))
	assert.DirExists(t, dir)
	os.Remove(dir)

	_ = os.Setenv(xdgDataHome, homeDir)
	dir, err = DefaultDataDir()
	require.NoError(t, err)
	assert.Equal(t, dir, path.Join(homeDir, "/serviceweaver"))
	assert.DirExists(t, dir)
	os.Remove(dir)
}
```

ryan961:
> > Hi @ryan961, sorry I missed your comment because I closed the PR.
> > We would prefer to not add any third party libraries unless they are really needed. Can you take a look at our tests to see how we compare the expected value and the output value? Usually we do cmp.Diff and check if the diff is as expected. I don't think you need to do assert.
> > Otherwise, send in a separate PR the unit test you want to add, and we can review it.
> 
> @rgrandl Thank you for your response. First of all, I really like the Weaver project and I would love to contribute to it (within my limited spare time). Before that, I have thoroughly read and studied the existing unit test code of the project. I understand that for a framework/library, third-party libraries are not introduced unless absolutely necessary (and even if they are introduced, it is done after careful consideration by the maintainers). The reason I raised the above question is that while attempting to write unit tests for https://github.com/ServiceWeaver/weaver/blob/main/internal/files/file.go, I used `assert.FileExists`, `assert.DirExists`, etc. to simplify the unit test code. Of course, I can directly implement the corresponding functions in the `file_test.go` file myself (it is possible that I am not aware of the existing implementation in the official library, so if there is one, I would appreciate being informed about it, thank you). I am concerned that other modules in the project may have similar functional requirements, which could lead to redundant code and increased workload in the future. Below is the unit test code I previously attempted, which used the assert library. In the future, I will modify it using the official standard library and attempt to submit a pull request.
> 
> ```go
> func TestWriter_Close(t *testing.T) {
> 	file := "test.txt"
> 
> 	writer := NewWriter(file)
> 	err := writer.Close()
> 	require.NoError(t, err)
> 	assert.FileExists(t, file)
> 	assert.NoFileExists(t, writer.tmpName)
> 
> 	os.Remove(file)
> }
> 
> func TestDefaultDataDir(t *testing.T) {
> 	homeDir, _ := os.UserHomeDir()
> 
> 	dir, err := DefaultDataDir()
> 	require.NoError(t, err)
> 	assert.Equal(t, dir, path.Join(homeDir, ".local/share/serviceweaver"))
> 	assert.DirExists(t, dir)
> 	os.Remove(dir)
> 
> 	_ = os.Setenv(xdgDataHome, homeDir)
> 	dir, err = DefaultDataDir()
> 	require.NoError(t, err)
> 	assert.Equal(t, dir, path.Join(homeDir, "/serviceweaver"))
> 	assert.DirExists(t, dir)
> 	os.Remove(dir)
> }
> ```

Hmm... I noticed that there have been significant changes to the code in the latest main branch regarding https://github.com/ServiceWeaver/weaver/blob/main/internal/files/file.go, and it seems that the content I mentioned earlier may not be applicable anymore. I will continue to follow the project and make some experimental pull request submissions. I hope to have more discussions with you under the PR regarding these changes.😊

PR #304, opened 2023-04-27
Remove weaver.Get (obsoleted by weaver.Ref).

ghemawat:
Remove the obsolete weaver.Get[T] API. Users should now declare weaver.Ref[T] fields in their component implementation struct instead.

Fixed all callers.

Tweaked examples/hello to more closely match the sample code in the documentation.


mwhittaker (approved):
`+92 -521` Nice! So long `weaver.Get`!

PR #305, opened 2023-04-27
Made weavelet instantiate main component directly.

mwhittaker:
Recall that when a weavelet is created, it receives an `EnvelopeInfo` proto from the envelope that contains a boolean `RunMain` field. When `RunMain` is true, the weavelet creates an instance of the main component and passes it to the user-written function passed to `weaver.Run`.

Before this PR, the weavelet created an instance of the main component by calling `getInstance`. `getInstance` returns either a local implementation of the component or a stub to a remote implementation. This PR changes the weavelet to instead call `getImpl` which always returns a local implementation.


ghemawat (approved):
Nice simplification. Thanks.

PR #308, opened 2023-04-28
Silent expected warning messages in generator_test.go.

ghemawat:
Add an Options argument to generate.Generate that contains a function used to report warnings. In generator_test.go fill this with a function that reports the warning via testing.T.Log.


mwhittaker (approved):
Thanks Sanjay!

PR #309, opened 2023-04-28
Changed deployers to persist data in one spot.

mwhittaker:
Recall that deployers often have to persist various data to the file system. The "weaver multi" deployer, for example, persists logs, registrations, and traces to files.

Before this PR, deployers would store logs in `/tmp` and everything else in `~/.local/share/serviceweaver`. Different deployers would sometimes store their data in the same directory. A `multi_perfetto.db` file would be right next to a `single_perfetto.db` file, for example.

This made purging data files annoying. It was like a scavenger hunt trying to track down the various places that data was stored. This PR changes deployers to store all their data in a single directly. For example, the "weaver multi" deployer now stores everything inside `~/.local/share/serviceweaver/multi`. The single process deployer stores everything inside `~/.local/share/serviceweaver/single`.

This makes purging much simpler. It also makes it clearer which files are being used by which deployers.

Note that this PR moves logs out of `/tmp`, which persists them longer. This means that logs won't be garbage collected automatically, but I think that's actually a good thing. We should implement a principled way of garbage collecting logs, rather than leaving it up to `/tmp`. I think it would be surprising and frustrating if an important log file was spuriously deleted by the OS.

### DIFF
--- a/internal/status/dashboard.go
+++ b/internal/status/dashboard.go
@@ -88,10 +88,10 @@ type Command struct {
 
 // DashboardSpec configures the command returned by DashboardCommand.
 type DashboardSpec struct {
-	Tool     string                                   // tool name (e.g., "weaver single")
-	Mode     string                                   // tool mode (e.g. "single", "multi", "ssh")
-	Registry func(context.Context) (*Registry, error) // registry of deployments
-	Commands func(deploymentId string) []Command      // commands for a deployment
+	Tool         string                                   // tool name (e.g., "weaver single")
+	PerfettoFile string                                   // perfetto database file
+	Registry     func(context.Context) (*Registry, error) // registry of deployments
+	Commands     func(deploymentId string) []Command      // commands for a deployment
 }
 
 // DashboardCommand returns a "dashboard" subcommand that serves a dashboard
@@ -133,7 +133,7 @@ Flags:
 			}
 			url := "http://" + lis.Addr().String()
 
-			traceDB, err := perfetto.Open(ctx, spec.Mode)
+			traceDB, err := perfetto.Open(ctx, spec.PerfettoFile)
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "cannot open Perfetto database: %v\n", err)
 			} else {
--- a/internal/tool/ssh/impl/manager.go
+++ b/internal/tool/ssh/impl/manager.go
@@ -27,8 +27,8 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/ServiceWeaver/weaver/internal/files"
 	imetrics "github.com/ServiceWeaver/weaver/internal/metrics"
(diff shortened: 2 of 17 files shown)


internal/tool/ssh/impl/manager.go
```diff
@@ -667,9 +675,5 @@ func serveHTTP(ctx context.Context, lis net.Listener, handler http.Handler) erro
 // $XDG_DATA_HOME/serviceweaver/ssh_registry, or
```
nipun-sehrawat, line 675: s/ssh_registry/ssh ?

internal/status/dashboard.go
```diff
@@ -88,10 +88,11 @@ type Command struct {
 
 // DashboardSpec configures the command returned by DashboardCommand.
 type DashboardSpec struct {
-	Tool     string                                   // tool name (e.g., "weaver single")
-	Mode     string                                   // tool mode (e.g. "single", "multi", "ssh")
-	Registry func(context.Context) (*Registry, error) // registry of deployments
-	Commands func(deploymentId string) []Command      // commands for a deployment
+	Tool         string                                   // tool name (e.g., "weaver single")
+	Mode         string                                   // tool mode (e.g. "single", "multi", "ssh")
```
nipun-sehrawat, line 92: Can we drop Mode now?

nipun-sehrawat (approved):
Nice simplification! 👍 

Regarding logs, users are likely to complain about logs taking up too much disk space in the future.  While we can leave it up to the users to clean up logs manually, using logrotate would be a good alternative, perhaps by exposing some log management related commands in weaver tool.  Nothing urgent but something to keep in mind.

internal/tool/ssh/impl/manager.go
```diff
@@ -67,6 +67,16 @@ const (
 	babysitterInfoKey = "SERVICEWEAVER_BABYSITTER_INFO"
 )
 
+var (
+	// The directories and files where "weaver multi" stores data.
```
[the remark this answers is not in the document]
mwhittaker: Oops, done.

internal/tool/ssh/impl/manager.go
mwhittaker: Done.

internal/status/dashboard.go
mwhittaker: Good catch.

mwhittaker (commented):
Thanks Nipun!

mwhittaker:
> Nice simplification! 👍
> 
> Regarding logs, users are likely to complain about logs taking up too much disk space in the future. While we can leave it up to the users to clean up logs manually, using logrotate would be a good alternative, perhaps by exposing some log management related commands in weaver tool. Nothing urgent but something to keep in mind.

Great point! I can think more about how to garbage collect logs.

PR #311, opened 2023-04-28
Add codegen method that returns the static call graph of components.

ghemawat:


runtime/codegen/registry.go
```diff
@@ -146,3 +147,35 @@ func ComponentConfigValidator(path, cfg string) error {
...
+type CallEdge struct {
+	Caller reflect.Type
+	Callee reflect.Type
+}
+
+// CallGraph returns the component call graph (as a list of CallEdge values).
+func CallGraph() []CallEdge {
+	var result []CallEdge
+	for _, reg := range Registered() {
+		impl := reflect.TypeOf(reg.New())
```
spetrovic77, line 163: Is creating a new impl instance always a safe/cheap thing to do? In my upcoming change, every weavelet will be calling this function on startup.

An alternative is to change the codegen code to return the impl reflect.Type directly.

spetrovic77 (approved):
Very clean!

runtime/codegen/registry.go
ghemawat: I was doing that in this change but it brought in a bunch of unrelated changes. How about I make the switch to that in another change?

PR #312, opened 2023-04-28
Change component registration to hold implementation type.

ghemawat:
Previously, codegen.Registration contained a function that created the implementation object and returned a pointer to it. We now store the reflect.Type for the implementation object instead.  This allows us to examine type properties without having to create an unnecessary object.


spetrovic77 (approved):
Looks great!

PR #313, opened 2023-05-01
Use the call graph to authorize peers in the multi deployer.

spetrovic77:
In this change:
      1. The multi deployer, when constructing colocation groups, also computes
        the set of components each colocation group is allowed to call methods
        on.
      2. When the server receives a new client connection, it calls
        `VerifyClientCertificate` on the multi deployer. The multi deployer
        extracts the colocation group name from the client certificate and
        returns the pre-computed set of components (see step 1 above) the client's
        colocation group is allowed to access.
      3. (Already implemented) The weavelet makes sure that a given connection
        only admits method calls on the allowlisted components (see step 2 above).
    
Other changes:
  * Verify WeaveletInfo inside envelope_conn. This removes a bunch of redundant code.


internal/envelope/conn/conn_test.go
```diff
@@ -182,3 +186,27 @@ func (*handlerForTest) VerifyClientCertificate(context.Context, *protos.VerifyCl
 func (*handlerForTest) VerifyServerCertificate(context.Context, *protos.VerifyServerCertificateRequest) (*protos.VerifyServerCertificateReply, error) {
 	panic("unused")
 }
+
+func register[Intf, Impl any](name string, configFn func(any) any) {
```
ghemawat, line 190: I think the callers always pass in a nil configFn. Maybe drop that argument?

internal/envelope/conn/envelope_conn.go
```diff
@@ -355,3 +354,37 @@ func (e *EnvelopeConn) rpc(request *protos.EnvelopeMsg) (*protos.WeaveletMsg, er
...
+	if err := checkVersion(wlet.Version); err != nil {
+		return err
+	}
+	return nil
+}
+
+// checkVersion checks that the deployer API version the deployer was built
+// with is compatible with the deployer API version the app was built with,
+// erroring out if they are not compatible.
+func checkVersion(appVersion *protos.SemVer) error {
```
ghemawat, line 376: Doesn't this overlap with the changes Michael is making? Should we be not mucking with version checking in this change? Or is this just code movement that will go away soon?

internal/tool/multi/deployer.go
```diff
@@ -198,6 +179,68 @@ func newDeployer(ctx context.Context, deploymentId string, config *protos.AppCon
...
+// components are placed in their own co-location group. We use the first
+// listed component as the co-location group name.
+func computeGroups(config *protos.AppConfig) (map[string]*group, error) {
+	groups := map[string]*group{}
+	ensureGroup := func(component string) *group {
+		if g, ok := groups[component]; ok {
+			return g
+		}
+		g := &group{
+			name:        component,
```
ghemawat, line 204: I wonder if we should deterministically pick one of the component names as the group name? I worry about map iteration non-determinism or something else causing group names to fluctuate over time.

But perhaps this is a pre-existing problem that should be fixed in another change?

internal/tool/multi/deployer.go
```diff
@@ -198,6 +179,68 @@ func newDeployer(ctx context.Context, deploymentId string, config *protos.AppCon
...
+	// Use the call graph information to (1) identify all components in the
+	// application binary and create groups for them, and (2) compute the
+	// set of components a given group is allowed to invoke methods on.
+	callGraph, err := bin.ReadComponentGraph(config.Binary)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read the call graph from the application binary: %w", err)
+	}
+	for _, edge := range callGraph {
+		src := edge[0]
+		dst := edge[1]
```
ghemawat, line 235: Should we change the edge from a [2]string to struct{Src, Dst string}?

runtime/version/version.go
```diff
@@ -59,7 +59,7 @@ const (
 var version string
```
ghemawat, line 59: Why is version a var instead of const?

internal/tool/multi/deployer.go
spetrovic77: I like it the way it is: simple.

runtime/version/version.go
spetrovic77: I believe @mwhittaker had to do it this way so that the compiler doesn't mark the
definition as unused and delete it from the read-only data.

ghemawat (approved):

PR #314, opened 2023-05-03
Moved logs back to /tmp/serviceweaver.

mwhittaker:
This PR reverts some of the behavior introduced in #309 by moving logs back to `/tmp`, rather than `~/.local/share/serviceweaver`. Thanks Sanjay for the recommendation.

### DIFF
--- a/runtime/dir.go
+++ b/runtime/dir.go
@@ -19,6 +19,16 @@ import (
 	"path/filepath"
 )
 
+// LogsDir returns the default directory for Service Weaver logs,
+// $DIR/tmp/serviceweaver/logs where $DIR is the default directory used for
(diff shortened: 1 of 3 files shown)


runtime/dir.go
```diff
@@ -19,6 +19,14 @@ import (
 	"path/filepath"
 )
 
+// LogsDir returns the default directory for Service Weaver logs,
+// /tmp/serviceweaver/logs. We recommend that deployers store their logs in a
```
spetrovic77, line 23: My understanding is that os.TempDir() isn't always /tmp/. Can you clarify it here?

spetrovic77 (approved):

PR #315, opened 2023-05-03
Removed -e flag from pkill in weaver purge.

mwhittaker:
The -e flag isn't supported on mac.


spetrovic77 (approved):
Sorry for the delay. Thanks for fixing!

PR #316, opened 2023-05-03
Add package to read component use graph from a Go binary.

ghemawat:
weaver generates emits edges as specially delimited strings in the binary. The bin package parses the binary and reads these edges out.

Added a test that builds a sample program for Linux, Windows, Mac, and checks that the expected edges can be read out of them.


runtime/codegen/graph.go
```diff
@@ -0,0 +1,70 @@
...
+		if a, b := result[i][0], result[j][0]; a != b {
+			return a < b
+		}
+		return result[i][1] < result[j][1]
+	})
+	return result
+}
+
+func checksum(src, dst string) string {
+	edge := fmt.Sprintf("wEaVeReDgE:%s→%s", src, dst)
```
spetrovic77, line 67: nit: if you remove the "wEaVeReDgE" prefix, the earlier comment would be true:

```
// checksum is the first 8 bytes of the hex encoding of the SHA-256 of the string "src→dst"
```

runtime/codegen/registry.go
```diff
@@ -62,6 +62,10 @@ type Registration struct {
 	LocalStubFn  func(impl any, tracer trace.Tracer) any
 	ClientStubFn func(stub Stub, caller string) any
 	ServerStubFn func(impl any, load func(key uint64, load float64)) Server
+
+	// String containing the result of MakeEdgeString(Name, Dst) for all
+	// components named Dst used by this component.
+	UseData string
```
spetrovic77, line 68: Might be clearer if renamed to "CallData"

internal/bin/bin.go
```diff
@@ -0,0 +1,104 @@
+// Copyright 2023 Google LLC
```
spetrovic77, line 1: This will be called from other deployers (i.e., GKE), so maybe it should go under `runtime/bin/`?

runtime/codegen/graph.go
```diff
@@ -0,0 +1,70 @@
...
+// ⟦checksum:wEaVeReDgE:src→dst⟧
+//
+// checksum is the first 8 bytes of the hex encoding of the SHA-256 of
+// the string "src→dst" and src and dst are the fully qualified
+// component type names.
+
+// MakeEdgeString returns a string that should be emitted into generated
+// code to represent an edge from src to dst.
+func MakeEdgeString(src, dst string) string {
+	return fmt.Sprintf("⟦%s:wEaVeReDgE:%s→%s⟧\n", checksum(src, dst), src, dst)
```
spetrovic77, line 39: I thought this would be hacky, but the addition of `wEaVeReDgE` makes it totally legit! :rofl:

spetrovic77 (approved):
This is so awesome!

runtime/codegen/registry.go
ghemawat: A might have weaver.Ref[B] but not call it.  I will rename it to RefData

internal/bin/bin.go
ghemawat: Done.

Issue #319, opened 2023-05-05
build failed in used latest exp


godcong:
```
gopath\pkg\mod\github.com\!service!weaver\weaver@v0.7.2\runtime\logging\logger.go:94:12: cannot use func(a slog.Attr) {…} (value of type func(a slog.Attr)) as func(slog.Attr) bool value in argument to rec.Attrs
```

The last `exp/slog` changed the function returns.

current version is: https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230419192730-864b3d6c5c2c/slog?tab=versions#:~:text=v0.0.0%2D20230321023759%2D10a507213a29
latest version is: https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230419192730-864b3d6c5c2c/slog?tab=versions#:~:text=v0.0.0%2D20230425010034%2D47ecfdc1ba53

PR #320, opened 2023-05-05
feat update exp

godcong:
update exp to latest and fixed the method.
relate #319


rgrandl (approved):
Thanks for updating slog to the latest version.

PR #321, opened 2023-05-05
[docs fix] Use query param as expected by the example

kars7e:
I've noticed that when following the docs and getting surprised by the result.

Current implementation uses static string, but subsequent examples expect the query param to be used and reversed instead.


rgrandl (changes requested):
Thanks for the fix. Would you mind updating the examples/hello/reverser.go code as well, so the docs and the code are in sync?

If you have time, maybe you can modify the hello app to prettify the text or something else, instead of using the reverser component which is a bit odd.

kars7e:
Updated the hello example.
IMHO Reverser makes sense here. It's simple, yet the result is immediately noticeable, so works fine as an example.

PR #322, opened 2023-05-08
Embed deployer API version in binary.

mwhittaker:
Recall that a weavelet communicates with a deployer over a pair of pipes using the deployer API. Every time the deployer API changes, we assign it a new version. When a deployer launches a Service Weaver binary, the deployer checks that the weavelet is running a compatible version.

This PR embeds the deployer API version in Service Weaver binaries, allowing a deployer to check the version of a binary without running it. In some cases, this is very convenient. For example, the "weaver gke" tool can check the version of a binary before submitting it to GKE. The GKE controller can also check the version without having to run potentially dangerous user code.

The mechanism to embed and extract versions is the same as the one in PR #316. A string of the form ⟦107c095e:wEaVeRvErSiOn:0.8.0⟧ is included in the generated code. The version.ReadVersion function parses this string from the RO data section of a binary.

In future PRs, I'll update the deployers to use the embedded versions.


spetrovic77 (commented):
Why do we need to embed the version inside the generated file? Couldn't we just define a constant once inside weaver.go?

mwhittaker:
> Why do we need to embed the version inside the generated file? Couldn't we just define a constant once inside weaver.go?

We discussed parts of this at the meeting today, but I'll reiterate it here for everyone. There's a couple of subtle things going on here. First, imagine I add the following code to the `version` package or the generated code:

```go
const version = "[wEaVeRvErSiOn:0.9.0]"
```

If this constant is unused, the compiler is smart enough to not include it in the binary at all. So, we have to do something to get the compiler to include it. `fmt.Println(version)` for example does the trick. In this PR, I realized that assigning to `version` inside an `init` function works as well. 

Second, should I generate the version string inside the generated code or leave it in the `version` package? At first, it seems like an implementation detail, but after the meeting today, I realized it's actually crucially important. If we put the version in the generated code, then the version is really the version of the service weaver module in the `weaver generate` binary, not the version of the user's binary. Conversely, if we leave the version in the `version` package, then it is the version of the weaver module linked into the user's binary. I want the latter, so updated the PR. 

We also had some discussion about enforcing a consistent version between the code generator and the weaver module and between the generated code. I think we can address in a separate PR&mdash;it's not fundamentally tied to this version&mdash;so I'll leave it out of this PR.

spetrovic77 (approved):

PR #325, opened 2023-05-08
SSH and multiprocess deployers check API version.

mwhittaker:
This PR changes the SSH and multiprocess deployers to check the deployer API version embedded in a Service Weaver binary.

A question for reviewers. Weavelets still report their deployer API version via the pipe. Should we get rid of this?


spetrovic77 (approved):
Very nice!

spetrovic77:
I would also say we don't transmit the version in the protocol.

PR #326, opened 2023-05-09
Experimental compile time version checks.

mwhittaker:
See https://groups.google.com/g/golang-nuts/c/w1-JQMaH7c4 for similar discussion.


internal/tool/generate/generator.go
```diff
@@ -880,6 +881,14 @@ func (g *generator) generateImports(p printFn) {
 	p(`)`)
 }
 
+func (g *generator) generateVersionCheck(p printFn) {
+	p(`var _ %s = %s[[%d][%d][%d]int](%q)`,
+		g.codegen().qualify("LatestVersion"), g.codegen().qualify("Version"),
+		codegen.Major, codegen.Minor, codegen.Patch,
```
ghemawat, line 887: Should we avoid complaining about patch number changes? Those changes should be for bug fixes. If there is a compatibility change, we should expect at least a minor version bump?

internal/tool/generate/generator.go
```diff
@@ -880,6 +881,14 @@ func (g *generator) generateImports(p printFn) {
 	p(`)`)
 }
 
+func (g *generator) generateVersionCheck(p printFn) {
+	p(`var _ %s = %s[[%d][%d][%d]int](%q)`,
```
ghemawat, line 885: Add a comment showing sample output.

internal/tool/generate/generator.go
```diff
@@ -880,6 +881,14 @@ func (g *generator) generateImports(p printFn) {
...
+		g.codegen().qualify("LatestVersion"), g.codegen().qualify("Version"),
+		codegen.Major, codegen.Minor, codegen.Patch,
+		fmt.Sprintf(`You used 'weaver generate' codegen version %d.%d.%d, but you built your code with an incompatible weaver module version. Try upgrading and re-running 'weaver generate'.`, codegen.Major, codegen.Minor, codegen.Patch),
```
ghemawat, line 888: Clarify what should be upgraded in the "Try upgrading..." part.

runtime/codegen/version.go
```diff
@@ -0,0 +1,80 @@
...
+	// API at v0.12.0.
+	//
+	// TODO(mwhittaker): Write a doc explaining this version as well as the
+	// module version and deployer API version.
+	Major = 0
+	Minor = 10
+	Patch = 0
+)
+
+// The following types are used to check, at compile time, that every
```
ghemawat, line 61: Very nice!

ghemawat (changes requested):

PR #257, opened 2023-04-14
implemented the solution described by @rgrandl

naivary:
I have not tested it yet but wanted get some feedback if the code already has issues that can lead to problems in the future. The issue is described in #106.

PR #328, opened 2023-05-10
no duplicate locs allowed for the SSH deployer

rgrandl:
After PR #257, the SSH locations should contain distinct names/addresses.

This is because now we keep track of the deployment directory at each particular location.


mwhittaker (approved):
Thanks Robert! We chatted offline about allowing duplicate locations. We'll do that in a future PR.

PR #329, opened 2023-05-11
Make weavertest handle benchmarks.

ghemawat:
weavertest.Run expects either *testing.T or *testing.B which indicates that it should work for benchmarking components. However, the weavertest implementation was always passing a -test.run argument. We now set the -test.bench argument instead (and disable all tests) if we detect that the argument passed to weavetest.Run is a *testing.B.

### DIFF
--- a/weavertest/init.go
+++ b/weavertest/init.go
@@ -69,6 +69,7 @@ type testMainInterface interface{}
 //	func(ComponentType, ComponentType2)
 //	...
 func Run(t testing.TB, opts Options, testBody any) {
+	_, isBench := t.(*testing.B)
 	t.Helper()
 	runner, err := checkRunFunc(testBody)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


weavertest/init.go
```diff
@@ -69,6 +69,7 @@ type testMainInterface interface{}
 //	func(ComponentType, ComponentType2)
 //	...
 func Run(t testing.TB, opts Options, testBody any) {
+	_, isBench := t.(interface{ Elapsed() time.Duration })
```
spetrovic77, line 72: I'm just curious why you didn't do: `_, isBench := t.(*testing.B)`

spetrovic77 (approved):

weavertest/init.go
ghemawat: I was worried about the case where the caller had wrapped a *testing.B into their own type that implemented the same interface as *testing.B.

Perhaps it is better to just do the simple thing for now.

PR #330, opened 2023-05-12
Made weavertest respect the -test.v flag.

mwhittaker:
Currently, weavertest logs do not respect the `-test.v` flag. For example, a single process weavertest logs to stdout whether or not the `-test.v` flag is present. Similarly, benchmarks logs are always shown, which makes it very hard to read the benchmark results. This PR fixes these bugs. Thanks Sanjay for pointing them out. 

**For reviewers**:

- I recommend reading the PR commit-by-commit. 
- Should I delete the `weavertest/internal/logging` tests I added? I'm not sure how useful they are.

### DIFF
--- a/runtime/bootstrap.go
+++ b/runtime/bootstrap.go
@@ -43,6 +43,7 @@ type Bootstrap struct {
 	ToWeaveletFile *os.File // Pipe to send to weavelet (weavertest only).
 	ToEnvelopeFile *os.File // Pipe to send to envelope (weavertest only).
 	TestConfig     string   // Config file contents (weavertest only).
+	Quiet          bool     // Don't log or print anything.
 }
 
 // BootstrapKey is the Context key used by weavertest to pass Bootstrap to [weaver.Run].
--- a/runtime/logging/logger.go
+++ b/runtime/logging/logger.go
@@ -129,41 +129,59 @@ func StderrLogger(opts Options) *slog.Logger {
 	return slog.New(&LogHandler{Opts: opts, Write: writeText})
 }
 
-// testLogger implements a logger for tests.
-type testLogger struct {
+// TestLogger implements a logger for tests.
+type TestLogger struct {
 	t        testing.TB     // logs until t finishes
 	pp       *PrettyPrinter // pretty prints log entries
 	mu       sync.Mutex     // guards finished
 	finished bool           // has t finished?
 }
 
-// log logs the provided log entry using t.log.
-func (t *testLogger) log(entry *protos.LogEntry) {
+// Log logs the provided Log entry using t.Log.
+func (t *TestLogger) Log(entry *protos.LogEntry) {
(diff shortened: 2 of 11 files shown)


weavertest/internal/logging/logger_test.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2023 Google LLC
```
ghemawat, line 1: I think weavertest/internal/simple/simple_test.go should also show the issues? If so maybe we don't need this new directory?

runtime/bootstrap.go
```diff
@@ -43,6 +43,7 @@ type Bootstrap struct {
 	ToWeaveletFile *os.File // Pipe to send to weavelet (weavertest only).
 	ToEnvelopeFile *os.File // Pipe to send to envelope (weavertest only).
 	TestConfig     string   // Config file contents (weavertest only).
+	Quiet          bool     // Don't log or print anything.
```
ghemawat, line 46: Will this be something we want to provide to non-test weaver runs as well?

Also, I wonder if the default should be quiet with extra logging only if the user asks for it. A downside is that we might not print out useful URLs? But perhaps weaver status commands fix that problem?

runtime/logging/logger.go
```diff
@@ -129,41 +129,51 @@ func StderrLogger(opts Options) *slog.Logger {
 	return slog.New(&LogHandler{Opts: opts, Write: writeText})
 }
 
-// testLogger implements a logger for tests.
-type testLogger struct {
+// TestLogger implements a logger for tests.
```
ghemawat, line 132: Perhaps not for this change, but I wonder if we can avoid placing a dependency on the testing package from a non test file somehow?

One option would be to depend on a smaller interface than testing.TB (do we need more than the Log() method?). Another option (perhaps infeasible) would be to move this code to the weavertest package.

ghemawat (commented):
Thanks!

runtime/bootstrap.go
mwhittaker: I think it's only useful for single process weavertests. Right now, non-test weaver runs only show messages explicitly logged by the user. System logs aren't shown by default. They have to use `weaver logs --system` to see those. 

The one exception is the little rolodex that gets printed right when the program starts. 

```
╭───────────────────────────────────────────────────╮
│ app        : collatz                              │
│ deployment : 9d029072-ef9e-4c3d-b03f-e6e92afda1ad │
╰───────────────────────────────────────────────────╯
```

I added this because the user needs some way to learn the deployment id. Otherwise when they run things like `weaver status`, they aren't sure which deployment is which. We could provide the deployment id via some API and not print it out though.

ghemawat: Okay. Let's leave as is for now.

ghemawat (approved):

Issue #331, opened 2023-05-15
Error Running Latest Version of 'Weaver': Updating 'Slog' Package and 'Weaver Generate' Command Not Responding


ryan961:
1. When using the latest version of `weaver`, an error occurs (indicating that the version of the `slog` package needs to be updated).
```bash
➜  go run .        
# github.com/ServiceWeaver/weaver
vendor/github.com/ServiceWeaver/weaver/singleprocess.go:327:66: slog.HandlerOptions{…}.NewTextHandler undefined (type slog.HandlerOptions has no field or method NewTextHandler)
vendor/github.com/ServiceWeaver/weaver/weavelet.go:112:69: slog.HandlerOptions{…}.NewTextHandler undefined (type slog.HandlerOptions has no field or method NewTextHandler)
```

2.  `weaver generate` doesn't respond for a long time and there is no output on the command line. Below is a simple reproducing code. I'm not sure if my way of opening it is correct.

go version: 1.20.4
weaver: 0.10.0

ryan961:
I tried using `weaver generate . ./bff` as instructed in the documentation, but it seemed to hang with no response, and I had to forcefully terminate the command using `ctrl + c`. However, I found that using `go generate weaver generate ./...` generated the code directly without any issues.

ryan961:
Thank you for the reply. Regarding the first question, the issue should be resolved in the next version with the merging of the pull request (#333) . However, the second question is unrelated to this particular submission. Today, I followed the official tutorial (https://serviceweaver.dev/docs.html#step-by-step-tutorial-components) and tested it (`$ weaver generate .`) in a new environment, but I still encountered the same problem (the command hangs without any response) . I attempted to debug it and it seems to be stuck at [`pkgList, err := packages.Load(cfg, pkgs...)`](https://github.com/ServiceWeaver/weaver/blob/d2179635a1d0d4d5656cd5993a0c8963c7009d26/internal/tool/generate/generator.go#L109).

mwhittaker:
Thanks for investigating further @ryan961! Sorry you're running into this problem. First, can you share the output of `weaver multi version`? 

[...]

PR #333, opened 2023-05-15
Upgrade to `golang.org/x/exp v0.0.0-20230510235704-dd950f8aeaea`

ryan961:
- Update `go.mod` to use a newer version of `golang.org/x/exp`
- Refactor `singleprocess.go` to use a different `slog` handler initialization
- Refactor `weavelet.go` to use a different `slog` handler initialization
- Refactor `handler_test.go` to use a different `slog` handler initialization


mwhittaker (approved):
Awesome! Thanks @ryan961! Does this solve your issue in #331 or that's unrelated?

PR #337, opened 2023-05-16
Added component-related "weaver generate" checks.

mwhittaker:
Through our experience writing Service Weaver apps, we've realized that it's easy to make changes to an app and then to forget to re-run `weaver generate`. Sometimes, this causes the app to not build (:smile:). Sometimes, it causes the app to panic immediately when run (:neutral_face:). Sometimes, the app appears to run smoothly but has some subtle bug (:sob:). 

This PR is the beginning of an effort to generate code that doesn't build if you forget to re-run `weaver generate`. See commit messages for details. I recommend reading the PR one commit at a time.


spetrovic77 (approved):
Very nice!

PR #338, opened 2023-05-16
Added automarshal-related "weaver generate" checks. 

mwhittaker:
See #337 for context. This PR changes `weaver generate` to prevent changes to a struct with an embedded `AutoMarshal` without re-running `weaver generate`. See commit messages for details.


spetrovic77 (approved):
Yes to second commit, because we don't want to rely on the compiler to (possibly) remove these unused vars.

PR #339, opened 2023-05-16
Added WithConfig-related "weaver generate" checks.

mwhittaker:
Recall that a component implementation struct can embed `weaver.WithConfig[T]` to access a config of type `T` parsed from a weaver.toml file. Before this PR, `weaver generate` would generate a `ConfigFn` function for component structs with an embedded `WithConfig`. Here's an example:

```go
ConfigFn: func(i any) any { return i.(*sqlStore).WithConfig.Config() }
```

When a weavelet constructed a component, it would use this function (registered through the registry) to populate the config.

However, imagine you embed a `weaver.WithConfig[T]` inside a struct but then forget to run `weaver generate`. The code will build and run but because there is no generated `ConfigFn`, the config will not be parsed.

This PR removes the `ConfigFn` as it is not needed. Instead a weavelet can use reflection directly to decide whether a component struct has an embedded `weaver.WithConfig[T]`.


runtime/codegen/registry.go
```diff
@@ -137,6 +136,33 @@ func (r *registry) find(path string) (*Registration, bool) {
 	return reg, ok
 }
 
+// Config calls the WithConfig.Config method on the provided value and returns
+// the result. If the provided value doesn't have a WithConfig.Config method,
+// Config returns nil.
+//
+// Config panics if the provided value is not a pointer to a struct.
+func Config(v reflect.Value) any {
```
spetrovic77, line 144: I'm getting bothered about this being an exported function that isn't called from the generated code.

Would it be possible to place this in Registration.ConfigFn? codegen.Register() can auto-fill it when called.

Not a big deal if too painful.

spetrovic77 (approved):

spetrovic77 (approved):

runtime/codegen/registry.go
mwhittaker: I moved this function to an internal `config` package. I also added a TODO to move the `ComponentConfigValidator` function right below it as well. It's also not used by the generated code but moving it is a bit trickier.

PR #340, opened 2023-05-16
weavertest tests can now run in a single process mode with RPCs.

ghemawat:
The "SingleProcess bool" option has been replaced by a mode that can be either Local or Multi or RPC. The RPC mode places every component in the testing process, but uses RPCs for method calls. This mode is typically useful when collecting profiles or coverage data that should include RPC costs. (Multi runs will also contain RPC costs, but are harder to profile since their execution is split across multiple processes.)

The mode argument has also been moved out of weavertest.Options{} so it no longer has a default value and the caller is forced to specify whether they want a Local or Multi run.

### DIFF
--- a/weavertest/doc.go
+++ b/weavertest/doc.go
@@ -49,8 +49,7 @@
 //
 //  3. RPC: Every component will be placed in a same process, but
 //     method calls will use RPCs. This mode is most useful when
-//     profiling or collecting
-//     coverage information.
+//     profiling or collecting coverage information.
 //
 // Example:
 //
--- a/weavertest/doc.go
+++ b/weavertest/doc.go
@@ -32,15 +32,31 @@
 //	}
 //
 // weavertest.Run is passed a weavertest.Options, which you can use to configure
-// the execution of the test. By default, weavertest.Run will run every
-// component in a different process. This is similar to what happens when you
-// run weaver multi deploy. If you set the SingleProcess option, weavertest.Run
-// will instead run every component in a single process, similar to what
-// happens when you "go run" a Service Weaver application.
+// the execution of the test.
+//
+// The mode argument to weavertest.Run controls whether or not components are
+// placed in different processes and whether or not RPCs are used for components
+// in the same process.
+//
+//  1. Local: all components are placed in a single process and method
+//     invocations are local procedure calls. This is similar to what
+//     happens when you run the Service Weaver binary directly or run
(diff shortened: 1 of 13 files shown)


weavertest/doc.go
```diff
@@ -32,15 +32,31 @@
...
+//     the application using "go run".
+//
+//  2. Multi: Every component will be placed in a separate process,
+//     similar to what happens when you "weaver multi deploy" a
+//     Service Weaver application.
+//
+//  3. RPC: Every component will be placed in a same process, but
+//     method calls will use RPCs. This mode is most useful when
+//     profiling or collecting
+//     coverage information.
```
rgrandl, line 53: nit: this can go on the previous line

weavertest/init.go
```diff
@@ -28,12 +28,43 @@ import (
...
+	// for method invocations.
+	// This is the default value.
+	Local Mode = iota
+
+	// Multi places all components in different process (unless explicitly colocated)
+	// and uses RPCs for method invocations on remote components and local procedure calls
+	// for method invocations on colocated components.
+	Multi
+
+	// RPC places all components in the same process and uses RPCs for method invocations.
```
rgrandl, line 45: nit: maybe we can add RPC after Local, so it's single process local calls, single process remote calls, then multi process

website/docs.md
```diff
@@ -1265,21 +1265,28 @@ func TestArithmetic(t *testing.T) {
...
+configure the execution of the test. The mode argument controls the placement of
+components and how they communicate with each other:
+
+1. **weavertest.Local**: Every component will be placed in the test process, and
+   all component method calls will use local procedure calls, happens when you
+   `go run` a Service Weaver application.
+2. **weavertest.Multi**: Every component will be placed in a
+   different process. This is similar to what happens when you run `weaver multi
+   deploy`.
+3. **weavertest.RPC**: Every component will be placed in the test process, but
```
rgrandl, line 1277: nit: same comment as before; if you agree to change the order (RPC before multi), maybe we can update the order in the doc too.

rgrandl (approved):
Looks great, Sanjay!

weavertest/doc.go
ghemawat: Done.

ghemawat (commented):
Thanks. I will submit after tests pass.

PR #343, opened 2023-05-17
Added WithRouter-related "weaver generate" checks.

mwhittaker:
See commit messages for details.


component.go
```diff
@@ -151,6 +151,8 @@ func (i *Implements[T]) setInstance(c *componentImpl) { i.componentImpl = c }
 //nolint:unused
 func (i *Implements[T]) implements(T) {}
 
+func (*componentImpl) routedBy(if_youre_seeing_this_you_probably_forgot_to_run_weaver_generate) {}
```
spetrovic77, line 154: Nit: add `var _ Unrouted = &componentImpl{}`?

component.go
```diff
@@ -292,6 +294,22 @@ type ListenerOptions struct {
 // correctness. Only use routing to increase performance in the common case.
 type WithRouter[T any] struct{}
 
+func (WithRouter[T]) routedBy(T) {}
+
+// RoutedBy[T] is the interface implemented by a struct that embeds
+// weaver.RoutedBy[T].
+type RoutedBy[T any] interface {
+	routedBy(T)
```
spetrovic77, line 302: This is really cool! :trophy: 

I thought Go didn't allow you to define multiple methods with the same name but different args/results. Looks like it does thanks to embeddings. And then they do this hide-y thing that most modern languages do. Very interesting!

spetrovic77 (approved):
Awesome!

component.go
mwhittaker: Done.

component.go
mwhittaker: Thanks! Yeah, I also thought it was interesting how we can leverage Go's embedding semantics. Definitely was a brain teaser trying to figure this all out.

mwhittaker (commented):
Thanks Srdjan!

PR #344, opened 2023-05-17
Fold weavertest Mode and Options into a single Runner object.

ghemawat:
Previously, the user would type something like:

```
weavertest.Run(t, weavertest.Local, weavertest.Options{}, func(...) {
  ...
})
```

They can now just get a suitable Runner object and call Run() on it:

```
weavertest.Local.Run(t, func(...) {
  ...
})
```

Options are now set via "With<OptionName>" methods on a Runner. These methods return a new Runner and therefore multiple option setting calls can be chained.

### DIFF
--- a/weavertest/deployer.go
+++ b/weavertest/deployer.go
@@ -52,7 +52,7 @@ const DefaultReplication = 2
 type deployer struct {
 	ctx        context.Context
 	ctxCancel  context.CancelFunc
-	mode       Mode
+	runner     Runner
 	wlet       *protos.EnvelopeInfo   // info for subprocesses
 	config     *protos.AppConfig      // application config
 	colocation map[string]string      // maps component to group
@@ -101,7 +101,7 @@ type connection struct {
 var _ envelope.EnvelopeHandler = &handler{}
 
 // newDeployer returns a new weavertest multiprocess deployer.
-func newDeployer(ctx context.Context, wlet *protos.EnvelopeInfo, config *protos.AppConfig, mode Mode, logWriter func(*protos.LogEntry)) *deployer {
+func newDeployer(ctx context.Context, wlet *protos.EnvelopeInfo, config *protos.AppConfig, runner Runner, logWriter func(*protos.LogEntry)) *deployer {
 	colocation := map[string]string{}
 	for _, group := range config.Colocate {
 		for _, c := range group.Components {
@@ -112,7 +112,7 @@ func newDeployer(ctx context.Context, wlet *protos.EnvelopeInfo, config *protos.
 	d := &deployer{
 		ctx:        ctx,
 		ctxCancel:  cancel,
-		mode:       mode,
+		runner:     runner,
 		wlet:       wlet,
 		config:     config,
 		colocation: colocation,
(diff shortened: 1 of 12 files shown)


weavertest/deployer.go
```diff
@@ -52,7 +52,7 @@ const DefaultReplication = 2
 type deployer struct {
 	ctx        context.Context
 	ctxCancel  context.CancelFunc
-	mode       Mode
+	runner     Runner
```
rgrandl, line 55: nit: add a comment

weavertest/init.go
```diff
@@ -28,48 +28,57 @@ import (
...
+// WithConfig returns a new Runner with the specified Service Weaver
+// configuration. The config value passed here is identical to what
+// might be found in a Service Weaver config file. It can contain
+// application level as well as component level configuration.
+func (r Runner) WithConfig(config string) Runner { r.config = config; return r }
+
+// Name returns the runner name. It is suitable for use as a sub-test or sub-benchmark name.
+func (r Runner) Name() string {
+	if r.name == "" {
+		return "Default"
```
rgrandl, line 72: nit: would "Undefined", "Unknown"  be a better name? "Default" sounds a little bit misleading (maybe).

weavertest/init.go
```diff
@@ -28,48 +28,57 @@ import (
...
 	}
+	return r.name
 }
 
-// AllModes returns a slice of all supported weavertest modes.
-func AllModes() []Mode { return []Mode{Local, RPC, Multi} }
+// SingleProcess returns true iff the runner will run all components in a single process.
+func (r Runner) SingleProcess() bool { return !r.multi }
```
rgrandl, line 78: I am wondering if it would be clearer for a developer, if we have explicit methods for all the 3 modes instead:

SingleProcessLocalCalls{ ... }
SingleProcessRemoteCalls{ ...}
MultiProcess { ... }

or something like that.

weavertest/init.go
```diff
@@ -131,11 +137,11 @@ func Run(t testing.TB, mode Mode, opts Options, testBody any) {
 		}
 	}()
 
-	if mode == Local {
-		ctx = initSingleProcess(ctx, opts.Config)
+	if !r.multi && !r.forceRPC {
+		ctx = initSingleProcess(ctx, r.config)
```
rgrandl, line 141: nit: this is initSingleProcessLocal i think

rgrandl (approved):
Very nice! Tests look so much better.

weavertest/init.go
ghemawat: Done.

weavertest/deployer.go
ghemawat: Done.

weavertest/init.go
ghemawat: At least the callers we have so far are interested in properties that don't exactly match the mode. E.g., SingleProcess() is true for Local+RPC, UsesRPC is true for RPC+Multi.

An alternative would be to just let the caller use "runner == weavertest.Local || runner == weavertest.Multi" etc.?

rgrandl: I think it would be cleaner to let the user use the Local/Multi/RPC modes directly. 

I find it confusing that we have to think about single vs multi, but also RPC vs local. 

I don't have strong opinions though, the current implementation is great too.

ghemawat: I decided to drop the accesors now and just let the caller compare runner directly. This should suffice since:

1. The need will mostly show up in the weaver implementation since we only use this to test that behavior changes as expected as the runtime changes.
2. The caller can keep info parallel to AllRunners() results if they want (perhaps by avoiding AllRunners() and making their own testCase slice).

rgrandl: Sounds good to me.

PR #347, opened 2023-05-18
Request fresh certificates on every connection establishment.

spetrovic77:
This is required as certificates in general have a short validity and typically get rotated. Previously, we would pass the certificate to the weavelet and use it indefinitely. We now request a new (i.e., latest) certificate from the weavelet each time a connection is established, ensuring that the freshest certificate is always used.


runtime/protos/runtime.proto
```diff
@@ -487,6 +486,19 @@ message ExportListenerReply {
...
+// and the corresponding private key.
+message GetSelfCertificateRequest {}
+
+// GetSelfCertificateReply is a reply to a GetSelfCertificateRequest.
+message GetSelfCertificateReply {
+  // PEM-encoded blocks containing (1) the certificate identifying the
+  // weavelet and (2) the corresponding private key. Used for enforcing mTLS
+  // security across components.
+  bytes cert = 1; // certificate identifying the weavelet
+  bytes key = 2;        // private key for the leaf cert in self_cert_chain
```
mwhittaker, line 499: Replace `self_cert_chain` with `cert`?

mwhittaker (approved):
Looks great! :fire:

Issue #350, opened 2023-05-19
weaver multi deploy problem


haroldyong:
envirement:
go version: 1.20.4 windows/amd64
weaver: v0.10.0

issue:
1. i follow the guide (https://serviceweaver.dev/docs.html#step-by-step-tutorial) locally .  everything is fine, until "weaver multi deploy weaver.toml" . there is is issue message "start main process: unknown component "github.com/ServiceWeaver/weaver/Main"
"
2. to find the answer,  i git clone the source code (https://github.com/ServiceWeaver/weaver) , go to example/hello(https://github.com/ServiceWeaver/weaver/tree/main/examples/hello) . then run "weaver multi deploy weaver.toml " . it run correctly . 

help:
who can tell me which wrong ? thanks.

my source tree:
![image](https://github.com/ServiceWeaver/weaver/assets/1647038/69198087-6242-438a-a3d7-940a9db607e7)

go.mod content:
```
module hello

go 1.20

require (
	github.com/ServiceWeaver/weaver v0.10.0
	go.opentelemetry.io/otel v1.15.1
	go.opentelemetry.io/otel/trace v1.15.1
)

require (
	github.com/BurntSushi/toml v1.2.0 // indirect
	github.com/DataDog/hyperloglog v0.0.0-20220214164406-974598347557 // indirect
	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20220418222510-f25a4f6275ed // indirect
	github.com/dustin/go-humanize v1.0.0 // indirect
	github.com/fsnotify/fsnotify v1.5.4 // indirect
	github.com/go-logr/logr v1.2.4 // indirect
	github.com/go-logr/stdr v1.2.2 // indirect
	github.com/google/cel-go v0.12.5 // indirect
	github.com/google/pprof v0.0.0-20221118152302-e6195bd50e26 // indirect
	github.com/google/uuid v1.3.0 // indirect
	github.com/hashicorp/golang-lru/v2 v2.0.1 // indirect
	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51 // indirect
	github.com/lightstep/varopt v1.3.0 // indirect
	github.com/mattn/go-isatty v0.0.16 // indirect
	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8 // indirect
	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
[...]

PR #351, opened 2023-05-19
Change component interface name separator character to slash on Windows.

lemon-1997:
`\` character is the path separator on Windows. This is different from the `/` separator on Linux. The component `Name` and `RefData` generated by `weaver generate` on Windows will have `\`, not `/`. Such as `github.com\ServiceWeaver\weaver\Main`. It will cause two problems:
1. differenet from runtime.Main
 ```go
// Main is the name of the main component.
const Main = "github.com/ServiceWeaver/weaver/Main"
 ```
2. `ExtractEdges` cannot match refData containing `\` characters, because the regexp only matches `/`. 

If deploy by weaver multi, it will fali. See issue:https://github.com/ServiceWeaver/weaver/issues/350

### DIFF
--- a/internal/tool/generate/generator.go
+++ b/internal/tool/generate/generator.go
@@ -26,6 +26,7 @@ import (
 	"go/types"
 	"io"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
@@ -548,7 +549,7 @@ type component struct {
 }
 
 func fullName(t *types.Named) string {
-	return filepath.Join(t.Obj().Pkg().Path(), t.Obj().Name())
+	return path.Join(t.Obj().Pkg().Path(), t.Obj().Name())
 }
 
 // intfName returns the component interface name.


lemon-1997:
I haven't followed weaver for a while, weaver is really great now. Compared with before, using weaver is simpler, the code is cleaner, and the generated assertion checks can reduce bugs.

mwhittaker:
Hi @lemon-1997! Just a clarifying question, does `t.Obj().Pkg().Path()` return `github.com\ServiceWeaver\weaver\Main`?

lemon-1997:
@mwhittaker Yes, `github.com\ServiceWeaver\weaver\Main` is generated on Windows

mwhittaker:
> @mwhittaker No, the return of `t.Obj().Pkg().Path()` is correct, the reson is `filepath.Join` changes the separator to `\` on Windows.

Ah, that makes sense! I was confused because `t.Obj().Pkg().Path()` returning backslashes didn't make sense to me, but I also didn't realize that `filepath.Join` replaces all forward slashes with backslashes. Thanks for looking into this and explaining!

internal/tool/generate/generator.go
```diff
@@ -548,7 +548,7 @@ type component struct {
 }
 
 func fullName(t *types.Named) string {
-	return filepath.Join(t.Obj().Pkg().Path(), t.Obj().Name())
```
mwhittaker, line 551: Does `path.Join(t.Obj().Pkg().Path(), t.Obj().Name())` work? I replaced the `filepath` package with the `path` package.

lemon-1997: Yes, it works, `path` package should be used instead of the `filepath` path package.

mwhittaker (approved):
Thanks @lemon-1997!

PR #354, opened 2023-05-24
Fold sub-test creation into weavertest.

ghemawat:
Previously, we required users to use a double-lambda: one to get a sub-test per Runner, and another to pass as the weaver test body to weavertest.Run. We now make weavertest responsible for creating the sub-test. This simplifies application code from two nested lambdas to a single lambda.

Split weavertest.Run() into weavertest.Test() and weavertest.Bench(), which gives us better type-checking.

Dropped TestFailer from diverge_test.go since we no longer have a way to catch component initialization erorrs (they cause a testing.T.Fatal error).


rgrandl (approved):
Looks amazing Sanjay. Happy to see one layer of lambdas gone.

PR #357, opened 2023-05-25
weavertest now supports fake component implementations.

ghemawat:
Some of the components in a Service Weaver application can now be replaced with fake test-specific implementations when using weavertest. The test code registers a set of such fake implementations with weavertest.Runner. When the test runs, the real component implementations are replaced by the supplied fakes.

### DIFF
--- a/weavertest/init.go
+++ b/weavertest/init.go
@@ -44,6 +44,12 @@ type Runner struct {
 	// config file. It can contain application level as well as
 	// component level configuration.
 	Config string
+
+	// Fakes holds a list of component implementations that should
+	// be used instead of the implementations registered in the binary.
+	// The typical use is to override some subset of the application
+	// code being tested with test-specific component implementations.
+	Fakes []FakeComponent
 }
 
 var (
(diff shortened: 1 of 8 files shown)


weavertest/init.go
```diff
@@ -65,6 +71,18 @@ var (
...
 
+// FakeComponent records the implementation to use for a specific component type.
+type FakeComponent struct {
+	intf reflect.Type
+	impl any
+}
+
+// Fake arranges to use impl as the implementation for the component type T.
+// The result is typically placed in Runner.Fakes.
+func Fake[T any](impl T) FakeComponent {
```
mwhittaker, line 82: Because `Fake` infers `T` from the type of `impl`, if you forget to explicitly pass `T`, I think `Fake` can sometimes infer the wrong type. Taking `simple_test.go` as an example, if the user were to write the following:

```go
fake := &fakeDest{}
runner.Fakes = append(runner.Fakes, weavertest.Fake(fake))
```

`Fake` infers `T` to be `*fakeDest`. What do you think about changing `Fake` to the following to force the user to explicitly pass `T`?

```go
func Fake[T any](impl any) FakeComponent {
    if _, ok := impl.(T); !ok {
        panic(...)
    }
    return FakeComponent{intf: reflect.TypeOf((*T)(nil)).Elem(), impl: impl}
}
```

A user can still easily mess things up though, and it complicates the signature, so it might not be worth it. Up to you.

mwhittaker (approved):
Brilliant! Very nice how you changed the weavertest deployer to run fakes locally to make things work.

weavertest/init.go
ghemawat: I like requiring T, so I will switch to your suggestion and ping back when that is done.

PR #358, opened 2023-05-25
Remove local_address from GetListenerAddressRequest

rgrandl:
When the weavelet sends a `GetListenerRequest` to the deployer, it also propagates the local address. However, the local address is not used by any deployer on this path. I found passing local address on `GetListenerRequest` very confusing, but then I realized that this field is never used.


mwhittaker (approved):
Great catch Robert! API is much nicer now. Can you also update `runtime/version/version.go` to 0.13.0?

PR #359, opened 2023-05-26
Move app initialization code to a weaver.Main method.

ghemawat:
Previously, application startup code was wrapped up in a function passed to weaver.Run. However this made it hard to test startup code. We now have a `Main` method in `weaver.Main` and place the startup code in the implementation of this method.

Refactored startup the internal API used by weavertest to separate out starting the application from running its startup code.


component.go
```diff
@@ -116,8 +117,15 @@ type component struct {
 var _ Instance = &componentImpl{}
 
 // Main is interface implemented by an application's main component.
-// This component is created automatically by `weaver.Run`.
-type Main interface{}
+// This component is instantiated and its Main() method called by
+// `weaver.Run`.
+type Main interface {
+	// TODO(reviewer): Should we make this method optional?
```
mwhittaker, line 123: If someone opts to not implement the `Main` method, what would be the behavior of `weaver.Run`? I guess we would still run the main component's `Init` method and then block forever? 

I like making `weaver.Ref[weaver.Main]` not compile somehow.

mwhittaker (approved):
Awesome change Sanjay! I'm loving the new API :)

component.go
spetrovic77, line 123: Making it optional would involve maintaining Main as an empty interface and using reflection to discover if the Main impl implements this method?

That seems less clear to me for the user. I like your alternatives below, however.

examples/factors/main.go
```diff
@@ -31,7 +31,7 @@ var localAddr = flag.String("local_addr", "localhost:9000", "Local address")
 func main() {
 	flag.Parse()
 	ctx := context.Background()
-	if err := weaver.Run(ctx, serve); err != nil {
+	if err := weaver.Run(ctx); err != nil {
```
spetrovic77, line 34: FYI, I really like that we don't pass a lambda here any more. It makes the code much simpler, as the user *has to* know there's a side main implementation that has to get called.

spetrovic77 (approved):

component.go
ghemawat: Yes, my intention was that it would block.

I  will try to make weaver.Ref[weaver.Main] not compile. And perhaps make the code generator refuse to make RPCs for Main?

ghemawat: Yes. The code is pretty simple, but the question is whether or not it is more complicated for the user. So I will try to just disable Ref[] and RPCs.

weavertest/multi.go
```diff
@@ -40,13 +42,14 @@ const matchNothingRE = "a^" // Regular expression that never matches
 // component level configs. config is allowed to be empty.
 //
 // Future extension: allow options so the user can control collocation/replication/etc.
-func initMultiProcess(ctx context.Context, name string, isBench bool, runner Runner, logWriter func(*protos.LogEntry)) (context.Context, func() error, error) {
+func initMultiProcess(ctx context.Context, t testing.TB, isBench bool, runner Runner, logWriter func(*protos.LogEntry)) (context.Context, func() error, error) {
```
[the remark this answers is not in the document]
ghemawat: That's a good point. I think that I am leaning towards sticking with letting weaver/weavertest call Main.Main (as opposed to the user explicitly calling it) to keep things similar between non-test and test code. Otherwise, the user has to remember that in scenario Main.Main is called automatically and in another they are responsible.

Note also that Main.Main will only run in weavelets that are hosting weaver.Main. So this could just be considered part of the issue of how test code controls replication.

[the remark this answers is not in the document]
mwhittaker: Sounds good to me!

mwhittaker (approved):
Looks great!

spetrovic77 (approved):

weavertest/multi.go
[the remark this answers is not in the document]
mwhittaker: I was writing some unit tests and ran into two things that are a bit annoying with the new API (assuming there is more than one test). They are both related to `Main` being run more than once.

1. You can't call `http.HandleFunc` (or any other function that registers an endpoint with the default mux). When the second test runs, it calls `Main` again, and `http.HandleFunc` panics. 
2. If your server listens on a fixed port, like `localhost:9000`, the second test will fail with a `bind: address already in use` if your `Main` doesn't respect the context it receives.

I'm not sure how to fix these issues, but wanted to document them somewhere.

PR #361, opened 2023-05-30
Added "weaver single deploy <config>" command.

mwhittaker:
This PR introduces a `weaver single deploy <config>` command that acts like `go run .` The new command has two benefits. First, it makes the weaver tool's UI more uniform, as there's already a `weaver multi deploy`, `weaver ssh deploy`, etc. Second, it makes it easier to provide config files to single process deployments. Previously, you had to use the `SERVICEWEAVER_CONFIG` environment variable, which was unergonomic and easy to forget.


rgrandl (approved):
Nice, thanks Michael!

PR #362, opened 2023-05-30
Fixed bug in logic deciding when to run main.

mwhittaker:
Previously, a weavelet decided to run the main component's Main method if it had a registered main component that was marked local. This worked for the single process deployers but broke all other deployers. For weavelets not started in singleprocess mode, the weavelet receives an explicit `RunMain` field in `EnvelopeOptions` that signals it to start running main.

This bug also re-emphasizes the importance of tests that automatically run all the deployers. I'll prioritize this and finish up PR #287.

### DIFF
--- a/weavelet.go
+++ b/weavelet.go
@@ -251,7 +251,9 @@ func (w *weavelet) start() error {
 }
 
 func (w *weavelet) Wait(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


ghemawat (approved):
Thanks. Sorry about missing this case.

PR #364, opened 2023-05-30
Added test to catch incorrect version updates.

mwhittaker:
Recall that we hardcode the deployer API version into every weaver application binary. The hardcoded string is called version. Additionally, we have constants for the Major, Minor, and Patch of the deployer API version. These two need to stay in sync, but because version has to be hardcoded, there is no way to programmatically keep them in sync. This makes it easy to change Major, Minor, and Patch but forget to update version. This PR adds a unit test that fails if you forget to change version.

### DIFF
--- a/runtime/version/version_test.go
+++ b/runtime/version/version_test.go
@@ -22,6 +22,23 @@ import (
 	"testing"
 )
 
+// Test that version matches Major, Minor, and Patch.
+func TestStaleVersion(t *testing.T) {
+	major, minor, patch, err := extractVersion([]byte(version))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if major != Major {
+		t.Errorf("major: got %d, want %d", major, Major)
+	}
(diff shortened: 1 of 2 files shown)


runtime/version/version_test.go
```diff
@@ -22,6 +22,23 @@ import (
 	"testing"
 )
 
+// Test that version matches Major, Minor, and Patch.
+func TestStaleVersion(t *testing.T) {
```
rgrandl, line 26: I am not sure we need this test. `TestReadVersion` is a superset of this test, and it fails as well if the hardcoded version string is not updated. 

I don't have strong feelings though.

rgrandl (approved):

runtime/version/version_test.go
mwhittaker: Oh oops! I didn't realize the other tests would fail. I'll abandon the PR.

PR #367, opened 2023-05-31
Add test for examples/chat server.

ghemawat:
Added a way for tests to get the host:port for a listener.

Reduce weavertest time spent waiting for weaver.Main.Main to exit to make tests run faster.

Run weaver.Main Init before starting test code.

Added examples/chat/server_test.go.

Allow chat server to accept GET requests for new threads and posts (for testing purposes).

### DIFF
--- a/examples/chat/server_test.go
+++ b/examples/chat/server_test.go
@@ -1,0 +1,126 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"regexp"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/ServiceWeaver/weaver"
+	"github.com/ServiceWeaver/weaver/weavertest"
+)
+
+// db is a fake store for the chat database.
+type db struct {
+	mu       sync.Mutex
+	lastPost PostID
+	threads  []Thread
+}
+
+var _ SQLStore = &db{}
+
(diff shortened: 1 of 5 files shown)


examples/chat/server_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (db *db) GetFeed(ctx context.Context, user string) ([]Thread, error) {
+	db.mu.Lock()
+	defer db.mu.Unlock()
+	return db.threads, nil
+}
+
+func (db *db) GetImage(ctx context.Context, _ string, image ImageID) ([]byte, error) {
+	db.mu.Lock()
```
mwhittaker, line 69: Remove locking?

examples/chat/server_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t := db.threads[int(thread-1)]
+	t.Posts = append(t.Posts, Post{ID: db.lastPost, Creator: creator, When: when, Text: text})
+	db.threads[int(thread-1)] = t
+	return nil
+}
+
+func (db *db) GetFeed(ctx context.Context, user string) ([]Thread, error) {
+	db.mu.Lock()
+	defer db.mu.Unlock()
+	return db.threads, nil
```
mwhittaker, line 65: Do we have to return a deep copy of `db.threads` to avoid races? I guess it doesn't really matter for these tests?

weavelet.go
```diff
@@ -367,6 +390,11 @@ func (w *weavelet) getListener(name string, opts ListenerOptions) (*Listener, er
 	if reply.Error != "" {
 		return nil, fmt.Errorf("getListener(%q): %s", name, reply.Error)
 	}
+
+	ls := w.getListenerState(name)
+	ls.addr = l.Addr().String()
```
mwhittaker, line 395: Is it okay that the assignment here isn't guarded by a lock? Can concurrent calls to `getListener` with the same name race?

mwhittaker (approved):
Nice!

examples/chat/server_test.go
ghemawat: Done.

examples/chat/server_test.go
ghemawat: Done.

weavelet.go
ghemawat: Great catch. Fixed.

PS. I am not sure our listener management code works well in the face of multiple calls to get the listener. E.g., we might listen multiple times on the same address.

ghemawat (commented):
Thanks!

weavelet.go
mwhittaker: Great point. Hopefully when we switch to the new listener API, we can also clean up some of these things.

PR #370, opened 2023-06-01
Add reflection.Type[T]() helper and use it.

ghemawat:
Previously we would have to write complicated code to get the reflect.Type for an interface since an interface cannot have a concrete value and therefore reflect.TypeOf() did not work. Encapsulated this logic into a single generic function.

### DIFF
--- a/internal/reflection/reflection_test.go
+++ b/internal/reflection/reflection_test.go
@@ -1,0 +1,26 @@
+package reflection_test
+
+import (
+	"fmt"
+	"io"
+	"reflect"
+	"testing"
+
+	"github.com/ServiceWeaver/weaver/internal/reflection"
+)
+
+func TestType(t *testing.T) {
+	for _, test := range []struct {
+		want string
+		t    reflect.Type
+	}{
+		{"int", reflection.Type[int]()},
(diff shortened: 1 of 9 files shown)


internal/reflection/reflection_test.go
```diff
@@ -0,0 +1,40 @@
...
+	for _, test := range []struct {
+		want string
+		t    reflect.Type
+	}{
+		{"int", reflection.Type[int]()},
+		{"io.Reader", reflection.Type[io.Reader]()},
+	} {
+		t.Run(test.want, func(t *testing.T) {
+			if got := fmt.Sprint(test.t); got != test.want {
+				t.Errorf("reflection.Type[io.Reader] = %s, expecting %s", got, test.want)
```
mwhittaker, line 36: Maybe replace with something like `t.Errorf("reflection.Type[%s] = %s, expecting %s", test.want, got, test.want)` to avoid the hardcoding of `io.Reader`?

mwhittaker (approved):
Very nice!!!

internal/reflection/reflection_test.go
ghemawat: I removed the type arg from the error message since the sub-test name has the useful info.

PR #372, opened 2023-06-01
Fixed args + env bug in "weaver single deploy".

mwhittaker:
This PR fixes two bugs. First, `weaver single deploy <config>` was not actually passing the config file to the underlying Service Weaver app, completely defeating the purpose of the command (d'oh). Second, I was incorrectly passing arguments using `Cmd.Args` instead of `exec.Command`.


rgrandl (approved):
Ha, I did such a poor review on this (last time). Thanks for fixing it.

mwhittaker:
It was the code that was poor, not the review! 🤦

Issue #369, opened 2023-06-01
runner.Test undefined (type weavertest.Runner has no field or method Test


Markbnak:
I'm running weaver v0.12.

I followed the [official documentation](https://serviceweaver.dev/docs.html#testing) to create tests.

I use a similar code:

```
func TestAdd(t *testing.T) {
     runner := weavertest.Local  // A runner that runs components in a single process
     runner.Test(t, func(t *testing.T, adder Adder) {
         ctx := context.Background()
         got, err := adder.Add(ctx, 1, 2)
         if err != nil {
             t.Fatal(err)
         }
         if want := 3; got != want {
             t.Fatalf("got %q, want %q", got, want)
         }
     })
}
```

It seems like weavertest.Local.Test does not exist.

`runner.Test undefined (type weavertest.Runner has no field or method Test)`

Youyou23:
Running onto the same issue. 
I've tried also with version v0.10 => same compilation error ...

mwhittaker:
Ah, this is unfortunately a known problem with our online documentation. The docs reflect the main branch of our repo which is ahead of the latest version (0.12.0 in this case). We'll release v0.13.0 soon which should fix the problem. We'll also work on changing the online docs to reflect the latest version. Sorry for this!

mwhittaker:
This should be fixed now! I'll close the issue, but please re-open if the problem persists.

Markbnak:
Thank you for this, it works now with version v0.13.0

PR #373, opened 2023-06-01
Only push website changes when a tag is created.

mwhittaker:
Before this PR, every commit merged into main would trigger an update to our public website. This meant that the docs would reference code that wasn't yet merged into a stable release. This led to a lot of confusion as people tried to follow along with the docs, but nothing worked.

This PR changes the GitHub Action to only update the website when we make a new tag for a release. You can also manually trigger the action if you want. In the future, we should consider versioning our documentation and allowing people to view old versions?

Related to #369.


rgrandl (approved):
Nice!

PR #376, opened 2023-06-06
Add a test for the examples

mwhittaker:
This PR was originally written by @dnephin in https://github.com/ServiceWeaver/weaver/pull/287. The tests were
working locally but not on GitHub Actions. After some debugging, I found
that running subcommands using "sh -c" was causing things to hang. I
still have no idea why that is, but things are working now :)

I also tweaked logger_test.go because it was printing a lot (and I mean
a LOT) of stuff. When debugging on GitHub Actions, these logs were very
annoying.

Original commit messages from @dnephin:

> This commit adds a test case for running each of the examples as both a
> single deploy and multi-deploy. This test should ensure that the
> examples are always in a working state, and provide some end-to-end
> coverage of the single and multi deployers.
>
> It also provides some test coverage of the integration of codegen with a
> real application.
>
> Register signal handler before listening
>
> TestExamples only waits for a listener to be listening and handling HTTP
> requests.
>
> The signal handler must be registered before the listener is started,
> otherwise a test waiting on the listener could send a SIGTERM before
> the signal handler is registered.

### DIFF
--- a/examples/examples_test.go
+++ b/examples/examples_test.go
@@ -1,0 +1,181 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package examples
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"syscall"
+	"testing"
+	"time"
+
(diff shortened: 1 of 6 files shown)


examples/examples_test.go
```diff
@@ -0,0 +1,181 @@
...
+			})
+
+			// "weaver multi deploy" the application.
+			t.Run("weaver-multi", func(t *testing.T) {
+				cmd := startCmd(ctx, t, "../../cmd/weaver/weaver", "multi", "deploy", "weaver.toml")
+				t.Cleanup(terminateCmdAndWait(t, cmd))
+				run(t, test)
+			})
+
+			// TODO: other deployers?
```
spetrovic77, line 100: Can we do other deployers? Actually, not sure if there's a point, since we're only testing the example application here, not the deployers. Multi already tests everything (replication, shared memory etc.)

spetrovic77 (approved):
Looks great!

examples/examples_test.go
mwhittaker: I think these tests are useful for testing the deployers themselves actually! We could `weaver ssh` and `weaver gke-local`, but those are trickier, so I'll leave them for later :)

dnephin:
Thank you for debugging this and getting it merged!

PR #377, opened 2023-06-07
Check database/sql Rows.err().

lemon-1997:
`Rows.next()` returns false if there is no next result row or an error happened while preparing it. So we need to check it.


mwhittaker (approved):
Great change! Thanks for cleaning this up!

PR #385, opened 2023-06-09
Added `weaver completion` bash completion command.

mwhittaker:
This PR introduces a `weaver completion` command that spits out a bash completion script that enables tab completion for the weaver command. For example:

```bash
$ weaver <tab><tab>
generate   gke        gke-local  multi      single     ssh        version
$ weaver single <tab><tab>
dashboard  deploy     help       metrics    profile    purge      status     version
$ weaver single p<tab><tab>
profile  purge
```

To enable the tab completion, you run `source <(weaver completion)`. This mirrors [`kubectl completion bash`][3]. Source the file in your `~/.bashrc` if you always want it to be enabled.

I added `weaver completion` selfishly, as I thought it was useful. If others find it useful, we can advertise it to users. There is a bit of overhead in updating it whenever commands change, and the gke commands could be stale, so it would require work to be more robust. Also happy to leave this out of the repo and have it privately for myself.

[3]: https://kubernetes.io/docs/reference/kubectl/cheatsheet/#kubectl-autocomplete

### DIFF
--- a/cmd/weaver/weaver_completion.bash
+++ b/cmd/weaver/weaver_completion.bash
@@ -1,0 +1,63 @@
+#/usr/bin/env bash
+
+# This file is a bash completion script that enables tab completion for the
+# weaver command. For example:
+#
+#     $ weaver <tab><tab>
+#     generate   gke        gke-local  multi      single     ssh        version
+#     $ weaver single <tab><tab>
+#     dashboard  deploy     help       metrics    profile    purge      status     version
+#     $ weaver single p<tab><tab>
+#     profile  purge
+#
+# To enable the tab completion, you must source this file, typically by running
+# `source <(weaver completion)`. This mirrors `kubectl completion bash` [3].
+# Source the file in your ~/.bashrc if you always want it to be enabled.
+#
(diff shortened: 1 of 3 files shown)


cmd/weaver/weaver_completion.bash
```diff
@@ -0,0 +1,63 @@
...
+  if [[ "${COMP_WORDS[1]}" == "version" ]]; then
+    return
+  fi
+
+  # weaver {single,multi,ssh,gke,gke-local}
+  declare -A subcommands
+  subcommands["single"]="dashboard deploy help metrics profile purge status version"
+  subcommands["multi"]="dashboard deploy help logs metrics profile purge status version"
+  subcommands["ssh"]="dashboard deploy help logs version"
+  subcommands["gke"]="dashboard deploy help kill logs profile status version"
```
spetrovic77, line 49: This is really nice, Michael!

Is there any way to defer to weaver-gke and weaver-gke-local executables for completion on those? This will also help with other deployer commands.

mwhittaker: Thanks! Oh great point. Let me do some research into that.

rgrandl:

PR #386, opened 2023-06-09
Remove versioning checks in multi/ssh deployers

rgrandl:
Note that we do these checks when we create an envelope, hence these checks are stale.


mwhittaker (approved):
Thanks Robert!

Issue #383, opened 2023-06-09
Weaver generate doesn't ignore the "weaver_gen.go" file while parsing pkgs


songsy13:
Hi, I define a golang struct embeded the `weaver.Automarshal`and run `weaver generate ./...` in my project dir, and it generates a weaver_gen.go file which includes the `WeaverMarshal` and `WeaverUnmarshal` methods of that struct. 
```
type Embed struct {
	weaver.AutoMarshal
	d1 int
	d2 int
	str   string
	slice []string
	kv map[string]int
}

func (x *Embed) WeaverMarshal(enc *codegen.Encoder) {
	if x == nil {
		panic(fmt.Errorf("Embed.WeaverMarshal: nil receiver"))
	}
	enc.Int(x.d1)
	enc.Int(x.d2)
	enc.String(x.str)
	serviceweaver_enc_slice_string_4af10117(enc, x.slice)
	serviceweaver_enc_map_string_int_c20ee031(enc, x.kv)
}

func (x *Embed) WeaverUnmarshal(dec *codegen.Decoder) {
	if x == nil {
		panic(fmt.Errorf("Embed.WeaverUnmarshal: nil receiver"))
	}
	x.d1 = dec.Int()
	x.d2 = dec.Int()
	x.str = dec.String()
	x.slice = serviceweaver_dec_slice_string_4af10117(dec)
	x.kv = serviceweaver_dec_map_string_int_c20ee031(dec)
}
```

However when I delete a field defined in the struct and re-run `weaver generate ./...`, it returns the following errors:
```
type Embed struct {
	weaver.AutoMarshal
	d1    int
	d2    int
	//str   string  // delete this field 
	slice []string
	kv    map[string]int
}
```

```
model/weaver_gen.go:28:15: x.str undefined (type *Embed has no field or method str)
model/weaver_gen.go:39:4: x.str undefined (type *Embed has no field or method str)
```
I have to manually delete the `weaver_gen.go` file first and then run `weaver generate` if I change any field of a struct. I  think that is probably not very reasonable :(

When I read the `generate` source code, I find out that there is a `parseNonWeaverGenFile` function whose purpose is to ignore the `weaver_gen.go` file. So I'm confused if the aboving errors are expected ?

mwhittaker:
Great catch. That's a bug. I'll work on fixing it.

mwhittaker:
[...]

PR #387, opened 2023-06-12
Fixed broken build constraints in weaver_gen.go.

mwhittaker:
Imagine you're writing a Service Weaver program with the following code:

```go
type Pair struct {
    weaver.AutoMarshal
    x int
    y int
}
```

You run `weaver generate` and `go build`, and everything works fine. Then, you delete the `y` field and try to `go build`. This fails because the `weaver_gen.go` file is out of sync with the source code. The code in `weaver_gen.go`, for example, could be trying to serialize the `y` field, but the field doesn't exist anymore.

To fix this, You have to re-run `weaver generate` first. For this reason, it is critical that `weaver generate` ignore the contents of `weaver_gen.go` files. Otherwise, `weaver generate` would also error out, and you'd have to delete your `weaver_gen.go` files every time you made a change to your source code.

Nipun fixed this problem in #261 by using [build constraints][bc]. `weaver_gen.go` files are not built when the `ignoreWeaverGen` tag is present, and `weaver generate` includes this tag. However, in #383, @songsy13 pointed out that this code wasn't working.

After some debugging, I realized it was the simple mistake of misformating the build constraint comments as `// go:build` instead of `//go:build` (note the extra space after the `//`). I'm very surprised we hadn't noticed this earlier!

[bc]: https://pkg.go.dev/cmd/go#hdr-Build_constraints


internal/tool/generate/generator.go
```diff
@@ -864,7 +864,7 @@ func (g *generator) componentRef(comp *component) string {
 
 // generateImports generates code to import all the dependencies.
 func (g *generator) generateImports(p printFn) {
-	p("// go:build !ignoreWeaverGen")
+	p("//go:build !ignoreWeaverGen")
```
spetrovic77, line 867: 😱

spetrovic77 (approved):
Thanks for fixing Michael!

PR #388, opened 2023-06-12
Encode listeners as strings inside the application binary.

spetrovic77:
This will allow deployers to extract the list of listeners and the associated component from the binary without running it This method is similar to our existing encoding of component graph edges.


internal/tool/generate/generator.go
```diff
@@ -422,6 +423,27 @@ func extractComponent(opt Options, pkg *packages.Package, file *ast.File, tset *
 					formatType(pkg, arg))
 			}
 			refs = append(refs, named)
+		} else if isWeaverListener(t) {
+			var lisName string
+			if f.Tag != nil {
+				tag := f.Tag.Value
```
mwhittaker, line 429: Would [`reflect.StructTag(tag).Get("weaver")`](https://pkg.go.dev/reflect#StructTag.Get) work here?

internal/tool/generate/generator.go
```diff
@@ -422,6 +423,27 @@ func extractComponent(opt Options, pkg *packages.Package, file *ast.File, tset *
...
+				const tagSuffix = "\"`"
+				if !strings.HasPrefix(tag, tagPrefix) || !strings.HasSuffix(tag, tagSuffix) {
+					return nil, errorf(pkg.Fset, f.Pos(),
+						"Listener tag must be of the format `weaver:\"name\"`, got %s",
+						tag)
+				}
+				lisName = strings.TrimSuffix(strings.TrimPrefix(tag, tagPrefix), tagSuffix)
+			} else if f.Names == nil { // embedded field
+				lisName = "Listener"
+			} else if len(f.Names) > 1 { // should never happen?
```
mwhittaker, line 440: I think this might happen when you do something like the following? 

```go
type foo struct {
    weaver.Implements[Foo]
    x, y weaver.Listener
}
```

rgrandl (approved):
Nice, thanks Srdjan!

internal/tool/generate/generator.go
```diff
@@ -422,6 +423,27 @@ func extractComponent(opt Options, pkg *packages.Package, file *ast.File, tset *
...
+			if f.Tag != nil {
+				tag := f.Tag.Value
+				const tagPrefix = "`weaver:\""
+				const tagSuffix = "\"`"
+				if !strings.HasPrefix(tag, tagPrefix) || !strings.HasSuffix(tag, tagSuffix) {
+					return nil, errorf(pkg.Fset, f.Pos(),
+						"Listener tag must be of the format `weaver:\"name\"`, got %s",
+						tag)
+				}
+				lisName = strings.TrimSuffix(strings.TrimPrefix(tag, tagPrefix), tagSuffix)
```
mwhittaker, line 437: Do we need to check that these tag names look like what we expect. For example, a user might do something like 


```go
type foo struct {
    weaver.Implements[Foo]
    x weaver.Listener `weaver:" 1 2 3 "`
}
```

runtime/codegen/listeners.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// The list of listener names associated with the component.
+	Listeners []string
+}
+
+// ExtractListeners returns the components and their listeners encoded using
+// MakeListenersString() in data.
+func ExtractListeners(data []byte) []ComponentListeners {
+	var results []ComponentListeners
+	re := regexp.MustCompile(`⟦([0-9a-fA-F]+):wEaVeRlIsTeNeRs:([a-zA-Z0-9\-.~_/]*?)→([a-zA-Z0-9\-.~_/,]*?)⟧`)
```
mwhittaker, line 62: I don't think it's super important, but we're deriving listener names from Go identifier names and Go identifiers can be more than `[a-zA-Z0-9\-.~_/]*?`: https://go.dev/ref/spec#Identifiers

mwhittaker (commented):
Sorry for the annoying drive-by comments.

internal/tool/generate/generator.go
spetrovic77: Unfortunately, it wouldn't, since this `Tag` includes the single quotes. However, I just realized that the user may attach other tags (e.g., ```x weaver.Listener `weaver:"foo",json:"bar"` ```). I'll change this in a separate PR.

internal/tool/generate/generator.go
spetrovic77: Ack, will do in a separate PR.

PR #394, opened 2023-06-14
Modify the listener format for local listeners.

spetrovic77:
This fixes the weirdness of the "local_address" field, which is only used by multi/single/ssh deployers, but not others (e.g., GKE, kube). The plan is to move the listener options into the deployer config sections, since they are largely different.

Based on discussions with @rgrandl and @mwhittaker.

Other changes:
* Simplify the parsing of the listener config, by reading into a `map[string]ListenerOptions`. This obviates the need for using the `toml.Primitive` type.

TODO: Change the format of the GKE/kube deployers to be similar.


website/docs.md
```diff
@@ -1705,8 +1702,15 @@ the runtime:
    listener. In fact, the proxy balances traffic across every replica of the
    listener. (Recall that components may be replicated, and so every component
    replica will have a different instance of the listener.)
-   The proxy address is by default `localhost:0`, or the local address assigned
-   to the listener in the config file, if any.
+
+The proxy address is by default `localhost:0`, unless a concrete address
```
mwhittaker, line 1706: Oops, I think I wrote this wrong originally. The proxy default to `":0"`, not `"localhost:0"`, right?

mwhittaker (approved):
Looks awesome!

website/docs.md
spetrovic77: Thanks, changed!

PR #395, opened 2023-06-14
Added examples page to site.

mwhittaker:
[Examples Page.webm](https://github.com/ServiceWeaver/weaver/assets/3654277/5a979732-aec3-4df1-b024-ff93ea913caf)

I'll work on adding live demos for emoji and online boutique. I'll also work on moving bank of anthos over.


spetrovic77 (approved):
Very nice!

Can we start with emoji search and online boutique. They would do better to convey that you can run "real" production apps using Service Weaver.

Actually, I would prefer that the reverser isn't included. It's a bit of a toy example. WDYT?

mwhittaker:
> Very nice!
> 
> Can we start with emoji search and online boutique. They would do better to convey that you can run "real" production apps using Service Weaver.
> 
> Actually, I would prefer that the reverser isn't included. It's a bit of a toy example. WDYT?

Great point, totally agree. I removed the reverser app.

PR #400, opened 2023-06-15
Small example fixes.

mwhittaker:
1. I added factors and onlineboutique to `examples_test.go`. Adding chat is hard because it requires a mysql database.
2. I made all examples `log.Fatal` the error returned by `weaver.Run`. Before, there was a mix of doing nothing, using `log.Fatal`, and writing to `os.Stderr`.
3. Quiet `examples_test.go` when -v isn't provided.

### DIFF
--- a/examples/chat/main.go
+++ b/examples/chat/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"context"
 	"flag"
+	"log"
 
 	"github.com/ServiceWeaver/weaver"
 )
@@ -25,5 +26,7 @@ import (
 
 func main() {
 	flag.Parse()
-	weaver.Run(context.Background())
+	if err := weaver.Run(context.Background()); err != nil {
+		log.Fatal(err)
+	}
 }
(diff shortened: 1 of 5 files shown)


spetrovic77 (approved):
Looks great. For tests, we could create a sqlite database in the test temp dir, then modify the config. LMK if you think it's worth it.

Btw., the chat app seems to be creating an in-memory sqlite database, which is wrong when multi is used. It pretty much only works for single. It's okay I guess, but a bit subtle. At the minimum we should change the instructions and add a multi config.

mwhittaker:
> Looks great. For tests, we could create a sqlite database in the test temp dir, then modify the config. LMK if you think it's worth it.
> 
> Btw., the chat app seems to be creating an in-memory sqlite database, which is wrong when multi is used. It pretty much only works for single. It's okay I guess, but a bit subtle. At the minimum we should change the instructions and add a multi config.

Got it. Let me fix up the chat app and add it to the tests in another PR then!

PR #401, opened 2023-06-15
Added `weaver callgraph` command.

mwhittaker:
This PR introduces a `weaver callgraph` command that outputs a [Mermaid][] diagram of the component call graph. I also updated every example's README with its callgraph diagram. See https://github.com/ServiceWeaver/weaver/blob/callgraph/examples/onlineboutique/README.md for an example.

[Mermaid]: https://mermaid.js.org/syntax/flowchart.html

### DIFF
--- a/cmd/weaver/main.go
+++ b/cmd/weaver/main.go
@@ -25,6 +25,7 @@ import (
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


spetrovic77 (approved):
Very cool! :sunglasses:

PR #405, opened 2023-06-20
Don't automatically run Main from weavertests.

mwhittaker:
## Overview

In #359, we changed weavertest to automatically run the main component's Main method for every test. This was done to allow a developer to test the server they started in the Main method.

However, we realized that this approach of automatically running the Main method has some drawbacks. Because the Main method is run for every test, if you do anything in the Main method that cannot be run more than once, the tests will fail. For example, if you register a handler with the default serve mux using `http.HandleFunc`, all tests after the first will panic.

After some offline discussion, we decided not to run the Main method automatically. Additionally, we'll allow people to get pointers to component implementations in a test. Then, developers can test the guts of their Main method using existing things like the httptest package. This PR changes weavertest to not run the Main method automatically. In a future CL, I'll allow developers to get pointers to component implementations.

## Details

- I removed `weavertest.GetListenerAddress`. It will no longer be needed.
- I removed the `apps` variable from the `weavertest` package. It was only needed to implement `GetListenerAddress`.
- I removed `testMainInterface` from the `weavertest` package. I think this was some stale code lingering around from an earlier implementation of `weavertest`.


ghemawat (approved):
Looks great! I like the simplification we got as well.

PR #406, opened 2023-06-20
Let weavertest test component implementations.

mwhittaker:
Before this PR, you could use a weavertest.Runner to test a set of component interfaces:

```go
runniner.Test(t, func(t *testing.T, a A, b B, c C) {
    ...
})
```

This PR extends the API to allow for testing of component implementations. Assuming `a`, `b`, and `c` are the implementing structs of component interfaces `A`, `B`, and `C`, we can now write:

```go
runniner.Test(t, func(t *testing.T, a *a, b *b, c *c) {
    ...
})
```

You can also combine interfaces and implementations:

```go
runniner.Test(t, func(t *testing.T, a *a, b B, c1 *c, c2 *c, cIntf C) {
    ...
})
```

You can also combine this with fakes. All possible combinations of interfaces, implementation pointers, and fakes are allowed with one exception. A fake and implementation pointer cannot be provided for the same component.

This functionality is useful for clearbox testing of components. It is also useful for testing the main HTTP handling code associated with the main component, as demonstrated in `examples/chat/server_test.go`.


weavertest/init.go
```diff
@@ -142,11 +149,23 @@ func (r Runner) Bench(b *testing.B, testBody any) {
 
 func (r Runner) sub(t testing.TB, isBench bool, testBody any) {
 	t.Helper()
-	runner, err := checkRunFunc(t, testBody)
+	body, intfs, err := checkRunFunc(t, testBody)
 	if err != nil {
 		t.Fatal(fmt.Errorf("weavertest.Run argument: %v", err))
 	}
 
+	// Assume a component Foo with implementing struct foo. We disallow tests
```
ghemawat, line 157: s/with implementing/implementing/

weavertest/init.go
```diff
@@ -142,11 +149,23 @@ func (r Runner) Bench(b *testing.B, testBody any) {
...
+	// like the one below where the user provides a fake and a component
+	// implementation pointer for the same component.
```
ghemawat, line 159: Why do we need to disallow this?

weavertest/internal/chain/chain_test.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			if got, want := d.val, 4; got != want {
+				t.Fatalf("got %d, want %d", got, want)
+			}
+		})
+	}
+}
+
+func TestThreeComponentImpls(t *testing.T) {
+	// Tests weaver.Test with three component implementation pointer arguments.
```
ghemawat, line 70: What code are we covering here that wasn't covered by the two component test?

ghemawat (approved):
Great!

component.go
```diff
@@ -131,6 +131,13 @@ type Main interface {
 // implementation type instance.
 type Implements[T any] struct {
 	*componentImpl
+
+	// Given a component implementation type, there is currently no nice way,
+	// using reflection, to get the corresponding component interface type [1].
+	// The __component_interface_type field exists to make it possible.
+	//
+	// [1]: https://github.com/golang/go/issues/54393.
+	__component_interface_type T //nolint:unused
```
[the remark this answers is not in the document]
mwhittaker: We don't. Deleted.

weavertest/init.go
mwhittaker: Done.

weavertest/init.go
mwhittaker: We could allow it! In code like the following, the fake is unused because of the presence of the `*foo` argument. I thought it was better to fail the test rather than letting the fake go silently unused? What do you think? Happy to allow it.

```go
runner.Fakes = append(runner.Fakes, weavertest.Fake[Foo](...))
runner.Test(t, func(t *testing.T, impl *foo) {...})
```

weavertest/internal/chain/chain_test.go
mwhittaker: None. I removed `D` completely and simplified things to a chain of `A`, `B`, and `C`.

mwhittaker (commented):
Thanks Sanjay!

ghemawat (approved):

PR #409, opened 2023-06-20
Removed Main; changed weaver.Run to take lambda.

mwhittaker:
This PR reverts #359. In particular, it removes the main component's Main method and changes `weaver.Run` to take a lambda.

We made this change for the following reason. The Main method is run by `weaver.Run`, but not by `weavertest.Runner.Test`. We felt this was confusing. With the lambda API, it is much clearer what is run. This change also unifies the `weaver.Run` and `weavertest.Runner.Test` API. Both take a lambda with component arguments.

Right now, the signature of `weaver.Run` is a bit gnarly. I think I can simplify it a bit in a future PR to remove PointerToMain at least.


ghemawat (approved):
Thanks.

Issue #414, opened 2023-06-25
`weaver generate`  fails when a component uses a router whose routing key is a struct


zailic:
Let's presume the following `Cache` component associated with routing behaviour.

```
...

// Cache caches emoji query results.
type Cache interface {
	// Get returns the cached emojis produced by the provided query
        // field within cacheRoutingKey. On cache miss, Get returns nil, nil.
	Get(context. Context, cacheRoutingKey) ([]string, error)

	// Put stores a query and its corresponding emojis in the cache.
	Put(context.Context, cacheRoutingKey, []string) error
}

type cacheConfig struct {
	Namespaces []string `toml:"namespaces"`
}

// cache implements the Cache component.
type cache struct {
	weaver.Implements[Cache]
	weaver.WithRouter[router]
	weaver.WithConfig[cacheConfig]

	mu     sync.Mutex
	emojis map[string]map[string][]string
}

type cacheRoutingKey struct {
	ns    string
	query string
}

var _ Cache = (*cache)(nil)

func (c *cache) Init(context.Context) error {
	ns := c.Config().Namespaces
	c.emojis = map[string]map[string][]string{}
	for _, n := range ns {
		c.emojis[string(n)] = map[string][]string{}
	}
	return nil
}

func (c *cache) Get(_ context.Context, rk cacheRoutingKey) ([]string, error) {
	c.mu.Lock()
	defer c.mu.Unlock()
	c.Logger().Debug("Get", "ns", rk.ns, "query", rk.query)
	if _, ok := c.emojis[rk.ns]; !ok {
		return nil, fmt.Errorf("cache namespace key:'%q' - not defined", rk.ns)
	}
	return c.emojis[rk.ns][rk.query], nil
}

func (c *cache) Put(_ context.Context, rk cacheRoutingKey, emojis []string) error {
	c.mu.Lock()
	defer c.mu.Unlock()
	c.Logger().Debug("Put", "ns", rk.ns, "query", rk.query)
	if _, ok := c.emojis[rk.ns]; !ok {
		return fmt.Errorf("cache namespace key:'%q' - not defined", rk.ns)
	}
	c.emojis[rk.ns][rk.query] = emojis
	return nil
}

// router routes requests to the Cache component. Both Get and Put use the
// query as the routing key. Calls to these methods with the same routing key will
// tend to be routed to the same replica.
type router struct{}

func (router) Get(_ context.Context, rk cacheRoutingKey) cacheRoutingKey {
	return rk
}

[...]

PR #415, opened 2023-06-25
Fix weaver generate process when routing key of Router component is a struct

zailic:
More details in https://github.com/ServiceWeaver/weaver/issues/414 issue.

### DIFF
--- a/internal/tool/generate/types.go
+++ b/internal/tool/generate/types.go
@@ -899,8 +899,10 @@ func isPrimitiveRouter(t types.Type) bool {
 }
 
 // isValidRouterType returns whether the provided type is a valid router type.
-// A router can be an integer (signed or unsigned), a float, or a string. Or,
-// it can be a struct where every field is an integer, float, or string.
+// A router type can be one of the following: an integer (signed or unsigned),
+// a float, or a string. Alternatively, it can be a struct that may optioanly
+// embed the weaver.AutoMarshal struct and rest of the fields must be either
+// integers, floats, or strings.
 func isValidRouterType(t types.Type) bool {
 	t = t.Underlying()
 	if isPrimitiveRouter(t) {
@@ -911,7 +913,8 @@ func isValidRouterType(t types.Type) bool {
 		return false
 	}
 	for i := 0; i < s.NumFields(); i++ {
-		if !isPrimitiveRouter(s.Field(i).Type()) {
+		ft := s.Field(i).Type()
+		if !isPrimitiveRouter(ft) && !isWeaverAutoMarshal(ft) {
 			return false
 		}
 	}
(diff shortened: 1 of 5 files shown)


mwhittaker (commented):
Amazing job! Would you mind updating the documentation of `isValidRouterType` and also the documentation of valid router key types in `component.go` and `website/docs.md`? You can also leave that to me if you'd like :)

zailic:
Thanks for your quick feedback! I just updated documentation as per your request. Feel free to refine it as my English is not as good as yours! :D

mwhittaker (approved):
Awesome, thanks so much for your contribution!

mwhittaker:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/ServiceWeaver/weaver/pull/415/checks?check_run_id=14538281254) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

@zailic I think you'll have to complete these instructions before I can merge in the PR.

internal/tool/generate/types.go
mwhittaker, line 1: Oh, I actually think your original approach is correct. A router key type does not need to embed AutoMarshal; it is never actually marshaled. But, ignoring an embedded `weaver.AutoMarshal` makes sense to me.

zailic: Hmm, then things are more complicated than I thought 😆. Ultimately the router key should be passed as an input argument of one of the method of a weaver component which is associated with the router. Currently the `weaver generate process` requires that the methods of weaver components should accept only serializable arguments - https://github.com/ServiceWeaver/weaver/blob/main/internal/tool/generate/generator.go#L666-L675 . Therefore to satisfy the requirement above I think it's mandatory that the router key does need to embed AutoMarshal

mwhittaker: Ah, I think I see the confusion. Router keys do not have to be method arguments. For example, we could re-write your example in #414 to the following: 


```go
type Cache interface {
    Get(ctx context. Context, namespace, key string) ([]string, error)	
    Put(ctx context.Context, namespace, key string, emojis []string) error
}

// Cache implementation...

type router struct{}

type routingKey struct {
    namespace string
    key string
}

func (router) Get(_ context.Context, namespace, key string) routingKey {
	return routingKey{namespace, key}
}

func (router) Put(_ context.Context, namespace, key string, _ []string) routingKey {
	return routingKey{namespace, key}
}
```

mwhittaker: But I think your example made a great point that it's probably not uncommon to use a method argument as a routing key, and allowing `weaver.AutoMarshal` in keys is easy and relatively intuitive, so I think it makes sense to support it.

zailic: Ahh, you're right...I should have read the documentation more carefully. Thanks for the clarification 🙏!

mwhittaker:
Hey @zailic, mind squashing and rebasing on the main branch?

PR #418, opened 2023-06-26
Added listener names to Registration.

mwhittaker:
This PR changes the generator to include listener names in `codegen.Registration`s. It also changes the singleprocess deployer to use these registered listeners, rather than the listeners embedded in a binary, when validating the listeners specified in a config file. See PR #413 for more information.


spetrovic77 (approved):
Ooops, just noticed the other fix as well. Thanks for fixing.

PR #420, opened 2023-06-27
Fix metrics benchmark measurement of Bytes stats maintenance.

ghemawat:


mwhittaker (approved):
Thanks Sanjay! Good catch.

PR #422, opened 2023-06-27
Fixed main-only bug in multi deployer.

mwhittaker:
Before this PR, if you tried to `weaver multi deploy` a Service Weaver application that only had a main component, it would error out immediately with a "github.com/ServiceWeaver/weaver/Main not found".

Here's what was happening. `weaver multi deploy` read the component callgraph from the application binary to know which components existed. Reading the callgraph returns the edges in the graph. However, for an application with only a main component, there are no edges. Thus, `weaver multi deploy` erroneously concluded that there are no components.

This PR makes a dirt simple fix. We may want to revise the callgraph API to return nodes and edges in a future PR to make this problem harder to run into?

Also, sadly, applications that exit from main still exhibit odd behavior when deployed with `weaver multi deploy`. The hello world application added in this PR for example, doesn't print anything. I believe we aren't waiting to flush all messages over the pipe, but I'll have to debug further in a future PR.


mwhittaker:
> Wowza! Can't believe we didn't run into this before.

Yeah, me too!

PR #427, opened 2023-06-28
Added test to check forgotten codegen updates.

mwhittaker:
In PR #388, we changed `weaver generate` to embed listener information in compiled Service Weaver binaries. This change landed in v0.15.0. However, we forgot to update the codegen version. This led to the following bug, reported by @renanbastos93 in the Discord:

- Run `weaver generate` at version v0.14.0.
- Run `go run .` at version v0.16.1.
- The app crashes looking for listener information in the binary, but the information is not there because of the stale `weaver generate`.

If we had updated the codegen version, then the code would not have compiled, and the user would have received an error message explaining that they needed to update `weaver generate`.

It's really easy to forget to update the codegen version, so I added a unit test that checks for changes to how `weaver_gen.go` files are generated. If there is any change, the test fails and you're prompted to update the codegen version.


renanbastos93 (approved):
LGTM

spetrovic77 (approved):
Thank you for fixing!

mwhittaker:
I chatted with Sanjay about this test a bit. An alternative is to have a separate script that compares the value of the `weaver_gen.go` files on HEAD with the ones at the latest tag (which can be fetched with `go mod download -json github.com/ServiceWeaver/weaver@latest). If the `weaver_gen.go` files differ, but the codegen API version does not differ, then the script fails. We can run the script as part of `build_and_test`. We can also run `apidiff` to check for changes to the codegen API itself. We can do something similar for the deployer API by checking for changes in `runtime.proto` and the envelope API itself. 

The tricky thing is that some changes to `weaver_gen.go` files might be inconsequential (e.g., changing variable names or adding comments), and the script would keep flagging these unimportant changes. I'll think about this some more. If anyone has ideas, let me know :)

PR #429, opened 2023-06-28
Restore metric collection for local calls.

ghemawat:
We had previously restricted metric collection to just remote calls for performance reasons. We now collect metrics for local calls as well since it was problematic to hide things like call counts for local calls.

This has a performance impact: on a typical machine, a local call slows down by ~100ns. By comparison, a local call slows down by 1000ns when tracing is enabled for that call.  Future changes will attempt to speed up metric collection (we hope we can drop the penalty by 10x).

Changes the method metric names from "serviceweaver_remote_*" to "serviceweaver_*".

Include whether or not a call is remote in the metric label.

Updated the codegen metric collection API. Mainly, the responsibility for updating metrics is now in the codegen package as opposed to scattered throughout generated code.

Not updating the codegen version number since we have alread updated it since the last tagged release.


mwhittaker (approved):
Awesome!

PR #431, opened 2023-06-29
Expanded codegen version mismatch error.

mwhittaker:
Recall that `weaver generate` embeds its codegen API version in the `weaver_gen.go` files it generates. If these files are compiled with a different codegen API version, the code fails to build, and a helpful error message is shown to the user. Before this PR, the error message looked like this:

```
You used 'weaver generate' codegen version 0.17.0, but you built your
code with an incompatible weaver module version. Try upgrading 'weaver
generate' and re-running it.
```

Now, it looks like this:

```
ERROR: You generated this file with 'weaver generate' v0.17.0 (codegen
version v0.17.0). The generated code is incompatible with the version of the
github.com/ServiceWeaver/weaver module that you're using. The weaver module
version can be found in your go.mod file or by running the following command.

    go list -m github.com/ServiceWeaver/weaver

We recommend updating the weaver module and the 'weaver generate' command by
running the following.

Then, re-run 'weaver generate' and re-build your code. If the problem persists,
please file an issue at https://github.com/ServiceWeaver/weaver/issues.
```

The old error message allowed people to ask us for help, and we could tell them how to fix it. The new error message should hopefully allow people to understand and fix the bug on their own.

### DIFF
--- a/examples/chat/weaver_gen.go
+++ b/examples/chat/weaver_gen.go
@@ -14,7 +14,26 @@ import (
 	"reflect"
 	"time"
 )
-var _ codegen.LatestVersion = codegen.Version[[0][17]struct{}]("You used 'weaver generate' codegen version 0.17.0, but you built your code with an incompatible weaver module version. Try upgrading 'weaver generate' and re-running it.")
+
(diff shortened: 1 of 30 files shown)


spetrovic77 (approved):
Nice!

PR #432, opened 2023-06-29
Improved deployer version mismatch error.

mwhittaker:
Recall that when a deployer deploys a binary, it first extracts the deployer API version from the binary and checks that its deployer API version is compatible. Before this PR, the resulting error message looked like this:

```
version mismatch: deployer's deployer API version v0.17.0 is
incompatible with app' deployer API version v0.14.0
```

Now, it looks like this:

```
ERROR: The binary you're trying to deploy ("collatz") was built with
github.com/ServiceWeaver/weaver module version v0.18.0. However, the 'weaver 
multi' binary you're using was built with weaver module version v0.17.0. 
These versions are incompatible.

We recommend updating both the weaver module your application is built with and
updating the 'weaver multi' command by running the following.

Then, re-build your code and re-run 'weaver multi deploy'. If the problem
persists, please file an issue at https://github.com/ServiceWeaver/weaver/issues.
```

This PR also makes the multi and ssh deployers check these versions earlier, before trying to deploy anything.

See #431 for a similar PR.

### DIFF
--- a/internal/tool/multi/deploy.go
+++ b/internal/tool/multi/deploy.go
@@ -24,16 +24,19 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"syscall"
 
 	"github.com/ServiceWeaver/weaver/internal/status"
 	"github.com/ServiceWeaver/weaver/internal/tool/config"
 	"github.com/ServiceWeaver/weaver/runtime"
+	"github.com/ServiceWeaver/weaver/runtime/bin"
 	"github.com/ServiceWeaver/weaver/runtime/codegen"
 	"github.com/ServiceWeaver/weaver/runtime/colors"
 	"github.com/ServiceWeaver/weaver/runtime/logging"
 	"github.com/ServiceWeaver/weaver/runtime/retry"
 	"github.com/ServiceWeaver/weaver/runtime/tool"
+	"github.com/ServiceWeaver/weaver/runtime/version"
 	"github.com/google/uuid"
 )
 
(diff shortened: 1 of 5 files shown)


internal/tool/multi/deploy.go
```diff
@@ -84,6 +87,37 @@ func deploy(ctx context.Context, args []string) error {
...
+		// Try to relativize the binary, defaulting to the absolute path if
+		// there are any errors..
+		binary := appConfig.Binary
+		if cwd, err := os.Getwd(); err == nil {
+			if rel, err := filepath.Rel(cwd, appConfig.Binary); err == nil {
+				binary = rel
+			}
+		}
+		return fmt.Errorf(`ERROR: The binary you're trying to deploy (%q) was built with
+github.com/ServiceWeaver/weaver module version %s (deployer API
```
spetrovic77, line 105: Remove all references of `deployer API`. They should just know that version of code and version of `weaver` are incompatible.

internal/tool/ssh/deploy.go
```diff
@@ -92,6 +94,37 @@ func deploy(ctx context.Context, args []string) error {
...
+	if versions.DeployerVersion != version.DeployerVersion {
+		// Try to relativize the binary, defaulting to the absolute path if
+		// there are any errors..
+		binary := app.Binary
+		if cwd, err := os.Getwd(); err == nil {
+			if rel, err := filepath.Rel(cwd, app.Binary); err == nil {
+				binary = rel
+			}
+		}
+		return fmt.Errorf(`ERROR: The binary you're trying to deploy (%q) was built with
```
spetrovic77, line 111: Ditto

spetrovic77 (approved):
Nice!

internal/tool/multi/deploy.go
mwhittaker: Done.

internal/tool/ssh/deploy.go
mwhittaker: Done.

mwhittaker (commented):
Thanks Srdjan!

PR #437, opened 2023-06-30
Introduce a size limit to the Perfetto database.

spetrovic77:
If the trace data exceeds this limit, we garbage-collect the oldest
trace entry.
    
DB size is computed using the database file size (thanks sanjay@),
in order to be able to compute the size efficiently.
    
When the DB limit is reached, the trace store operation takes 2x time
(but not more than that).
    
    ```
    cpu: Intel(R) Xeon(R) Gold 6154 CPU @ 3.00GHz
    BenchmarkStore
    BenchmarkStore/nolimit_small
    BenchmarkStore/nolimit_small-72                      100          10830029 ns/op
    BenchmarkStore/limit_small
    BenchmarkStore/limit_small-72                         54          21509631 ns/op
    BenchmarkStore/nolimit_medium
    BenchmarkStore/nolimit_medium-72                     100          10694646 ns/op
    BenchmarkStore/limit_medium
    BenchmarkStore/limit_medium-72                        54          22720641 ns/op
    BenchmarkStore/nolimit_big
    BenchmarkStore/nolimit_big-72                        100          11460602 ns/op
    BenchmarkStore/limit_big
    BenchmarkStore/limit_big-72                           54          21542238 ns/op
    ```
    
NOTE:
We tried the trigger-based mechanism (thanks mwhittaker@ !) but the complex
SQL logic expressed in the trigger was way too expensive to perform on
every insertion.

### DIFF
--- a/internal/tool/multi/deployer.go
+++ b/internal/tool/multi/deployer.go
@@ -133,7 +133,11 @@ func newDeployer(ctx context.Context, deploymentId string, config *MultiConfig) 
 	}
 
 	// Create the trace saver.
-	traceDB, err := perfetto.Open(ctx, perfettoFile)
+	traceDB, err := perfetto.Open(ctx, perfettoFile, perfetto.DBOptions{
+		// 100MB limit, which allows for storage of upto 200k traces at 500
+		// bytes each.
+		MaxTraceBytes: 100 * 1024 * 1024,
+	})
 	if err != nil {
 		return nil, fmt.Errorf("cannot open Perfetto database: %w", err)
 	}
--- a/runtime/perfetto/db.go
+++ b/runtime/perfetto/db.go
@@ -81,12 +81,12 @@ type replicaCacheKey struct {
 // [1] https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview#
 // [2] https://ui.perfetto.dev/
 type DB struct {
+	opts DBOptions
+
 	// Trace data is stored in a sqlite DB spread across three tables:
 	// (1) traces:           trace data in a Perfetto-UI-compattible JSON format
 	// (2) replica_num:      map from colocation group replica id to a replica
 	//                       number
-	// (3) next_replica_num: the next replica number to use for a given
-	//                       colocation group
 	fname string
 	db    *sql.DB
 
@@ -94,9 +94,19 @@ type DB struct {
 	replicaNumCache *lru.Cache[replicaCacheKey, int]
 }
 
+// DBOptions specifies options for the database.
+type DBOptions struct {
+	// MaxTraceBytes, if non-zero, specifies the maximum size of trace
+	// data stored in DB.
+	//
+	// Once the trace data exceeds this limit, the DB will start deleting
+	// its oldest traces.
+	MaxTraceBytes int64
+}
+
 // Open opens the default trace database on the local machine, which is
 // persisted in the provided file.
-func Open(ctx context.Context, fname string) (*DB, error) {
+func Open(ctx context.Context, fname string, opts DBOptions) (*DB, error) {
 	if err := os.MkdirAll(filepath.Dir(fname), 0700); err != nil {
 		return nil, err
 	}
@@ -120,6 +130,7 @@ func Open(ctx context.Context, fname string) (*DB, error) {
 	}
 
 	t := &DB{
+		opts:            opts,
 		fname:           fname,
 		db:              db,
 		replicaNumCache: cache,
@@ -142,6 +153,7 @@ CREATE TABLE IF NOT EXISTS replica_num (
 	PRIMARY KEY(app,version,weavelet_id)
 );
 `
+
 	if _, err := t.execDB(ctx, initTables); err != nil {
 		return nil, fmt.Errorf("open trace DB %s: %w", fname, err)
 	}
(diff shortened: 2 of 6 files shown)


ghemawat:
Before getting into the guts of the changes, I want to ask if it would be better to make the limit time based. E.g., only keep traces for the last N days.

Pro: I suspect it might be more useful for users; they can set a single default (e.g., 30 days) without worrying about what rate of traffic the service gets and how big each trace is.

Con: disk may fill up. 

I weigh the pro more since in the vast majority of cases the users will be happy with our default time limit, and it will be very unlikely they will run into disk space full issue. Even the case that prompted this PR did not result in a disk-full error, just a somewhat larger than expected file.

ghemawat:
Also, consider what the system default should be: with time-based limits, I would expect something like a month or a year. I don't quite know how to set the default for a size based limit without knowing about the service stats, which we don't when setting a system-wide default.

spetrovic77:
Another `Pro` for the time-based control is that it will be simpler to implement. Keeping track of size efficiently has been a  bit of a pain.

However, I am quite strongly in favor of size-based limit because for the local case, my guess is that's what folks would care about the most. Even you have brought up this issue offline by saying "the database file has grown too big". I believe it filled up quickly because of your testing/benchmarking, and not because it had records in there older than 1 month.

I feel like unless we limit the size (i.e., the number of trace records in there), any other limit is meaningless.

spetrovic77:
> Also, consider what the system default should be: with time-based limits, I would expect something like a month or a year. I don't quite know how to set the default for a size based limit without knowing about the service stats, which we don't when setting a system-wide default.

Yeah, defaults are tricky. I think that 30 day limit is reasonable as a default. With size it's trickier.

How about we set some large size limit (say, 1GB), and use it in conjunction with a time-based limit?

ghemawat:
I think we will add sampling on the tracing side, perhaps a time based one like once a second on average (so that infrequently used services also get some traces). Given that, I am not so worried about the size getting out of hand (the rate of accumulation will be bounded to trace-size bytes per second).

I was prompted to ask about time-based since it would be simpler (I noticed the need to keep size stats in your PR).

Could we go with just time-based for now for simplicity? (More important to get sampling of traces done first)

spetrovic77:
I'm going to try the simplest, trigger-based, time limit solution and we can decide from there.

mwhittaker:
> I'm going to try the simplest, trigger-based, time limit solution and we can decide from there.

This sounds good to me! I think if the trigger-based approach isn't slow, then it's the simplest, and we should go with that. 

If it is too slow,  I think we could simplify the size-based approach by using built-in size metadata (https://stackoverflow.com/a/58251635) or by garbage collecting only periodically to avoid having to maintain size stats. At that point, the size-based and time-based approach seem similarly complex to me? And I do slightly prefer garbage collecting based on size rather than time. 

If the time-based approach does end up simpler though, I definitely support doing the simplest thing for now!

ghemawat:
We could simplify the size computation if we want by using the size of the DB file? But perhaps it does not shrink predictably after a bulk deletion and we would repeatedly attempt to shrink it?

spetrovic77:
There is also a possibility of using the `dbstat` table. It wasn't working
for me from Go code, but it does work from `sqlite3` command-line.

https://www.sqlite.org/dbstat.html

I can try to get it to work.

spetrovic77:
Here is the time-based PR using triggers. It's very simple. I'll try to
simplify the space-based solution and we can make a decision.

https://github.com/ServiceWeaver/weaver/pull/439

internal/tool/multi/deployer.go
```diff
@@ -133,7 +133,9 @@ func newDeployer(ctx context.Context, deploymentId string, config *MultiConfig)
 	}
 
 	// Create the trace saver.
-	traceDB, err := perfetto.Open(ctx, perfettoFile)
+	traceDB, err := perfetto.Open(ctx, perfettoFile, perfetto.DBOptions{
+		MaxTraceBytes: 100 * 1024 * 1024, // 100MB
```
ghemawat, line 137: Is there some reasoning we can write down in a comment for picking this size? Perhaps based on every trace size?

runtime/perfetto/db.go
```diff
@@ -142,6 +156,17 @@ CREATE TABLE IF NOT EXISTS replica_num (
 	PRIMARY KEY(app,version,weavelet_id)
 );
 `
+	if opts.MaxTraceBytes > 0 {
+		initTables += `
+-- Single-row table that stores the total size of all trace data.
```
ghemawat, line 161: We should try to get rid of this size maintenance. Perhaps by looking at the size of the perfetto.db file?

Apart from code complexity, this is also imposes an invariant across tables that could be violated by somebody coming along and deleting or inserting trace data directly.

runtime/perfetto/db.go
```diff
@@ -164,14 +189,91 @@ func (d *DB) Store(ctx context.Context, app, version string, spans []sdktrace.Re
...
+	if totalBytes < d.opts.MaxTraceBytes {
+		return nil
+	}
+
+	// Shrink trace data down to MaxTraceBytes / 2, to amortize the
+	// deletion cost.
+	toDeleteBytes := totalBytes - (d.opts.MaxTraceBytes / 2)
+
+	// Compute the trace length for every trace row, in insertion order.
+	const lenQuery = `SELECT rowid, length(events) FROM traces ORDER BY insertion_time`
```
ghemawat, line 251: Could this be a massive temporary spike in resource usage (infrequent, but could hurt latency sensitive ops).

spetrovic77:
Given the simplicity of the trigger-based solution, I'd say let's go with
that.

I'll make it into a more of a "real" change.

runtime/perfetto/db.go
spetrovic77: Done, this is much cleaner now. PTAL.

runtime/perfetto/db.go
spetrovic77: Good point. Please look at the new solution which deletes a single row once the DB size crosses a threshold.

internal/tool/multi/deployer.go
spetrovic77: Done.

spetrovic77 (commented):
Thanks for the suggestions. By checking the file size, the code is much simpler. I've also made it backward compatible). PTAL.

spetrovic77:
Abandoning this change.

Per offline discussion, if we ever need to implement a size-related limiting, we can do it by using a more aggressive time-based limit.

PR #439, opened 2023-06-30
Implement time-based trace DB compaction using a trigger.

spetrovic77:


runtime/perfetto/db.go
```diff
@@ -94,9 +96,17 @@ type DB struct {
 	replicaNumCache *lru.Cache[replicaCacheKey, int]
 }
 
+// DBOptions specifies options for the database.
+type DBOptions struct {
+	// MaxTraceAge, if non-zero, specifies the maximum age of any trace stored
+	// in the database. Once older than this parameter, traces will be
+	// automatically deleted.
+	MaxTraceAge time.Duration
```
ghemawat, line 104: Where is the default value?

runtime/perfetto/db.go
```diff
@@ -120,14 +130,16 @@ func Open(ctx context.Context, fname string) (*DB, error) {
...
 		fname:           fname,
 		db:              db,
 		replicaNumCache: cache,
 	}
 
-	const initTables = `
+	initTables := `
 -- Trace data.
 CREATE TABLE IF NOT EXISTS traces (
+	insertion_time INTEGER NOT NULL,
```
ghemawat, line 142: What is the compatibility story for old DBs. Will users have to blow them away?

runtime/perfetto/db.go
```diff
@@ -120,14 +130,16 @@ func Open(ctx context.Context, fname string) (*DB, error) {
...
 	app TEXT NOT NULL,
 	version TEXT NOT NULL,
 	events TEXT NOT NULL
```
ghemawat, line 145: For future consideration: I just noticed the events field. Is that some text based rendering of the trace data? Would we be better off preserving the structure inherent in the trace entries by storing each event as a separate row with columns corresponding to event timestamp, etc.

runtime/perfetto/db.go
ghemawat, line 142: Is there some ordering imposed here that would allow the DB to find and delete old records in time proportional to the number of old records (and not proportional to the DB size)?

runtime/perfetto/db.go
```diff
@@ -142,6 +154,15 @@ CREATE TABLE IF NOT EXISTS replica_num (
 	PRIMARY KEY(app,version,weavelet_id)
 );
 `
+	if opts.MaxTraceAge > 0 {
+		initTables += fmt.Sprintf(`
+CREATE TRIGGER IF NOT EXISTS delete_expired AFTER INSERT ON traces
+BEGIN
+	DELETE FROM traces
+	WHERE insertion_time < unixepoch('now','-%d seconds');
+END;
```
ghemawat, line 163: What does this do to insertion speed?

ghemawat (commented):
Thanks! This looks nice and simple. I will also leave comments on the sibling PR.

spetrovic77:
Abandoning for now, since the tracing flow is going through a rewrite.

PR #440, opened 2023-06-30
Speed-up metric maintenance for component method calls.

ghemawat:
```
name           old time/op  new time/op  delta
Call/Local-12   119ns ± 2%    96ns ± 1%  -19.32%  (p=0.008 n=5+5)
```

Removed the use of expvar in Metrics implementation. We are mostly doing this so that we can drop the promise that Service Weaver metrics are reflected in expvars. That frees us to speed things up by using a better representation.

Introduced an atomicFloat64 value so we can atomically add a float64. (This functionality was previously provided by expvar.)

Atomic storage for all metrics (though used just for counters) is now split into a float64 and a uint64. Integer increments are applied to the uint64 so we can avoid a more expensive compare-and-swap. The stored values are added together on a read. Doubles counter metric increment speed (from ~9.2ns to ~4.1ns).

Add a fast-path for histogram updates for small values by avoiding binary search.

Add a fast-path for histogram updates of zero by avoiding an atomic add.

### DIFF
--- a/godeps.txt
+++ b/godeps.txt
@@ -835,7 +835,6 @@ github.com/ServiceWeaver/weaver/runtime/logging
     time
 github.com/ServiceWeaver/weaver/runtime/metrics
     encoding/binary
-    expvar
     fmt
     github.com/ServiceWeaver/weaver/runtime/protos
     github.com/google/uuid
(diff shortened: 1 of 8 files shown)


mwhittaker (approved):
Amazing! Thanks, Sanjay!

PR #444, opened 2023-07-05
Fixed bug in metric version management.

ghemawat:
Previous change forgot to update Metric.version on an Inc() call. Fixed by removing Metric.version and using the actual metric value to decide whether or not to export the metric.

Histograms gain a special putCount field that is used to determine whether or not to export since the histogram state does not fit into a single floating point value.

Further speed-up of local call metric maintenance: we now convert to microseconds once after finding the elapsed duration, instead of twice (once on the starting time, and once on the ending time).  This speeds up a local call from ~96ns to ~85ns.

Added test that Inc() call is reflected in subsequent export.


mwhittaker (approved):
Nice!

PR #445, opened 2023-07-06
Rename Metric.Init to Metric.initIdAndLabel.

ghemawat:
Clarify via a comment and a more appropriate method name as to why we defer initialization of metric id and labels.


mwhittaker (approved):
Thanks, Sanjay!

PR #450, opened 2023-07-11
Automatically sample traces in InstrumentHandler.

spetrovic77:
This makes for a good default behavior of some tracing always enabled "by default", i.e., when the recommended `InstrumentHandler` is used.

Users that wish to enable tracing manually can still do so.


http.go
```diff
@@ -98,6 +102,10 @@ func InstrumentHandler(label string, handler http.Handler) http.Handler {
 		}
 		httpRequestBytesReturned.Get(labels).Put(float64(writer.responseSize(r)))
 	})
+	const traceSampleRate = 0.01
```
ghemawat, line 105: I think we had discussed potentially doing time-based sampling. I.e., sample if X seconds have elapsed since the last sample. One benefit of that is that even low frequency services will get a decent set of samples while heavier ones will be sampled more aggressively.

Maybe once a second or once every five or ten seconds?

website/docs.md
```diff
@@ -1131,29 +1144,21 @@ func serve(ctx context.Context, app *app) error {
...
+tracing, once tracing is enabled for a given HTTP request, that request
+and the resulting component method calls will be automatically traced. Service
+Weaver will collect and export the traces for you. Refer to the
 deployer-specific documentation for [single process](#single-process-tracing),
 [multiprocess](#multiprocess-tracing), and [GKE](#gke-tracing) to learn about
-deployer specific exporters.
+deployer-specific exporters.
 
-The step above is all you need to get started with tracing. If you want to add
+The steps above is all you need to get started with tracing. If you want to add
```
ghemawat, line 1161: s/is/are/

ghemawat (commented):
Nice!

I also like how we now don't force users to use the otel API directly and let InstrumentHandler deal with things.

http.go
spetrovic77: Thanks, this is much better indeed. Set it to once every second for now.

spetrovic77:
> Nice!
> 
> I also like how we now don't force users to use the otel API directly and let InstrumentHandler deal with things.

Yes, thank you for all your suggestions.

http.go
```diff
@@ -106,6 +116,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
...
+	now := time.Now()
+	if now.Before(s.nextSampleTime) {
+		return false
+	}
+
+	// Pick a new value for nextSampleTime and return.
+	const jitter = 0.4
+	rv := 2*s.rng.Float64() - 1.0 // range [-1.0, 1.0)
+	mult := 1 + jitter*rv
+	s.nextSampleTime = s.nextSampleTime.Add(time.Duration(float64(s.interval) * mult))
```
ghemawat, line 154: I think this needs to be:

```
s.nextSampleTime = now.Add(...)
```

That way if we ever have a long idle period, we won't sample a lot of requests in the "past".

http.go
```diff
@@ -106,6 +116,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
...
+// shouldTrace returns true iff a request should be traced.
+func (s *traceSampler) shouldTrace() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	now := time.Now()
+	if now.Before(s.nextSampleTime) {
+		return false
+	}
+
+	// Pick a new value for nextSampleTime and return.
```
ghemawat, line 150: I suspect that unlike retry we can just randomly pick a duration in the range [0,2*period].

```
mult := rng.Float64(2*float64(s.interval))
```

(Retry was being clever by staying within [0.6,1.4] to ensure we don't retry too quickly, but here I think the simpler approach works)

http.go
```diff
@@ -106,6 +116,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
...
+		rng:            rand.New(rand.NewSource(now.UnixNano())),
+		nextSampleTime: now,
+	}
+}
+
+// shouldTrace returns true iff a request should be traced.
+func (s *traceSampler) shouldTrace() bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	now := time.Now()
```
ghemawat, line 145: Performance tweak -- may not matter:

We are potentially causing cache line contention here by locking the mutex on every request, even non-traced ones. An alternative would be to store the nextSampleTime in an atomic int and then change the initial check to not be with a lock:

```
   now := time.Now().UnixNano()
   if now < s.nextSampleTime.Load() {
      return
   }
```

Should check again after acquiring the mutex in case another goroutine has changed nextSampleTime. Not sure if this is worthwhile.

But at the very least, hoist the time.Now() call so it happens before the mutex is acquired.

http.go
```diff
@@ -106,6 +116,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
 	return InstrumentHandler(label, http.HandlerFunc(f))
 }
 
+// traceSampler is a time-based request sampler for tracing.
+//
+// It allows at most one request to be traced during each time interval.
+type traceSampler struct {
+	interval time.Duration
+	rng      *rand.Rand
```
ghemawat, line 124: I think rand.Rand is not thread-safe, so it should be guarded by a mutex and therefore it should be declared just below the mu field below.

http.go
spetrovic77: Good point, done!

http.go
spetrovic77: Nice, done.

http.go
spetrovic77: Good point, it is not thread-safe.

http.go
spetrovic77: Nice, done!

spetrovic77:
Thanks! PTAL.

http.go
```diff
@@ -106,6 +117,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
 	return InstrumentHandler(label, http.HandlerFunc(f))
 }
 
+// traceSampler is a time-based request sampler for tracing.
+//
+// It allows at most one request to be traced during each time interval.
+type traceSampler struct {
```
ghemawat, line 123: Perhaps not in this change, but we should add a test for traceSampler.

http.go
```diff
@@ -106,6 +117,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
...
+func newTraceSampler(interval time.Duration) *traceSampler {
+	return &traceSampler{
+		intervalNs: float64(interval / time.Nanosecond),
+		rng:        rand.New(rand.NewSource(time.Now().UnixNano())),
+	}
+}
+
+// shouldTrace returns true iff a request should be traced.
+func (s *traceSampler) shouldTrace() bool {
+	now := time.Now().UnixNano()
```
ghemawat, line 142: Add a comment:
  // Fast path check avoids acquiring a mutex or modifying the sampler in any way.

http.go
```diff
@@ -106,6 +117,45 @@ func InstrumentHandlerFunc(label string, f func(http.ResponseWriter, *http.Reque
...
+
+	// Check again while holding the lock.
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if now < s.nextSampleTimeNs.Load() {
+		return false
+	}
+
+	// Pick a new value for nextSampleTime and return.
+	s.nextSampleTimeNs.Store(now + int64(2*s.rng.Float64()*s.intervalNs))
```
ghemawat, line 155: Add comment:
   // Pick a random value in [0,s*s.intervalNs] (averages out to s.intervalNs).

ghemawat (approved):

http.go
spetrovic77: Ack.

PR #456, opened 2023-07-13
Support for returning custom errors from remote calls.

ghemawat:
Error types that implement WeaverMarshal/WeaverUnmarshal and are registered by calling codegen.RegisterSerializable[T] are now serialized properly when returned from remote calls instead of being emulated by their Error() strings.

A future change will change the code generator to automatically call RegisterSerializable for structs that embed weaver.AutoMarshal.

Also handle errors that wrap multiple errors by sending the entire tree instead of just the single error chain we used to send.

Compatibility note: the format for sending errors has changed.


runtime/codegen/automarshal.go
```diff
@@ -14,9 +14,77 @@
...
+// RegisterSerializable records type T as serializable. This is needed to
+// instantiate the appropriate concrete type when an interface is sent over the
+// wire (currently only used for AutoMarshal errors returned from remote method
+// calls). The registration is automatically done by generated code for custom
+// error structs that embed weaver.AutoMarshal.
+//
+// The extra PT type parameter is used to check that *T is an AutoMarshal
+// (since AutoMarshal methods have pointer receivers, but T itself may be a
+// struct).
+// func RegisterSerializable[T any, PT autoMarshalPointer[T]]() {
```
mwhittaker, line 54: Forgot to uncomment?

runtime/codegen/encoder.go
```diff
@@ -250,23 +250,86 @@ func (e *Encoder) Len(l int) {
 	e.Int32(int32(l))
 }
 
+//	Error encoding
```
mwhittaker, line 253: Spurious tab?

runtime/codegen/encoder.go
```diff
@@ -250,23 +250,86 @@ func (e *Encoder) Len(l int) {
...
+		case interface{ Unwrap() error }:
+			dfs(u.Unwrap())
+		case interface{ Unwrap() []error }:
+			for _, child := range u.Unwrap() {
+				dfs(child)
+			}
+		}
+	}
+	dfs(err)
+	e.Uint8(0)
```
mwhittaker, line 310: `endOfErrors` instead of `0`?

runtime/codegen/encoder_test.go
```diff
@@ -180,22 +180,41 @@ func TestErrorUnableToDecBytes(t *testing.T) {
...
-		{"flat", errors.New("hello")},
-		{"wrap1", fmt.Errorf("hello %w", os.ErrNotExist)},
-		{"wrap2", fmt.Errorf("hello %w", fmt.Errorf("world %w", os.ErrNotExist))},
-		{"custom", customTestError{"x"}},
-		{"wrap-custom", fmt.Errorf("hello %w", customTestError{"a"})},
+		{"nil", nil, nil, nil},
+		{"flat", errors.New("hello"), nil, []error{os.ErrNotExist}},
+		{"wrap1", fmt.Errorf("hello %w", os.ErrNotExist), []error{os.ErrNotExist}, nil},
+		{"wrap2", fmt.Errorf("hello %w", fmt.Errorf("world %w", os.ErrNotExist)), []error{os.ErrNotExist}, nil},
+		{"custom", customTestError{"x"}, nil, []error{os.ErrNotExist, alternateError{"x"}}},
```
mwhittaker, line 215: Replace `nil` with `[]error{customTestError{"x"}}`?

runtime/codegen/encoder.go
```diff
@@ -250,23 +250,86 @@ func (e *Encoder) Len(l int) {
...
+	// Convert the tree of wrapped errors into a single list.
+	// We do not use errors.Unwrap() since it does not visit the
+	// children found by "Unwrap()[]error".
+	//
+	// The encoding is a list of encoded errors terminated by a <0>.
+	// An encoded error is either <1,typKey,serialization> for
+	// registered custom error types, or <2,message,fmtError> for
+	// unregistered error types.
+	var dfs func(error)
+	dfs = func(err error) {
```
mwhittaker, line 281: Do you think it's eventually worth worrying about cyclic errors (e.g., https://go.dev/play/p/_6g1xLGBQ6f) that could make this code infinite loop? I think we currently disallow recursive types, but error "trees" can be cyclic. I guess even calling the `Error` method on an error could infinite loop, so maybe it's a moot point.

mwhittaker (approved):
Very nice! 

I noticed that right now, the tree structure of an encoded error is flattened into a list. In the future, do you plan to preserve the structure or it's not important? I don't think it affects the behavior of `errors.Is` or `errors.As`, so maybe it's not important?

ghemawat (commented):
W.r.t. flattening into a list: we could preserve the original structure, but that won't affect errors.Is/As behavior.

It will be more complicated though since we will need two decodedError types: one that implements "Unwrap() error" and one that implements "Unwrap() []error" to match the original behavior.

Do you think it is worthwhile?

mwhittaker:
> W.r.t. flattening into a list: we could preserve the original structure, but that won't affect errors.Is/As behavior.
> 
> It will be more complicated though since we will need two decodedError types: one that implements "Unwrap() error" and one that implements "Unwrap() []error" to match the original behavior.
> 
> Do you think it is worthwhile?

Maybe we can defer it for now :)

PR #457, opened 2023-07-13
Allow registered serializable types to be pointers.

ghemawat:
Remove type constraint from RegisterSerializable[T] and instead check at runtime that either T or *T implements AutoMarshal.

### DIFF
--- a/runtime/codegen/automarshal.go
+++ b/runtime/codegen/automarshal.go
@@ -27,12 +27,6 @@ type AutoMarshal interface {
 	WeaverUnmarshal(dec *Decoder)
 }
 
-// autoMarshalPointer[T] is an interface which asserts that *T implements AutoMarshal.
-type autoMarshalPointer[T any] interface {
-	*T
-	AutoMarshal
-}
-
 // Table of registered serialized types.
 var (
 	typesMu  sync.Mutex
@@ -45,11 +39,7 @@ var (
 // wire (currently only used for AutoMarshal errors returned from remote method
 // calls). The registration is automatically done by generated code for custom
 // error structs that embed weaver.AutoMarshal.
-//
-// The extra PT type parameter is used to check that *T is an AutoMarshal
-// (since AutoMarshal methods have pointer receivers, but T itself may be a
-// struct).
-func RegisterSerializable[T any, PT autoMarshalPointer[T]]() {
+func RegisterSerializable[T AutoMarshal]() {
 	var value T
 	t := reflect.TypeOf(value)
 	typesMu.Lock()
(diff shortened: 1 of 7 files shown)


runtime/codegen/automarshal.go
```diff
@@ -73,12 +76,25 @@ func RegisterSerializable[T any, PT autoMarshalPointer[T]]() {
...
+		pkg = t.Elem().PkgPath()
+	}
+
+	return fmt.Sprintf("%s(%s)", t.String(), pkg)
 }
 
 // canEncodeInterface returns true if the concrete type of value is registered
 // as a serializable type and can therefore be sent using Encoder.Interface.
 func canEncodeInterface(value any) bool {
+	if _, ok := value.(AutoMarshal); !ok {
```
mwhittaker, line 95: If I'm understanding correctly, if `value` is of type `T`, `T` implements `error`, but `*T` implements `AutoMarshal` instead of `T`, then we return false here? Is it possible for us to encode `value` by taking a pointer to it? 

I noticed that values of type `codegen.customTestError` return false here and end up being decoded as `codegen.decodedError`, but I would have expected them be decoded into a `codegen.customTestError`.

runtime/codegen/encoder.go
```diff
@@ -329,14 +329,21 @@ func (e *Encoder) Interface(value any) {
...
 
-	// We want a pointer to the value wrapped in any.
-	// XXX Support pointers to structs as well.
-	copy := reflect.New(t)
-	copy.Elem().Set(reflect.ValueOf(value))
-	am, ok := copy.Interface().(AutoMarshal)
+	// We allow either value, or a pointer to value to implement AutoMarshal.
+	am, ok := value.(AutoMarshal)
+	if !ok {
+		// Try using a pointer to value.
```
mwhittaker, line 335: Can we use the trick you used in `automarshal.go` here? 

```go
if !ok {
    value = &value
    am, ok = value.(AutoMarshal)
}
```

runtime/codegen/automarshal.go
```diff
@@ -46,11 +40,20 @@ var (
 // calls). The registration is automatically done by generated code for custom
 // error structs that embed weaver.AutoMarshal.
 //
-// The extra PT type parameter is used to check that *T is an AutoMarshal
-// (since AutoMarshal methods have pointer receivers, but T itself may be a
-// struct).
-func RegisterSerializable[T any, PT autoMarshalPointer[T]]() {
+// REQUIRES: T can be serialized (either T or a pointer to T must implement
```
mwhittaker, line 43: I believe right now, we always generate `AutoMarshal` methods on `*T`, and I don't think users are supposed to implement `AutoMarshal` methods manually. Does it simplify things if we always require `*T` to implement `AutoMarshal`?

mwhittaker (approved):
Awesome!

runtime/codegen/automarshal.go
ghemawat: That's a good point. Maybe I should abandon this change.

mwhittaker: Oh, maybe I misunderstood, but I thought most of the changes here were still needed to support errors where the `Error` method has a pointer receiver?

ghemawat: You are correct. I was wondering if we need to support Error() with a pointer receiver.

mwhittaker: Ah, I'm in favor of supporting it.

ghemawat: Ack. Just to clarify, yes the AutoMarshal methods are always defined on a pointer to a struct S.

The code here allows T to be either S or *S. This way, the user is free to place their Error() method on either S or *S.

runtime/codegen/automarshal.go
ghemawat: Good catch. Fixed.

runtime/codegen/encoder.go
ghemawat: That was a good idea. Until I realized that the "a = &value" trick was only working because value had the generic type T. Here, value has type any, and so "&value" does not have the type we want.

ghemawat:
PTAL

mwhittaker (approved):
LGTM!

PR #458, opened 2023-07-13
Preserve serializable custom errors from remote method calls.

ghemawat:
The code generator emits code to keep track of all error types that contain an embedded weaver.AutoMarshal in a global table.  The sender sends such errors by prefixing their serialization with they corresponding key in the global table. The receiver looks up the key in its global table, creates a value of the correct type, and deserializes the error contents into that value.

This allows custom errors to be returned from remote methods without losing type information.

### DIFF
--- a/website/docs.md
+++ b/website/docs.md
@@ -2755,10 +2755,17 @@ type Pair[A any] struct {
 To serialize generic structs, implement `BinaryMarshaler` and
 `BinaryUnmarshaler`.
 
-Finally note that while [Service Weaver requires every component method to
-return an `error`](#components-interfaces), `error` is not a
-serializable type. Service Weaver serializes `error`s in a way that does not
-preserve any custom `Is` or `As` methods.
+## Errors
+
+Service Weaver requires every component method to [return an
+error](#components-interfaces).  If a non-nil error is returned, Service Weaver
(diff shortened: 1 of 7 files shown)


website/docs.md
```diff
@@ -2755,10 +2755,17 @@ type Pair[A any] struct {
...
 
-Finally note that while [Service Weaver requires every component method to
-return an `error`](#components-interfaces), `error` is not a
-serializable type. Service Weaver serializes `error`s in a way that does not
-preserve any custom `Is` or `As` methods.
+## Errors
+
+Service Weaver requires every component method to [return an
+error](#components-interfaces).  If a non-nil error is returned, Service Weaver
+by default transmits the textual representation of the error and therefore any
```
spetrovic77, line 2762: Break this sentence into two, just before "and therefore" sounds good to me.

website/docs.md
```diff
@@ -2755,10 +2755,17 @@ type Pair[A any] struct {
...
+custom information stored in the error value, or custom `Is` or `As` methods are
```
spetrovic77, line 2763: comma after `methods`

spetrovic77 (approved):
Looks great!

ghemawat:
Thanks!

PR #459, opened 2023-07-14
Store trace spans in a relational DB format.

spetrovic77:
This format allows for traces to be queries. Here are the changes to the trace storage format:
  * Traces are stored in encoded proto format. This format is more efficient than the earlier Perfetto JSON format.
  * Some trace fields are also available as table columns, for fast query matching.
  * Traces are converted to the Perfetto JSON format during trace retrieval and display.

As discussed offline with sanjay@, traces are no longer displayed in Perfetto in bulk. Instead, each trace has a separate link and the Perfetto UI always displays a single trace. This makes it easier to visualize individual trace executions, as they are no longer interleaved on the same timeline.

Other changes:
  * Fold Span proto messages into the `Span` type, as their names are quite generic (e.g., `Attribute`).
  * Pass trace data between weavelet and envelope in Proto format, since the Proto format can be efficiently serialized, if needed. (The OTel format doesn't support efficient serialization.)
  * Minor code re-organization.

[Service Weaver Tracing.webm](https://github.com/ServiceWeaver/weaver/assets/5449850/95493490-fe2b-45b8-89b0-27ec327cecf7)


runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		if err := rows.Scan(&traceID, &startTimeUs, &endTimeUs); err != nil {
+			return nil, err
+		}
+		traces = append(traces, TraceSummary{
+			TraceID:   traceID,
+			StartTime: time.UnixMicro(startTimeUs),
+			EndTime:   time.UnixMicro(endTimeUs),
+		})
+	}
+	return traces, nil
```
ghemawat, line 255: Shouldn't we check for rows.Err() at the end?

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		if err := rows.Scan(&encoded); err != nil {
+			return nil, err
+		}
+		span := &protos.Span{}
+		if err := proto.Unmarshal(encoded, span); err != nil {
+			return nil, err
+		}
+		spans = append(spans, span)
+	}
+	return spans, nil
```
ghemawat, line 280: Shouldn't we check for rows.Err() at the end?

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+// (e.g., "127.0.0.1"), and <trace_id> is the trace ID in hexadecimal format.
+//
+// Perfetto UI requires that the server runs on the local port 9001. For that
+// reason, this method will block until port 9001 becomes available.
+func (d *DB) Serve(ctx context.Context) {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/status", func(w http.ResponseWriter, _ *http.Request) {
+		w.Write([]byte("OK")) //nolint:errcheck // response write error
+	})
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
```
ghemawat, line 322: This handler is pretty big. Consider extracting it out into a top-level DB method:
    mux.HandleFunc("/", d.fetchTrace)

ghemawat, line 322: Is "/" the wrong path for this server? Seems like "/trace" would be better and would allow us to put other functionality on this server later.

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		//   https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui.
+		w.Header().Set("Access-Control-Allow-Origin", "https://ui.perfetto.dev")
+		traceID := r.URL.Query().Get("trace_id")
+		if traceID == "" {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		spans, err := d.fetchSpans(r.Context(), traceID)
+		if err != nil || len(spans) == 0 {
+			w.WriteHeader(http.StatusNotFound)
```
ghemawat, line 333: We are swallowing errors here silently. Perhaps if err is not nil, log it and then return an "internal server error".

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+	mux.HandleFunc("/status", func(w http.ResponseWriter, _ *http.Request) {
+		w.Write([]byte("OK")) //nolint:errcheck // response write error
+	})
+	mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		// Set access control according to:
+		//   https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui.
+		w.Header().Set("Access-Control-Allow-Origin", "https://ui.perfetto.dev")
+		traceID := r.URL.Query().Get("trace_id")
+		if traceID == "" {
+			w.WriteHeader(http.StatusNotFound)
```
ghemawat, line 328: This one is probably a StatusBadRequest

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		traces, err := encodeSpans(spans)
+		if err != nil {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		w.Write(traces) //nolint:errcheck // response write error
+	})
+	server := http.Server{Handler: mux}
+	ticker := time.NewTicker(time.Second)
+	var warnOnce sync.Once
```
ghemawat, line 345: Add a comment explaining what this loop is doing.

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+	for {
+		select {
+		case <-ticker.C:
+			lis, err := net.Listen("tcp", "localhost:9001")
+			if err != nil {
+				warnOnce.Do(func() {
+					fmt.Fprintf(os.Stderr, "Perfetto server failed to listen on port 9001: %v\n", err)
```
ghemawat, line 352: Why do we need to hard-wire 9001?

I am a little lost as to who generates URLs pointing at this server. Could we let the system pick a port but communicate the picked port to the place that generates URLs?

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+			return
+		}
+	}
+}
+
+// encodeSpans encodes the given spans in a format that can be read by the
+// Perfetto UI.
+func encodeSpans(spans []*protos.Span) ([]byte, error) {
+	var buf bytes.Buffer
+	buf.Write([]byte{'['})
```
ghemawat, line 375: Add comment:
   // We are returning a JSON array, so surround everyting in [].

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+// encodeSpans encodes the given spans in a format that can be read by the
+// Perfetto UI.
+func encodeSpans(spans []*protos.Span) ([]byte, error) {
+	var buf bytes.Buffer
```
ghemawat, line 374: Consider using strings.Builder. You can then do buf.WriteByte() and avoid unnecesary []byte{...} boilerplate below.

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		if buf.Len() > 1 { // NOTE: buf always starts with a '['
+			buf.Write([]byte{','})
+		}
+		_, err := buf.Write(event)
+		return err
+	}
 	for _, span := range spans {
-		if err := d.encodeSpan(ctx, app, version, span, &b); err != nil {
+		s := traceio.ReadSpan{Span: span}
+		events, err := encodeSpanEvents(&s)
```
ghemawat, line 385: Any reason to pass &s instead of s?

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		}
+	}
+}
+
+// encodeSpans encodes the given spans in a format that can be read by the
+// Perfetto UI.
+func encodeSpans(spans []*protos.Span) ([]byte, error) {
+	var buf bytes.Buffer
+	buf.Write([]byte{'['})
+	append := func(event []byte) error {
```
ghemawat, line 376: Can you pick a name different than append to avoid confusion with the builtin append.

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+		if buf.Len() > 1 { // NOTE: buf always starts with a '['
+			buf.Write([]byte{','})
```
ghemawat, line 378: You are checking errors on some writes and not on others. From what I can tell, these buffered destinations don't fail. So perhaps we can just ignore errors on Write calls?

ghemawat (commented):
This is great!

internal/envelope/conn/trace_readwrite_test.go
```diff
@@ -77,96 +76,96 @@ func (*pipeForTest) VerifyServerCertificate(context.Context, *protos.VerifyServe
...
-	return pipe.spans[0], nil
+	return pipe.spans.Span[0], nil
 }
 
-func testAttrs() []*protos.Attribute {
-	num := func(t protos.Attribute_Value_Type, v uint64) *protos.Attribute_Value {
-		return &protos.Attribute_Value{
+func testAttrs() []*protos.Span_Attribute {
+	num := func(t protos.Span_Attribute_Value_Type, v uint64) *protos.Span_Attribute_Value {
+		return &protos.Span_Attribute_Value{
```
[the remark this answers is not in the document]
spetrovic77: It wasn't so bad, just a simple rename.

internal/status/dashboard.go
```diff
@@ -311,3 +319,46 @@ func (d *dashboard) handleMetrics(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	// TODO(mwhittaker): Change to /<deployment id>/traces?
+	id := r.URL.Query().Get("id")
+	if id == "" {
+		http.Error(w, "no deployment id provided", http.StatusBadRequest)
+		return
+	}
+
+	// Weavelets export traces every 5 seconds. In order to (semi-)guarantee
```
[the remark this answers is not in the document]
spetrovic77: Good point, done.

runtime/perfetto/db.go
spetrovic77: Nice. Done.

runtime/perfetto/db.go
spetrovic77: Thanks for checking. Done.

internal/status/dashboard.go
```diff
@@ -311,3 +319,46 @@ func (d *dashboard) handleMetrics(w http.ResponseWriter, r *http.Request) {
...
+	// that the database contains all spans for the selected traces, we put
+	// an upper time limit on matching traces.
```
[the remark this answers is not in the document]
spetrovic77: Thanks, much clearer.

spetrovic77:
> This is great!

Thank you for suggesting this change!

runtime/perfetto/db.go
```diff
@@ -65,33 +70,18 @@ type metadataEvent struct {
...
-// [2] https://ui.perfetto.dev/
+// TODO(spetrovic): Separate the database from the Pefetto serving code.
 type DB struct {
-	// Trace data is stored in a sqlite DB spread across three tables:
-	// (1) traces:           trace data in a Perfetto-UI-compattible JSON format
-	// (2) replica_num:      map from colocation group replica id to a replica
-	//                       number
-	// (3) next_replica_num: the next replica number to use for a given
-	//                       colocation group
+	// Trace data is stored in a sqlite DB spread across two tables:
```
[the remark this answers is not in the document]
spetrovic77: In Otel, all spans in a given trace form a tree. Every span is identified via a `(trace_id, span_id)`, but it also has a `parent_span_id`, which establishes the tree structure. The root span has a `parent_span_id` of `0`. There's nothing special about it other than the value for that id.

Re: end timestamps. The root span encompasses all of the other spans, and we can just use its (start,end) time as the trace (start,end) interval.

Reading your comments below, I think there's a query performance benefit for separating traces (i.e., root spans) from the rest. That's a good point. The downside is simplicity of storing all spans in a single table. This simplicity might come useful when we start garbage-collecting traces. Although to be honest, I'm not sure which side would win w.r.t. garbage-collection.

We can leave it as-is for now and decide once the GC change is out?

runtime/perfetto/db.go
```diff
@@ -65,33 +70,18 @@ type metadataEvent struct {
...
+	// (1) spans:           serialized trace spans
+	// (2) span_attributes: span attributes stored in key,value format
```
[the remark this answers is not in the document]
spetrovic77: Good point about encoded_span. I haven't really thought about how table size impacts querying performance.

Btw., attributes need to be separated so that they can be queried (a TODO for now), as they are repeated.

See my earlier comment. I can definitely move encoded_spans into a separate table.

runtime/perfetto/db.go
```diff
@@ -112,34 +102,32 @@ func Open(ctx context.Context, fname string) (*DB, error) {
...
+		fname: fname,
+		db:    db,
 	}
 
 	const initTables = `
--- Trace data.
-CREATE TABLE IF NOT EXISTS traces (
+-- Trace span data.
+CREATE TABLE IF NOT EXISTS spans (
+	id INTEGER PRIMARY KEY AUTOINCREMENT,
```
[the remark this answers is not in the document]
spetrovic77: Good point, done.

runtime/perfetto/db.go
```diff
@@ -112,34 +102,32 @@ func Open(ctx context.Context, fname string) (*DB, error) {
...
--- Map from a weavelet id to a replica number.
-CREATE TABLE IF NOT EXISTS replica_num (
-	app TEXT NOT NULL,
-	version TEXT NOT NULL,
-	weavelet_id TEXT NOT NULL,
-	num INTEGER NOT NULL,
-	PRIMARY KEY(app,version,weavelet_id)
+-- Trace span attributes.
+CREATE TABLE IF NOT EXISTS span_attributes (
+	id INTEGER NOT NULL,
```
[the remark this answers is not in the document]
spetrovic77: In my upcoming change, I'd like to use attributes while querying. In particular, I'd like the HTTP InstrumentHandler to add attributes that mark traces with errors, so that we can query them.

If used by querying, I need the attributes table to be keyed to the traces table.

runtime/perfetto/db.go
```diff
@@ -112,34 +102,32 @@ func Open(ctx context.Context, fname string) (*DB, error) {
...
+		db:    db,
 	}
 
 	const initTables = `
--- Trace data.
-CREATE TABLE IF NOT EXISTS traces (
+-- Trace span data.
+CREATE TABLE IF NOT EXISTS spans (
+	id INTEGER PRIMARY KEY AUTOINCREMENT,
 	app TEXT NOT NULL,
```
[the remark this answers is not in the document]
spetrovic77: That's a good point in favor of a single querying table. I can make this change in a subsequent PR?

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+	res, err := d.execDB(ctx, spanStmt, app, version, span.Name, hex.EncodeToString(span.TraceId), hex.EncodeToString(span.SpanId), hex.EncodeToString(span.ParentSpanId), span.StartMicros, span.EndMicros, encoded)
+	if err != nil {
+		return fmt.Errorf("write span to %s: %w", d.fname, err)
+	}
+	id, err := res.LastInsertId()
+	if err != nil {
+		return fmt.Errorf("get last insertion id to %s: %w", d.fname, err)
+	}
+
+	// NOTE: we perform attribute inserts non-transactionaly, for performance
```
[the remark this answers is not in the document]
spetrovic77: Same reason as above: I was worried about concurrency issues with bulkier writes.

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
...
+	const query = `
+SELECT trace_id, start_time_unix_us, end_time_unix_us
+FROM spans
+WHERE
+	(parent_span_id="0000000000000000") AND
+	(app=? OR ?="") AND (version=? OR ?="") AND
+	(start_time_unix_us>=? OR ?=0) AND (end_time_unix_us<=? OR ?=0) AND
+	((end_time_unix_us - start_time_unix_us)>=? OR ?=0) AND
+	((end_time_unix_us - start_time_unix_us)<=? OR ?=0)
+GROUP BY trace_id
```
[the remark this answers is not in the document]
spetrovic77: Oh, this was a spurious "GROUP BY". We're selecting only the root spans, which have the parent_span_id set to 0, and those are unique. (See earlier comment).

I removed the stanza.

runtime/perfetto/db.go
spetrovic77: Good point, my error checking code was wrong. Done.

runtime/perfetto/db.go
spetrovic77: Changed to using http.Error(), which allows for error string to be included.

runtime/perfetto/db.go
spetrovic77: This is unfortunately a Perfetto requirement. I'm not fully sure why, Robert added this support originally. But he tried with other ports and it didn't work.

runtime/perfetto/db.go
spetrovic77: I'm not 100% sure TBH, Robert added this change and he was forced to choose certain URLs for Chrome to load them correctly.

This whole `https://ui.perfetto.dev` URL handling is a bit of a magic and something I didn't dive into too deeply.

spetrovic77:
PTAL

runtime/perfetto/db.go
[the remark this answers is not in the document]
ghemawat: Separate PR sounds great

runtime/perfetto/db.go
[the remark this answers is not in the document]
ghemawat: Ack

runtime/perfetto/db.go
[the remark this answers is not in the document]
ghemawat: Ack

runtime/perfetto/db.go
```diff
@@ -155,47 +143,269 @@ func (d *DB) Close() error {
 }
 
 // Store stores the given traces in the database.
-func (d *DB) Store(ctx context.Context, app, version string, spans []sdktrace.ReadOnlySpan) error {
-	encoded, err := d.encodeSpans(ctx, app, version, spans)
+func (d *DB) Store(ctx context.Context, app, version string, spans *protos.TraceSpans) error {
+	var errs []error
+	for _, span := range spans.Span {
```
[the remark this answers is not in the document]
ghemawat: I don't have great insights here either. Perhaps we can measure it later and then change if appropriate.

internal/status/templates/traces.html
```diff
@@ -0,0 +1,68 @@
...
+              <th scope="col">Start Time</th>
+              <th scope="col">Duration</th>
+            </tr>
+          </thead>
+          <tbody>
+            {{range .Traces}}
+              <tr>
+                <td><a href="{{traceurl .TraceID}}">link</a></td>
+                <td>{{.StartTime}}</a></td>
+                <td>{{dur .StartTime .EndTime}}</a></td>
```
ghemawat, line 59: What is "dur"?  Would it be simpler to add a "Latency" field to the struct that is being rendered and just print that here?

ghemawat (approved):

internal/status/templates/traces.html
spetrovic77: This is displaying the repeated version of an existing struct (i.e., `TraceSummary`), and I didn't want to invent another struct and convert into it.

I renamed "dur" -> "sub" to be more accurate.

PR #461, opened 2023-07-18
Moved most of weaver code to internal package.

mwhittaker:
This PR moves the majority of the code in the `weaver` package into a newly created `internal/weaver` package.

Recall that weavertests run full Service Weaver applications with a couple of small tweaks. For example, a weavertest can provide a set of fake component implementations that replace the real implementations.

Before this PR, there was a bit of awkwardness between the weaver package and the weavertest package. Because the main weaver package is public, it cannot export any functions that might be useful to the weavertest package. To solve this, we did a couple of hacks. First, we pass some arguments to a weavelet via a context. Second, we had a `private` package that acted as an intermediary. To avoid cyclic dependencies, this `private` package used `any` liberally and relied on the callers of the package to cast to the appropriate types.

This PR is the first step in resolving the awkwardness. By moving most of the weaver code into an internal package, we can export functions useful to weavertest. This PR deletes the `private` package entirely.

In future PRs, I think we can continue to clean up the APIs. This PR tries to make as few changes as possible.


internal/weaver/weavelet.go
```diff
@@ -54,10 +53,11 @@ var readyMethodKey = call.MakeMethodKey("", "ready")
 
 // A weavelet runs and manages components. As the name suggests, a weavelet is
 // analogous to a kubelet.
-type weavelet struct {
-	ctx       context.Context
-	env       env                  // Manages interactions with execution environment
-	info      *protos.EnvelopeInfo // Setup info sent by the deployer.
+type Weavelet struct {
+	ctx context.Context
+	env env // Manages interactions with execution environment
```
spetrovic77, line 58: TODO to delete env?

internal/weaver/types.go
```diff
@@ -0,0 +1,31 @@
...
+var (
+	// IsImplements, IsRef, and IsListener return whether the provided type is
+	// a weaver.Implements[T], weaver.Ref[T], and weaver.Listener respectively.
+	// These values are populated by an init function in the main weaver
+	// package. The circuitousness is to avoid cyclic dependencies.
+	//
+	// TODO(mwhittaker): Improve these abstractions to initialize values as
+	// well. The code in fillrefs, filllisteners, etc. relies on some unsafe
+	// reflection stuff.
+	IsImplements func(reflect.Type) bool
```
spetrovic77, line 28: You must have thought of it, but would it work if you defined `Implements` (and other structs) in this file, and weaver.go had:

```
type Implements[T any] weaver_internal.Implements[T]
type Ref[T any] weaver_internal.Ref[T]
...
```

weaver.go
```diff
@@ -131,35 +149,355 @@ func Run[T any, _ PointerToMain[T]](ctx context.Context, app func(context.Contex
...
+// package. It exists so that a component struct that embeds Implements[T]
+// implements the InstanceOf[T] interface.
+//
+//nolint:unused
+func (Implements[T]) implements(T) {}
+
+// isImplements is an internal method that is only implemented by Implements[T]
+// and is used internally to check that a value is of type Implements[T].
+//
+//nolint:unused
```
spetrovic77, line 227: The GitHub Actions errors is fantastic!! :smile: 

`directive `//nolint:unused` is unused for linter "unused" (nolintlint)`

spetrovic77 (approved):
Very clean, I expected more churn.

internal/weaver/types.go
ghemawat: I like having the type directly in weaver.go. I suspect it is better in godoc for users of the system.

ghemawat (commented):
Nice cleanups!

internal/weaver/types.go
mwhittaker: Yeah, I had tried this but also preferred the godoc being simpler. I'll probably think harder about it because there might be a way to have nice godoc and simplify things.

internal/weaver/weavelet.go
mwhittaker: Yup, I already deleted it in a PR I'm about to send out in a second.

weaver.go
mwhittaker: :laughing: should be fixed now!

mwhittaker (commented):
Thanks Srdjan and Sanjay!

internal/weaver/types.go
mwhittaker: I looked into this more. I was able to replace `IsImplements` with something simpler, but `IsRef` and `IsListener`&mdash;and more generally, the way we initialize `weaver.Ref`s and `weaver.Listener`s&mdash;are hard to simplify even if all the code were in the same package. When we go to initialize a `weaver.Listener` field in a component struct, for example, the field may not be exported. Because of this, we're forced to use the icky unsafe parts of the reflect package. I'll open a PR with some slight simplifications though.

PR #465, opened 2023-07-18
Add a time-based trigger to delete trace entries older than 30 days.

spetrovic77:
This will help keep the size of the trace table in check.

NOTE: an earlier iteration of this change used to accept the time horizon value (i.e., 30 days) as an option. However, since the database can be shared across multiple clients with potentially different options, there is no way to guarantee that the time horizon value will be respected across the clients. For this reason, we hard-code a reasonable value for the time horizon of 30 days, at least for now.

Performance neutral (at least when no deletions are happening).

Before:
BenchmarkStore/1-72  	     100	  10824509 ns/op
BenchmarkStore/10-72 	     100	  11441102 ns/op
BenchmarkStore/100-72         	      91	  13537496 ns/op

After:
BenchmarkStore/1-72  	     100	  10735265 ns/op
BenchmarkStore/10-72 	     100	  11530419 ns/op
BenchmarkStore/100-72         	      93	  14008470 ns/op

Other changes:
  * Delete the attributes table as it is currently unused.

### DIFF
--- a/runtime/perfetto/db.go
+++ b/runtime/perfetto/db.go
@@ -73,14 +73,12 @@ type metadataEvent struct {
 
 // DB is a trace database that stores traces on the local file system.
 //
-// Trace spans are stored in a single sqlite DB table. Each span is stored in
-// a separate row in the table.
-//
 // TODO(spetrovic): Separate the database from the Pefetto serving code.
 type DB struct {
 	// Trace data is stored in a sqlite DB spread across two tables:
-	// (1) spans:           serialized trace spans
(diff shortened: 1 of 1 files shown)


ghemawat (approved):
Nice!

spetrovic77:
Note: I had to implement a two-trigger solution as suggested by @ghemawat offline. `PRAGMA foreign_keys=ON` doesn't work as spans are collected asynchronously, so a span may be inserted before its root span (i.e., its trace), which fails the foreign key enforcement.

PR #466, opened 2023-07-18
Split the weavelet in two.

mwhittaker:
Before this PR, we had a weavelet that was used by all the deployers. This made the weavelet code a bit hard to follow. I would say 75% of the weavelet code only ran in multiprocess mode (e.g., setting up internal listeners, updating routing info, setting up mTLS), but some bits of code only ran in single process mode. The control flow of this code was also dictated by a combination of a `Bootstrap` struct passed via a `context.Context`, some environment variables, and the `Env` interface (which was only partially implemented in the single process case).

This PR splits the weavelet in two. First, I introduce a new `Weavelet` interface that borrows from the old `private.App` interface:

```go
type Weavelet interface {
    GetIntf(t reflect.Type) (any, error)
    GetImpl(t reflect.Type) (any, error)
}
```

Conceptually, a `Weavelet` is something that hosts components, and it provides an API to get a handle to these components.

Next, I introduce a `SingleWeavelet` implementation that is used by `go run .` and `weaver single deploy` and a `RemoteWeavelet` implementation that is used by all other deployers.


ghemawat (approved):
Really nice cleanups. Thanks.

Issue #468, opened 2023-07-19
Issues with Logging Functionality and Incomplete Logs in Weaver Multi Deployment


ryan961:
weaver version:  weaver v0.17.0

1. The Logging functionality within Weaver does not support predefined configurations (such as setting the default `Handler` to `NewJSONHandler`), which poses a great inconvenience for me as I intend to independently deploy a single process using the built-in `s.Logger()` and collect logs.
2. When using the built-in Logger and deploying with Weaver Multi, the logs collected through the command `weaver multi logs --format=json` are incomplete, as the fields I added using `s.Logger().With(slog.String("logType", "attribute"))` are missing.
Default output when deploying with `weaver multi deploy weaver.toml`:
![image](https://github.com/ServiceWeaver/weaver/assets/32825698/6317716a-e6e5-4279-bc7e-457a8dcf7f45)
What I see when using `weaver multi logs --format=json`:
![image](https://github.com/ServiceWeaver/weaver/assets/32825698/9e02b6fd-e3ca-47a0-87c5-958bd4b989a1)

mwhittaker:
I opened PR #470 to address your second point. Great catch! 

Do you mind elaborating on your first point? I'm not sure I fully understand. Are you saying that calling the `Logger` method on a component implementation should return the logger set by `slog.SetDefault`?

ryan961:
Thank you for your timely reply and the PR fix, @mwhittaker.

Indeed, it is somewhat similar to the functionality of `slog.SetDefault`, which allows me to customize the `slog.Logger` in `weaver.Main`. For example, setting a custom `io.Writer` (such as `os.File`...), using `slog.NewJSONHandler` instead of `slog.NewTextHandler` by default, and utilizing `slog.ReplaceAttr()` for field manipulation or even data obfuscation...

I noticed that the documentation only briefly explains the logging for single processes (https://serviceweaver.dev/docs.html#single-process-logging), and upon inspecting the source code (v0.17.0), it seems that the logger is hardcoded as `slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelError + 1}))`(https://github.com/ServiceWeaver/weaver/blob/b181a14717694b15882d7fc23af405292a07118a/weavelet.go#L130). Although `weaver multi logs ` can support most of the mentioned requirements, it is not very convenient for single-process deployments and development debugging stages.

[...]

PR #470, opened 2023-07-19
Added missing attributes to json logs.

mwhittaker:
@ryan961 pointed out in #468 that the json formatted logs produced by `weaver multi logs` are missing attributes. This PR fixes that bug and adds them.


spetrovic77 (approved):
Nice catch!

PR #472, opened 2023-07-19
Rename the runtime perfetto/ package into traces/.

spetrovic77:
Some deployers may wish to use the trace database without relying on the Perfetto UI.

Other changes:
  * Move the ReadSpan into the traces/ package, as we now send the proto across the wire, and some deployers (e.g,. GKE, Kube) will want to convert back to the OTel ReadOnlySpan format.
  * Move the perfetto UI code into a separate file.
  * Minor cleanups.


rgrandl (approved):
LGTM

PR #473, opened 2023-07-19
Slightly simplified how components are filled.

mwhittaker:
This PR replaces weaver.IsImplements, weaver.IsRef, and weaver.IsListener with weaver.SetLogger, weaver.FillRefs, and weaver.FillListeners.

weaver.SetLogger is able to set the logger of a weaver.Implements without having to use reflection. For this to work, the function must be in the root weaver package.

weaver.FillListeners is a slightly simplified version of the existing FillListeners function. It checks if a field is a Listener by comparing to the Listener type directly. It also sets the listener and proxy address by casting directly to a *Listener. This function doesn't have to be in the root weaver package, but it can't be in the internal weaver package because of a cyclic dependency.

weaver.FillRefs is pretty much unmodified from the existing FillRefs function, but I moved it into the root weaver package to sit next to SetLogger and FillListeners.


fill.go
```diff
@@ -0,0 +1,133 @@
...
+		f := s.Field(i)
+		t := s.Type().Field(i)
+		if f.Type() != reflection.Type[Listener]() {
+			continue
+		}
+
+		// The listener's name is the field name, unless a tag is present.
+		name := t.Name
+		if tag := t.Tag.Get("weaver"); tag != "" {
+			if !token.IsIdentifier(tag) {
```
ghemawat, line 105: Is token too big of a dependency to pull in just to be able to use IsIdentifier? Do we have an identifier checker somewhere else we can call instead?

You are just moving code for now, so no reason to change things in this PR, but I spotted the dep now and wanted to note the issue.

ghemawat (approved):

fill.go
mwhittaker: Good point. Let me replace it in another PR. The implementation is dirt simple, so I'll just copy it over: https://cs.opensource.google/go/go/+/refs/tags/go1.20.6:src/go/token/token.go;l=331-341

PR #474, opened 2023-07-19
Removed dependency on token package.

mwhittaker:
This PR replaces token.IsIdentifier with a custom implementation to remove the dependency on the token package.

Note that this PR now allows listener names to be go keywords, but I think that's okay. I don't think we intended to disallow these names in the first place.

### DIFF
--- a/fill.go
+++ b/fill.go
@@ -102,7 +102,7 @@ func fillListeners(impl any, get func(name string) (net.Listener, string, error)
 		// The listener's name is the field name, unless a tag is present.
 		name := t.Name
 		if tag := t.Tag.Get("weaver"); tag != "" {
-			if !isIdentifier(tag) {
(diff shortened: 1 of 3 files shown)


fill.go
```diff
@@ -123,6 +123,22 @@ func fillListeners(impl any, get func(name string) (net.Listener, string, error)
 	return nil
 }
 
+// isIdentifier returns whether the provided string is a valid Go identifier,
```
ghemawat, line 126: Do we want this to be a Go identifier, or a valid listener name? If it is the latter, the comment is misleading.

mwhittaker: The latter. I renamed things and added some clarifying comments.

ghemawat (approved):
Looks great. Thanks!

PR #478, opened 2023-07-20
Display Perfetto UI from JavaScript.

spetrovic77:
This fixes the issue where the Perfetto server has to run on port 9001 and serve traces on the "/" path.

Now, the dashboard serves traces on its local server. This simplifies trace serving and allows multiple
dashboard commands to serve traces independently.

Other changes:
  * Moved the perfetto code into a separate runtime package.

[cast2.webm](https://github.com/ServiceWeaver/weaver/assets/5449850/ced37aae-83e3-4054-ba65-c949a3f37e0c)

### DIFF
--- a/internal/status/templates/traces.html
+++ b/internal/status/templates/traces.html
@@ -38,46 +38,83 @@
   <header class="navbar">
     <a href="/">{{.Tool}} dashboard</a>
   </header>
-
+  <script type="text/javascript">
+    // The code below largely taken from:
+    //   https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui
+    const ORIGIN = 'https://ui.perfetto.dev';
+    
+    async function fetchAndOpen(traceUrl) {
+      const resp = await fetch(traceUrl);
+      const blob = await resp.blob();
+      const arrayBuffer = await blob.arrayBuffer();
+      openTrace(arrayBuffer, traceUrl);
+    }
+    
+    function openTrace(arrayBuffer, traceId, traceUrl) {
+      const win = window.open(ORIGIN);
+      if (!win) {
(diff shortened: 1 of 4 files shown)


internal/status/templates/traces.html
```diff
@@ -38,46 +38,83 @@
...
+      const arrayBuffer = await blob.arrayBuffer();
+      openTrace(arrayBuffer, traceUrl);
+    }
+    
+    function openTrace(arrayBuffer, traceId, traceUrl) {
+      const win = window.open(ORIGIN);
+      if (!win) {
+        alert('Popups blocked. Please allow popups in order to be able to' +
+              'see traces');
+      }
```
ghemawat, line 58: Should there be a return after the alert() call?

internal/status/templates/traces.html
```diff
@@ -38,46 +38,83 @@
...
+      const timer = setInterval(() => win.postMessage('PING', ORIGIN), 50);
```
ghemawat, line 59: 50ms seems a bit aggressive? I wonder if we can reliably expect a response in 50ms. I think a few seconds may be fine? If the server responds quickly, we should be able to continue immediately?

ghemawat (approved):
Nice simplifications!

internal/status/templates/traces.html
spetrovic77: Unfortunately, a larger delay causes the trace display to be delayed by the passed-in duration (e.g., 5 seconds).

I think what's happening is that `postMessage()` fails the first time while the window is opening, and we end up waiting for the full interval before retrying.

There's no harm in pinging the new window at a higher rate: we stop the pinging and ignore replies when we receive the first response.

I've left it at 50ms for now.

PR #481, opened 2023-07-24
Generate reflection-based stubs.

mwhittaker:
## Background

Recall that `weaver generate` generates two types of stubs for every component: local stubs and client stubs.

A **local stub** for a component with interface type `T` has type `T`. A local stub is constructed with an instance of the struct that implements component `T`. Method calls on the local stub are dispatched directly to this instance. Here's a simplified example for the `Even` component from collatz:

```go
type even_local_stub struct {
    impl Even
}

var _ Even = (*even_local_stub)(nil)

func (s even_local_stub) Do(ctx context.Context, a0 int) (r0 int, err error) {
    // Do some metrics and tracing stuff...
    return s.impl.Do(ctx, a0)
}
```

A **client stub** for component `T` also has type `T`, but executes method calls on `T` as remote procedure calls, serializing arguments and return values. Again, here's a simplified example for the `Even` component from collatz:

```go
type even_client_stub struct {
    stub codegen.Stub
}

var _ Even = (*even_client_stub)(nil)

func (s even_client_stub) Do(ctx context.Context, a0 int) (r0 int, err error) {
    // Do some metrics and tracing stuff...

    // Encode arguments.
    enc := codegen.NewEncoder()
    enc.Int(a0)
    var shardKey uint64

    // Call the remote method.
    requestBytes = len(enc.Data())
    var results []byte
    results, err = s.stub.Run(ctx, 0, enc.Data(), shardKey)
    replyBytes = len(results)
    if err != nil {
        err = errors.Join(weaver.RemoteCallError, err)
        return
    }

    // Decode the results.
    dec := codegen.NewDecoder(results)
    r0 = dec.Int()
    err = dec.Error()
    return
}
```

## Reflect Stubs

This PR introduces a new reflection-based stub called a **reflect stub**. A reflect stub for component `T` executes method calls on `T` by dispatching to generic reflection-based caller function that receives the method call's arguments as `reflect.Value`s and returns its results as `reflect.Value`s. Again, here's `Even`:

```go
type even_reflect_stub struct {
    caller func(reflect.Type, string, []reflect.Value) []reflect.Value
}

var _ Even = (*even_reflect_stub)(nil)

func (s even_reflect_stub) Do(ctx context.Context, a0 int) (r0 int, err error) {
    component := reflect.TypeOf((*Even)(nil)).Elem()
    args := make([]reflect.Value, 2)
    args[0] = reflect.ValueOf(ctx)
    args[1] = reflect.ValueOf(a0)
    results := s.caller(component, "Do", args)
    r0 = results[0].Interface().(int)
    if x := results[1].Interface(); x != nil {
            err = x.(error)
    }
    return
}
```

## Motivation

I introduced reflect stubs to implement [deterministic simulation][testing_doc] of Service Weaver apps. A simulator has to intercept every method call, pausing its execution until a later time. Reflect stubs make this possible.

The team had discussed offline the possibility of using client stubs to implement a simulator rather than introducing a whole new kind of stub. The idea was to inject a fake RPC client (i.e. a `codegen.Stub`) into the client stub that allowed the simulator to interpose.

I played with this idea for a bit, but ran into a challenge. The method call arguments provided to the fake RPC client are already encoded into a `[]byte`, meaning that the simulator cannot see what they are. When reporting an execution history to the user, the simulator needs to include the values of these arguments in order for the user to understand what happened. I couldn't think of a way to get around this without using reflection.

[testing_doc]: https://github.com/ServiceWeaver/weaver/blob/3cb82f53ee959c6c7bbcef50d6b202c3e1d7394f/docs/randomized_testing.md


runtime/version/version.go
```diff
@@ -28,7 +28,7 @@ const (
 	//
 	// [1]: https://go.dev/doc/modules/version-numbers
 	ModuleMajor = 0
-	ModuleMinor = 17
+	ModuleMinor = 18
```
spetrovic77, line 31: I thought we're bumping the module version only once every two weeks. Or is that only for creating a new release?

I'm guessing that codegen changes require immediate module version bumps. Is that so? And protocol changes don't require immediate module version bumps?

It would be better if we would stick with bumping the module version only on a schedule.

mwhittaker:
We discussed offline measuring the binary size of our example applications with and without reflect stubs. Here are the results. The change in binary size is insignificant. 

| App | Without Stubs | With Stubs | Δ |
| - | - | - | - | 
| chat | 27896 | 27932 | +36 |
| collatz | 26900 | 26912 | +12 |
| factors | 26940 | 26944 | +4 |
| hello | 26872 | 26876 | +4 |
| helloworld | 26464 | 26460 | -4 | 
| onlineboutique | 31808 | 31880 | +72 |
| reverser | 26900 | 26904 | +4 |

We also discussed putting the stubs in a separate file guarded with a [build tag](https://pkg.go.dev/go/build) that is only set for tests. There's no built-in tag that is set for tests (see https://github.com/golang/go/issues/21360), so this would be challenging. Given that the binary sizes didn't increase significantly, I also think this approach isn't needed.

examples/chat/weaver_gen.go
```diff
@@ -997,6 +1009,136 @@ func (s sQLStore_server_stub) getImage(ctx context.Context, args []byte) (res []
...
+
+type imageScaler_reflect_stub struct {
+	caller func(reflect.Type, string, []reflect.Value) []reflect.Value
+}
+
+// Check that imageScaler_reflect_stub implements the ImageScaler interface.
+var _ ImageScaler = (*imageScaler_reflect_stub)(nil)
+
+func (s imageScaler_reflect_stub) Scale(ctx context.Context, a0 []byte, a1 int, a2 int) (r0 []byte, err error) {
+	component := reflect.TypeOf((*ImageScaler)(nil)).Elem()
```
ghemawat, line 1022: Could we reduce the size of generated code. Perhaps something like:

```
results, err := s.caller("Scale", []any{a0, a1, a2})
return results[0].([]byte), err
```

I am assuming the component type (ImageScaler here) is already baked into the s.caller closure.

ghemawat (commented):
Looks nice and simple.

examples/chat/weaver_gen.go
mwhittaker: Great idea! I updated the code with your suggestion. There's no longer any use of `reflect.Value`, so let me know if you think we should call these stubs something different than "reflect stubs".

runtime/version/version.go
mwhittaker: I updated the version in the code but still plan on releasing a new tag every two weeks.

examples/chat/weaver_gen.go
```diff
@@ -997,6 +1009,71 @@ func (s sQLStore_server_stub) getImage(ctx context.Context, args []byte) (res []
...
+type localCache_reflect_stub struct {
+	caller func(string, context.Context, []any) ([]any, error)
+}
+
+// Check that localCache_reflect_stub implements the LocalCache interface.
+var _ LocalCache = (*localCache_reflect_stub)(nil)
+
+func (s localCache_reflect_stub) Get(ctx context.Context, a0 string) (r0 string, err error) {
+	results, err := s.caller("Get", ctx, []any{a0})
+	return results[0].(string), err
```
ghemawat, line 1035: Do we need to avoid touching results when err is non-nil?

mwhittaker: I was thinking we should still preserve the return values even if err is non-nil in case a method returns meaningful values alongside a non-nil error? I think [Buffer.ReadBytes](https://pkg.go.dev/bytes#Buffer.ReadBytes) does this, for example.

mwhittaker: Ah, sorry, you're saying that if err is non-nil, the returned `results` could be an empty slice for example, in which case indexing into it would panic? I overlooked that. I can update the comments to say that that `caller` should always return a valid `results`?

ghemawat: Yes, I was worried about: (1) results being too short (2) entries in results not having the right type.

I worry that just saying call should always return a valid results may be easy to miss. What code are we calling: if it is generated code, it might be fine to place such a restriction. If it is hand-written code, I would worry about that.

mwhittaker: Great point. It is hand-written. I could generate some code to panic if `results` is not well-formed? For example: 

```go
func (s localCache_reflect_stub) Get(ctx context.Context, a0 string) (r0 string, err error) {
    results, err := s.caller("Get", ctx, []any{a0})
    if len(results) != 1 {
        panic(fmt.Errorf("..."))
    }
    if _, ok := results[0].(string); !ok {
        panic(fmt.Errorf("..."))
    }
    return results[0].(string), err
}
```

ghemawat: What if we made s.caller responsible for filling the return variables:

```
err = s.caller("Get", ctx, []any{a0}, []any{&r0})
return
```

That will keep the generated code size short. This stub will become a bit more expensive, but that is probably fine for the intended use.

mwhittaker: Great idea! I switched to your suggestion.

ghemawat (approved):
Looks great. Thanks.

PR #482, opened 2023-07-24
Added bare-bones "simulator".

mwhittaker:
This PR introduces a skeleton of a determinstic simulator. The simulator constructs a couple of replicas of every registered component. The `weaver.Ref` fields inside the components are instantiated with reflect stubs that, for now, just call the requested method.

The simulator in its current state isn't very useful. In the future, I'll expand the simulator to actually simulate things. I'll also add more comments in the future once we settle on an API. I'm also doing everything in an internal package for now.


internal/sim/sim.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+type Options struct {
+	Seed           int64
+	MaxReplicas    int
+	ConfigFilename string
+	Config         string
+}
+
+type Simulator struct {
```
spetrovic77, line 40: Is this thing meant to implement the `Weavelet` interface? If so, should we rename it to `SimulatorWeavelet`?

(I suppose that if you you mean to drive it using `weaver.Run`, then you'd have to make it implement the `Weavelet` interface.)

internal/sim/sim.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/ServiceWeaver/weaver/internal/config"
+	"github.com/ServiceWeaver/weaver/internal/weaver"
+	"github.com/ServiceWeaver/weaver/runtime"
+	"github.com/ServiceWeaver/weaver/runtime/codegen"
+	"github.com/ServiceWeaver/weaver/runtime/protos"
+	"golang.org/x/exp/slog"
+)
+
+type Options struct {
```
spetrovic77, line 33: Comments for a bunch of these?

spetrovic77 (approved):
Very nice and simple change. I'm still a bit unsure how this all fits together (e.g., what calls the `call` method).

internal/sim/sim.go
mwhittaker: Done.

internal/sim/sim.go
mwhittaker: It implements `GetIntf` but not `GetImpl`, so it doesn't implement `Weavelet` for now. I might remove (or unexport) `GetIntf` in the future as well. Not sure yet. I don't plan on using `weaver.Run` though!

mwhittaker (commented):
> Very nice and simple change. 

Thanks Srdjan!

> I'm still a bit unsure how this all fits together (e.g., what calls the call method).

Sorry, it is confusing. Let me try and explain. When you call `Simulator.GetIntf` for some interface type `T`, `GetIntf` returns a reflect stub that implements `T`. When you call methods on the reflect stub, they are executed by calling `Simulator.call`, which picks a random replica and executes the method using reflection. In the future, the simulator will not immediately execute method calls and will instead re-order them, but that's not implemented in this PR.

mwhittaker:
@spetrovic77 mind taking another look?

PR #485, opened 2023-07-28
Test calls to a mix of healthy and unhealthy backends.

ghemawat:
Added a call test with two backends, one healthy, and one unhealthy. The test currently doesn't work. Some ideas on how to fix things are listed in a comment.

### DIFF
--- a/internal/net/call/call_test.go
+++ b/internal/net/call/call_test.go
@@ -374,6 +374,87 @@ func (d *deadEndpoint) Address() string {
 	return fmt.Sprintf("dead://%s", d.name)
 }
 
+// hangingConn is a net.Conn that never actually responds to calls.
+type hangingConn struct {
+	net.Conn // Inherit normal behavior from wrapped conn
+}
+
+func (h hangingConn) Write(b []byte) (n int, err error) {
+	// Throw away the message.
+	fmt.Fprintf(os.Stderr, "dropping write to %s\n", h.RemoteAddr().String())
+	return len(b), nil
+}
+
+// hangingEndpoint returns a hangingConn.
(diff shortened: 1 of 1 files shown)


internal/net/call/call_test.go
```diff
@@ -1093,6 +1174,74 @@ func TestRefreshDraining(t *testing.T) {
...
+	// endpoint.
+	//
+	// 3. We return a special error to startCall if we are attempting to
+	// use a connection that hasn't completed its initial health
+	// check. startCall retries after a delay if it sees this error.
+	//
+	// Option 2 seems the best? Its main potential downside is that we may
+	// make too many connections. However we can avoid having too many
+	// connections when there are many clients and many servers by
+	// implementing backend subsetting inside the Resolver.
```
mwhittaker, line 1203: Option 2 seems the best to me as well. 

Just summarizing to make sure I understand. A `Resolver` returns a set of endpoints. A client eagerly establishes connections to these endpoints and performs health checks against them. Whenever the set of healthy connections changes, the client updates the `Balancer` with them. Client requests are only made on the healthy connections picked by a `Balancer`. If a request is pending on a connection and we think the connection is unhealthy, we abort the call if it is not retriable or retry it on another connection if it is retriable. Am I understanding correctly?

mwhittaker (approved):

internal/net/call/call_test.go
ghemawat: Yes. Though initially, I was imagining we just do a single health-check at connection setup time, which will simplify the implementation.

mwhittaker: Got it, sounds great to me!

PR #491, opened 2023-08-01
Added blog post on "weaver generate".

mwhittaker:


website/blog/index.html
```diff
@@ -15,6 +15,15 @@
 -->
 
 <!-- TODO: Avoid boilerplate. -->
+<div class="blog-listing">
+  <div>
+    <a class="blog-title" href="weaver_generate.html">Using Advanced Go Features to Detect Stale Code</a>
+    <span class="blog-date">TODO</span>
+  </div>
+  <div class="blog-author">Michael Whittaker</div>
+  <div>A deep dive on the advanced, and sometimes esoteric, ways we use Go to detect when you forget to run <code>weaver generate</code>.</div>
```
spetrovic77, line 24: A passing note: a ton of folks in the workshop were forgetting to run `weaver generate` when they added a listener field to their component implementation. It would be nice to detect this.

website/blog/weaver_generate.md
```diff
@@ -0,0 +1,499 @@
...
+
+We can then define `weaver.Unrouted` as an alias of `weaver.RoutedBy[private]`:
+
+```go
+type Unrouted interface {
+    routedBy(private)
+}
+```
+
+Now, every component struct that embeds `weaver.Implements` implements the
```
spetrovic77, line 346: Do you need this paragraph, it's obvious 🙄 .

Just kidding, this is great reading for hard-core folks!

spetrovic77 (approved):
Looks great!

website/blog/weaver_generate.md
mwhittaker: Haha, yeah this stuff gets pretty hard-core!

website/blog/index.html
mwhittaker: Great idea! I think we could detect when someone removes or changes a listener field. I can't think of a good way to detect when someone adds a new listener field though.

We could easily detect whenever a component implementation struct changes at all and require that the user re-runs `weaver generate`, but some changes to a component implementation struct don't require re-running `weaver generate`. I guess there's a trade-off between running `weaver generate` more than necessary vs forgetting to run `weaver generate`. I'll think harder about it!

mwhittaker (commented):
Thanks Srdjan!

PR #493, opened 2023-08-02
Report weavelet stdout/stderr on error.

mwhittaker:
This PR is a slight tweak of #447 by @giautm to always report the stdout and stderr from a crashed weavelet, not just on EPIPE or EOF. It also kills and waits for the weavelet subprocess if there's an error during the handshake. See #447 for more context.


spetrovic77 (approved):
Nice!

PR #496, opened 2023-08-03
Color log messages based on component.

mwhittaker:
Before this PR, log messages were printed white. This PR changes them to show in the same color as the component. I think this makes the logs look uglier, but it is useful to be able to visually track the logs from a particular weavelet or component.

| Before | After |
| - | - |
| dimmed | colored |

### DIFF
--- a/runtime/logging/pretty.go
+++ b/runtime/logging/pretty.go
@@ -175,7 +175,7 @@ func (pp *PrettyPrinter) Format(e *protos.LogEntry) string {
 
 	// Write the message.
 	pp.b.WriteString("] ")
-	pp.b.WriteString(pp.colorize(levelColor, e.Msg))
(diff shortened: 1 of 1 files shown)


runtime/logging/pretty.go
```diff
@@ -175,7 +175,7 @@ func (pp *PrettyPrinter) Format(e *protos.LogEntry) string {
 
 	// Write the message.
 	pp.b.WriteString("] ")
-	pp.b.WriteString(pp.colorize(levelColor, e.Msg))
+	pp.b.WriteString(pp.colorize(colors.ColorHash(e.Node), e.Msg))
```
spetrovic77, line 178: As we discussed offline, let's make the color be component-based.

spetrovic77 (approved):

PR #498, opened 2023-08-03
Avoid using connections until they are healthy.

ghemawat:
* Change Balancer to track the set of healthy connections.
* Added state machine to clientConnection.
* Create clientConnection as soon as a resolver returns an address.
* Do version handshake on connection before adding it to the balancer.
* Removed CallOptions.Balancer (call.Connection now has one balancer).
* Dropped unused Sharded balancer.
* Dropped some obsolete tests.
* Tweaked some tests to account for changed behavior.
* Added deadlines to some tests to make them behave better when things hang due to a bug.
* Undid earlier bad renaming of object to component.
* Split a large test into multiple tests.

### DIFF
--- a/internal/net/call/balancer.go
+++ b/internal/net/call/balancer.go
@@ -14,63 +14,54 @@
 
 package call
 
-import (
-	"fmt"
-	"math/rand"
-)
+// ReplicaConnection is a connection to a single replica. A single Connection
+// may consist of many ReplicaConnections (typically one per replica).
+type ReplicaConnection interface {
+	// Address returns the name of the endpoint to which the ReplicaConnection
+	// is connected.
+	Address() string
+}
 
-// A Balancer picks the endpoint to which which an RPC client performs a call. A
-// Balancer should only be used by a single goroutine.
+// Balancer manages a set of ReplicaConnections and picks one of them per
+// call. A Balancer should only be used by a single goroutine.
 //
 // TODO(mwhittaker): Right now, balancers have no load information about
 // endpoints. In the short term, we can at least add information about the
 // number of pending requests for every endpoint.
-//
-// TODO(mwhittaker): Right now, we pass a balancer the set of all endpoints. We
-// instead probably want to pass it only the endpoints for which we have a
-// connection. This means we may have to form connections more eagerly.
-//
-// TODO(mwhittaker): We may want to guarantee that Update() is never called
-// with an empty list of addresses. If we don't have addresses, then we don't
-// need to do balancing.
 type Balancer interface {
-	// Update updates the current set of endpoints from which the Balancer can
-	// pick. Before Update is called for the first time, the set of endpoints
-	// is empty.
-	Update(endpoints []Endpoint)
+	// Add adds a ReplicaConnection to the set of connections.
+	Add(ReplicaConnection)
 
-	// Pick picks an endpoint. Pick is guaranteed to return an endpoint that
-	// was passed to the most recent call of Update. If there are no endpoints,
-	// then Pick returns an error that includes Unreachable.
-	Pick(CallOptions) (Endpoint, error)
+	// Remove removes a ReplicaConnection from the set of connections.
+	Remove(ReplicaConnection)
+
+	// Pick picks a ReplicaConnection from the set of connections.
+	// Pick returns _,false if no connections are available.
+	Pick(CallOptions) (ReplicaConnection, bool)
 }
 
 // balancerFuncImpl is the implementation of the "functional" balancer
 // returned by BalancerFunc.
 type balancerFuncImpl struct {
-	endpoints []Endpoint
-	pick      func([]Endpoint, CallOptions) (Endpoint, error)
+	connList
+	pick func([]ReplicaConnection, CallOptions) (ReplicaConnection, bool)
 }
 
 var _ Balancer = &balancerFuncImpl{}
 
-// BalancerFunc returns a stateless, purely functional load balancer that uses
-// the provided picking function.
-func BalancerFunc(pick func([]Endpoint, CallOptions) (Endpoint, error)) Balancer {
+// BalancerFunc returns a stateless, purely functional load balancer that calls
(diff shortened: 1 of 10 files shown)


internal/net/call/call.go
```diff
@@ -43,14 +43,14 @@ package call
...
 // When the client wants to send an RPC, it selects one of its server
-// connections to use, creates a call component, assigns it a new request-id, and
-// registers the components in a map in the connection. It then sends a request
-// message over the connection and waits for the call component to be marked as
+// connections to use, creates a call object, assigns it a new request-id, and
+// registers the object in a map in the connection. It then sends a request
+// message over the connection and waits for the call object to be marked as
 // done.
 //
 // When the response arrives, it is picked up by readResponses().
```
mwhittaker, line 51: I think some of the comments on `readResponses()` are stale now.

internal/net/call/call.go
```diff
@@ -454,198 +480,276 @@ func (rc *reconnectingConnection) watchResolver(ctx context.Context, version *Ve
...
-		cbuf:     bufio.NewReader(nc),
-		mu:       &rc.mu,
-		version:  initialVersion, // Updated when we hear from server
-		calls:    map[uint64]*call{},
-		lastID:   0,
+	switch c.state {
+	case checking, idle, active:
+		c.setState(disconnected)
+	case draining:
+		c.setState(missing)
```
mwhittaker, line 623: Could you elaborate on why draining goes to missing instead of disconnected?

internal/weaver/routing.go
```diff
@@ -28,22 +28,44 @@ import (
...
+		conns:     map[string]call.ReplicaConnection{},
+	}
 }
 
-// Update implements the call.Balancer interface.
-func (rb *routingBalancer) Update(endpoints []call.Endpoint) {
-	rb.balancer.Update(endpoints)
+// Add adds c to the set of connections we are balancing across.
+func (rb *routingBalancer) Add(c call.ReplicaConnection) {
+	rb.balancer.Add(c)
```
mwhittaker, line 55: I don't think RoundRobin balancers are thread safe. Should we do this with `rb.mu` held as well?

internal/weaver/routing.go
```diff
@@ -28,22 +28,44 @@ import (
...
+
+	rb.mu.Lock()
+	defer rb.mu.Unlock()
+	rb.conns[c.Address()] = c
+}
+
+// Remove removes c from the set of connections we are balancing across.
+func (rb *routingBalancer) Remove(c call.ReplicaConnection) {
+	rb.balancer.Remove(c)
```
mwhittaker, line 64: Same as above, should we do this with `rb.mu` held?

mwhittaker (approved):
Very readable for such a complex change!

internal/net/call/call.go
```diff
@@ -454,198 +480,276 @@ func (rc *reconnectingConnection) watchResolver(ctx context.Context, version *Ve
...
-	if rc.closed {
-		return nil, fmt.Errorf("Call on closed Connection")
+// State transition actions: all of these are called with rc.mu held.
+
+func (c *clientConnection) register() {
+	switch c.state {
+	case missing:
+		c.setState(disconnected)
+	case draining:
+		c.setState(active)
```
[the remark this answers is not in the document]
ghemawat: Good catch. Fixed.

internal/net/call/call.go
ghemawat: Thanks for catching that.

internal/net/call/call.go
```diff
@@ -129,33 +115,79 @@ type reconnectingConnection struct {
...
+	mu     sync.Mutex
+	conns  map[string]*clientConnection
+	closed bool
 
 	resolver       Resolver
 	cancelResolver func()         // cancels the watchResolver goroutine
 	resolverDone   sync.WaitGroup // used to wait for watchResolver to finish
 }
 
+// connState is the state of a clientConnection (connection to a particular
```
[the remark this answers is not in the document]
ghemawat: It was very helpful when I was trying to understand what I was doing. And I like the ability to isolate all the state transitions in one place (setState).

internal/weaver/routing.go
ghemawat: I think this might be okay?

Balancer requires external synchronization. (It is protected by rc.mu in call.go.) The comment in balancer.go isn't great though; I will improve it.

routingBalancer.update() provides a path where rc.mu is not held, but that state is guarded by routingBalancer.mu and does not touch the balancer part.

internal/net/call/call.go
ghemawat: My initial reasoning was as follows: for some reason the server just transiently dropped out of the resolver results. Since the connection is still around in draining mode, it should be fine to just mark it as non-drained and continue using it.

Though I am now a bit worried about a quick server restart. The old draining connection might be pointing at a dead server incarnation, so maybe we should not try to send new requests on that. Though even that will be just a partial fix since the resolver might have missed the quick restart and we would never get into a draining state. I think the right fix might be to include an incarnation number in the assignment and use the <endpoint,incarnation> pair as the key?

For now, I added a comment saying we are intentionally reusing the old connection since it is still around.

internal/weaver/routing.go
ghemawat: Ditto.

internal/net/call/balancer.go
```diff
@@ -80,37 +71,35 @@ func RoundRobin() *roundRobin {
...
+}
+
+func (cl *connList) Remove(c ReplicaConnection) {
+	for i, elem := range cl.list {
+		if elem != c {
+			continue
 		}
-		return endpoints[opts.ShardKey%uint64(n)], nil
-	})
+		// Replace removed entry with last entry.
```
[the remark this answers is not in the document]
ghemawat: Common idiom from my Clu days.

ghemawat (commented):
Thanks!

PR #499, opened 2023-08-04
Improve call.go logging and errors.

ghemawat:
* Pass appropriate details to endCalls for better call failure errors.
* Drop connection and balancer addresses from log messages. These were intended to be temporary debugging aids.


mwhittaker (approved):
Nice!

PR #500, opened 2023-08-04
Validate registrations when app is launched.

mwhittaker:
This PR introduces a `validateRegistrations` function that checks the validity of component registrations. For now, it simply checks that every `weaver.Ref[T]` field refers to a registered component interface `T`. This mistake is actually quite common, when you add a component and forget to run `weaver generate`.

Before this PR, if a non-main component had a `weaver.Ref[T]` to an unregistered component, the app would run but silently not work. With this change, all weavelets will imediately crash, and for `go run .` and `weaver multi deploy`, the app will terminate immediately with a helpful error message:

```shell
$ weaver multi deploy weaver.toml
start main process: NewEnvelope: connect to weavelet: read protobuf
length: EOF
-----BEGIN STDERR-----
2023/08/04 13:31:50 component implementation struct main.odd has field
weaver.Ref[main.this_is_not_a_registered_component], but component
main.this_is_not_a_registered_component was not registered; maybe you
forgot to run 'weaver generate'

-----END STDERR-----
exit status 1
```

## Alternatives

Srdjan had a great idea of checking the validity of `weaver.Ref[T]`s at compile time, but I couldn't think of any way to do that. If we do think of something, we should switch to it.

I did have one idea that turned out to be terrible. We could require users to embed a special `weaver.Interface` interace into their component interfaces:

```go
// In weaver package.
type Interface interface { isComponentInterface() }
type Ref[T Interface] struct { ... }
type Implements[T Interface] struct { ... }
func (Implements[T]) isComponentInterface()

// In user's code.
type ExampleComponent {
    weaver.Interface
    Foo(context.Context) error
}
```

This is not only ugly and super janky, it also doesn't work. It checks that a `weaver.Ref[T]` refers to a component interface `T`, but it doesn't check that `T` was actually registered.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -1,0 +1,59 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package weaver
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+
+	"github.com/ServiceWeaver/weaver/internal/reflection"
+	"github.com/ServiceWeaver/weaver/runtime/codegen"
+)
+
+// validateRegistrations validates the provided registrations, returning an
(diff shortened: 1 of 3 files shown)


validate.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"errors"
+	"fmt"
+	"reflect"
+
+	"github.com/ServiceWeaver/weaver/internal/reflection"
+	"github.com/ServiceWeaver/weaver/runtime/codegen"
+)
+
+// validateRegistrations validates the provided registrations, returning an
```
spetrovic77, line 26: nit: s/an/a/

spetrovic77 (approved):
Very nice!

PR #501, opened 2023-08-04
Validate listener names at application startup.

mwhittaker:
This PR extends the `validateRegistrations` function to also validate listener names when a Service Weaver application begins. As with PR #500, this PR aims to catch permanent and easily avoidable programming errors immediately.

I also fixed a small bug in listener name validation. Previously, if a listener had a `weaver:""` annotation, it was being ignored. Now, it is being rejected (since `""` is not a valid listener name).

I also updated the website documentation to describe legal listener names.

### DIFF
--- a/fill.go
+++ b/fill.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"net"
 	"reflect"
-	"unicode"
 
 	"github.com/ServiceWeaver/weaver/internal/reflection"
 	"github.com/ServiceWeaver/weaver/internal/weaver"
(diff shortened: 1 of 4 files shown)


fill.go
```diff
@@ -101,7 +100,7 @@ func fillListeners(impl any, get func(name string) (net.Listener, string, error)
 
 		// The listener's name is the field name, unless a tag is present.
 		name := t.Name
-		if tag := t.Tag.Get("weaver"); tag != "" {
+		if tag, ok := t.Tag.Lookup("weaver"); ok {
```
spetrovic77, line 103: This was intentional: I was writing an `empty` service and wanted to use the empty listener name. 🤣 

Thanks for fixing!

spetrovic77 (approved):
A lot of people in our workshops ran into a problem where they added a listener to weaver.Main, but forgot to run `weaver generate`.

Do you know if we print a helpful error in that case?

mwhittaker:
> A lot of people in our workshops ran into a problem where they added a listener to weaver.Main, but forgot to run `weaver generate`.
> 
> Do you know if we print a helpful error in that case?

I don't think we do. Let me work on that in another PR.

PR #516, opened 2023-08-11
Use new slog, slices, maps packages in Go 1.21.

mwhittaker:
Go 1.21 was just released. This PR replaces our use of exp/slog, exp/slices, and exp/maps with the new log/slog, slices, and maps packages in the standard library.


mwhittaker:
Oops, this makes us incompatible with Go 1.20. I'll abandon the change.

PR #520, opened 2023-08-11
Switch to Go 1.21.

ghemawat:
* Use the standard library slog, maps, and slices packages.
* Use sync.OnceValue instead of hand-rolling our own wrapper.


mwhittaker (approved):
Nice! Did you want to update `exp/maps` to `maps` as well? And do you mind updating the "Installation" section of `website/docs.md` to say that we require 1.21 instead of 1.20?

ghemawat:
> Nice! Did you want to update `exp/maps` to `maps` as well?

Can't. We need maps.Keys and maps.Values which don't seem to have made it into the standard library.

> And do you mind updating the "Installation" section of `website/docs.md` to say that we require 1.21 instead of 1.20?

Done.

PR #521, opened 2023-08-11
Removed SingleProcess field from EnvelopeInfo.

mwhittaker:
The SingleProcess field was no longer being used. Since this is a change to the deployer API, I also bumped the deployer API version.


rgrandl (approved):
Nice!

PR #522, opened 2023-08-11
Removed Pid field from WeaveletInfo proto.

mwhittaker:
This PR removes the mostly unused Pid field from the WeaveletInfo proto. The field is unnecessary, as an envelope already knows the pid of the weavelet its running.


rgrandl (approved):
Nice! Glad to see code being deleted

PR #526, opened 2023-08-14
Replaced InternalPort with InternalAddress.

mwhittaker:
Recall that a weavelet listens on an internal listener to serve remote procedure calls from other weavelets. This PR simplifies how a weavelet picks the address for this internal listener.

Previously, the address was dictated by two fields inside `EnvelopeInfo`: `SingleMachine` and `InternalPort`. The weavelet would listen on `$HOSTNAME:InternalPort` if `SingleMachine` was false and `localhost:InternalPort` otherwise.

This PR removes the `SingleMachine` field and replaces `InternalPort` with `InternalAddress`. A weavelet listens on the `InternalAddress` provided by the envelope.


rgrandl (approved):
Nice simplification! That SingleMachine param was really annoying.

PR #528, opened 2023-08-15
Added unit tests for remote weavelet.

mwhittaker:
This PR adds unit tests for the remote weavelet. Specifically, it tests that the weavelet still operates correctly even when some of its interactions with the envelope fail. For example, the weavelet should retry failed ActivateComponent calls. A handful of the tests are currently failing because of "bugs" in the remote weavelet. I skipped the tests for now and will try to fix the bugs in future PRs.

Fixing the bugs is a bit tricky because it's not always straightforward the best way to handle failures. For example, if a weavelet calls GetListenerAddress and receives an address that is already in use, should it keep retrying the GetListenerAddress? Or should it just self-terminate?


internal/testdeployer/remoteweavelet_test.go
```diff
@@ -0,0 +1,620 @@
...
+	}
+	got, err := x.(a).A(d.ctx, want)
+	if err != nil {
+		d.t.Fatal(err)
+	}
+	if got != want {
+		d.t.Fatalf("A(%d): got %d, want %d", want, got, want)
+	}
+
+	// Test component b.
```
spetrovic77, line 240: Just for my understanding: are these tests necessary, given that there's a call chain A -> B -> C ?

spetrovic77 (approved):
Awesome testing!!

PR #529, opened 2023-08-15
Fixed RemoteWeavelet UpdateRoutingInfo bugs.

mwhittaker:
This PR fixes four bugs in RemoteWeavelet's UpdateRoutingInfo.

1. Before this PR, UpdateRoutingInfo would panic if you passed it a nil RoutingInfo. This PR fixes UpdateRoutingInfo to instead return an error.
2. Recall that a component is either always local or always remote. We don't currently allow a component to switch between the two. This PR adds code to UpdateRoutingInfo to check this.
3. Before this PR, we were using the assignment from one component as the assignment for all components' load collectors.
4. UpdateRoutingInfo updates a resolver, balancer, and load collector. Before this PR, it was possible to update the load collector but not the resolver or balancer. This PR reorders things so either all are updated or none of them are. The update is still not transactional, as concurrent calls to UpdateRoutingInfo can interleave. We should consider fixing that in a future PR.

### DIFF
--- a/internal/weaver/remoteweavelet.go
+++ b/internal/weaver/remoteweavelet.go
@@ -430,10 +430,15 @@ func (w *RemoteWeavelet) UpdateComponents(req *protos.UpdateComponentsRequest) (
 // UpdateRoutingInfo implements the conn.WeaverHandler interface.
 func (w *RemoteWeavelet) UpdateRoutingInfo(req *protos.UpdateRoutingInfoRequest) (reply *protos.UpdateRoutingInfoReply, err error) {
 	defer func() {
-		name := logging.ShortenComponent(req.RoutingInfo.Component)
-		routing := fmt.Sprint(req.RoutingInfo.Replicas)
-		if req.RoutingInfo.Local {
-			routing = "local"
+		name := "?"
+		routing := "?"
+		if req.RoutingInfo != nil {
+			name = logging.ShortenComponent(req.RoutingInfo.Component)
+			if req.RoutingInfo.Local {
+				routing = "local"
+			} else {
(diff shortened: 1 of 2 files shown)


internal/weaver/remoteweavelet.go
```diff
@@ -430,10 +430,15 @@ func (w *RemoteWeavelet) UpdateComponents(req *protos.UpdateComponentsRequest) (
 // UpdateRoutingInfo implements the conn.WeaverHandler interface.
 func (w *RemoteWeavelet) UpdateRoutingInfo(req *protos.UpdateRoutingInfoRequest) (reply *protos.UpdateRoutingInfoReply, err error) {
 	defer func() {
-		name := logging.ShortenComponent(req.RoutingInfo.Component)
-		routing := fmt.Sprint(req.RoutingInfo.Replicas)
-		if req.RoutingInfo.Local {
-			routing = "local"
+		name := "?"
```
spetrovic77, line 433: Nice catch.

This is too magical, would the following make sense:

```
func (w *RemoteWeavelet) UpdateRoutingInfo(...) (...) {
  if req.RoutingInfo == nil {
    w.syslogger.Debug("nil RoutingInfo")
    return nil, fmt.Errorf("nil RoutingInfo")
  }
  info := req.RoutingInfo
  defer func() {
    // assume info != nil
    ...
  }()
  ...
```

spetrovic77 (approved):
🏆

internal/weaver/remoteweavelet.go
mwhittaker: Much better, thanks!

PR #530, opened 2023-08-15
Fixed bugs in RemoteWeavelet UpdateComponents.

mwhittaker:
This PR fixes two small bugs in UpdateComponents.

1. UpdateComponents now returns an error if the provided components don't exist. Previously, the error was logged but not returned.
2. Previously, if a component failed to construct, the construction of all subsequent components was short circuited. Now, we try to construct all components, even if some fail to construct.

### DIFF
--- a/internal/weaver/remoteweavelet.go
+++ b/internal/weaver/remoteweavelet.go
@@ -24,7 +24,6 @@ import (
 	"net"
 	"os"
 	"reflect"
-	"slices"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


internal/weaver/remoteweavelet.go
```diff
@@ -402,29 +416,21 @@ func (w *RemoteWeavelet) UpdateComponents(req *protos.UpdateComponentsRequest) (
...
-			if err != nil {
-				// TODO(mwhittaker): Propagate errors.
-				w.syslogger.Debug(fmt.Sprintf("Failed to update components %v", shortened), "err", err)
-				return
-			}
-			if _, err = w.GetImpl(c.reg.Impl); err != nil {
-				// TODO(mwhittaker): Propagate errors.
-				w.syslogger.Debug(fmt.Sprintf("Failed to update components %v", shortened), "err", err)
-				return
+	if len(components) > 0 {
```
spetrovic77, line 419: Can we short-circuit this to be:

```
if len(components) == 0 {
  return &protos.UpdateComponentsReply{}, nil
}
...
```

internal/weaver/remoteweavelet.go
```diff
@@ -392,6 +391,21 @@ func (w *RemoteWeavelet) GetLoad(*protos.GetLoadRequest) (*protos.GetLoadReply,
 
 // UpdateComponents implements the conn.WeaverHandler interface.
 func (w *RemoteWeavelet) UpdateComponents(req *protos.UpdateComponentsRequest) (*protos.UpdateComponentsReply, error) {
+	var errs []error
```
spetrovic77, line 394: Do/should we retry this on the deployer side?

spetrovic77 (approved):

internal/weaver/remoteweavelet.go
mwhittaker: Done, thanks!

internal/weaver/remoteweavelet.go
mwhittaker: Great question. I still haven't wrapped my head around exactly what should be retried. This case is tricky because UpdateComponents currently only returns an error if a deployer asks to start a component that wasn't registered. So, retrying would yield the same error. 

Creating the components themselves happens in a goroutine and can also fail, and right now we don't have a good way to report that error to the deployer. I'll have to think more about that.

mwhittaker (commented):
Thanks Srdjan!

PR #531, opened 2023-08-15
Replaced os.Exit with RemoteWeavelet Wait method.

mwhittaker:
Recall that a RemoteWeavelet maintains (1) an RPC server to serve remote component method calls from other weavelets and (2) a connection to the envelope. Before this PR, a RemoteWeavelet would os.Exit if either (1) or (2) went down. This made testing failure scenarios impossible, as the process would simply exit.

This PR removes the calls to os.Exit. Instead, I added a Wait method to the RemoteWeavelet that blocks until the weavelet has shut down and returns any errors encountered during shutdown. With this new Wait method, I was able to add a unit test to test the behavior of a weavelet when its connection to the envelope breaks.


spetrovic77 (approved):
Nice, complex, change!

PR #538, opened 2023-08-16
Switched to using staticcheck linter.

mwhittaker:
Previously, we used [golangci-lint](https://github.com/golangci/golangci-lint), but the Go team recommended we use staticcheck instead.


.github/workflows/go.yml
```diff
@@ -41,12 +41,12 @@ jobs:
       - name: Cache linter
         uses: actions/cache@v3
         with:
-          path: ~/go/bin/golangci-lint
-          key: golangci-lint-v1.51.2
+          path: ~/go/bin/staticcheck
+          key: staticcheck-v0.4.3
```
ghemawat, line 45: Can you extend the commit message to say why we are doing this and what the pros and cons are for staticcheck vs golangci-lint.

internal/benchmarks/benchmarks_test.go
```diff
@@ -362,10 +361,6 @@ func BenchmarkPing(b *testing.B) {
 	}
 }
 
-func init() {
-	rand.Seed(time.Now().UnixNano())
-}
-
```
ghemawat, line 368: Why are we removing this seed initialization?

staticcheck.conf
```diff
@@ -0,0 +1,2 @@
+# See https://staticcheck.dev/docs/configuration/options/#checks.
+checks = ["all", "-ST1000", "-ST1003", "-ST1005", "-ST1012", "-ST1016", "-ST1020", "-ST1021", "-ST1022"]
```
ghemawat, line 2: Can we add comments saying what the dropped checks are and why we are dropping them?  I think some of them are dropped to match the default value, but some are new.

ghemawat (commented):
Thanks!

mwhittaker (commented):
Thanks Sanjay!

staticcheck.conf
```diff
@@ -1,2 +1,17 @@
-# See https://staticcheck.dev/docs/configuration/options/#checks.
+# In addition to the default checks [1], we also disable the following style
```
ghemawat, line 1: How about:
  In addition to the checks disabled by default [1], we also ...

ghemawat (approved):
Drop the "financially ..." part from the change description. Doesn't seem relevant to what checker we pick.

staticcheck.conf
mwhittaker: Much clearer, thanks!

mwhittaker (commented):
> Drop the "financially ..." part from the change description. Doesn't seem relevant to what checker we pick.

Got it, will do.

PR #539, opened 2023-08-16
internal/net/call now supports optional retries.

ghemawat:
Added call.CallOptions.Retry, which causes calls to be retried after a communication or unreachable error. Subsequent changes will make ServiceWeaver methods use this functionality when appropriate.


mwhittaker (approved):
Looks awesome! Looking forward to enabling retries by default :)

PR #542, opened 2023-08-18
Clarify "weaver generate" version error message.

mwhittaker:
Recall that `weaver generate` generates code to check, at compile time, that the codegen version used by `weaver generate` is the same as the codegen version linked into the application. This check includes a lengthy error message that is shown to the user when they try to `go build` or `go run` with a mismatched codegen version:

```
ERROR: You generated this file with 'weaver generate' v0.20.0 (codegen
version v0.20.0). The generated code is incompatible with the version of the
github.com/ServiceWeaver/weaver module that you're using. The weaver module
version can be found in your go.mod file or by running the following command.

    go list -m github.com/ServiceWeaver/weaver

We recommend updating the weaver module and the 'weaver generate' command by
running the following.

Then, re-run 'weaver generate' and re-build your code. If the problem persists,
please file an issue at https://github.com/ServiceWeaver/weaver/issues.
```

People were reading `weaver_gen.go` and seeing this error message, thinking that something went wrong with `weaver generate`. This PR tries to clarify the situation in two ways.

1. I moved the generated error message to the middle of `weaver_gen.go`. Before it was at the very top, and people saw it immediately when opening `weaver_gen.go`. Now, it's buried more in the middle, making it less likely for people to read it.
2. I added a comment above the error message explaining that it's nothing to worry about.

### DIFF
--- a/examples/chat/weaver_gen.go
+++ b/examples/chat/weaver_gen.go
@@ -15,26 +15,6 @@ import (
 	"time"
 )
 
-var _ codegen.LatestVersion = codegen.Version[[0][20]struct{}](`
-
-ERROR: You generated this file with 'weaver generate' v0.20.0 (codegen
-version v0.20.0). The generated code is incompatible with the version of the
-github.com/ServiceWeaver/weaver module that you're using. The weaver module
-version can be found in your go.mod file or by running the following command.
-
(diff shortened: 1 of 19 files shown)


rgrandl (approved):
Nice, can't imagine how many people got into this and they thought that something is wrong.

PR #544, opened 2023-08-18
Enabled more lint checks.

mwhittaker:
This PR enables some of the previously disabled linter checks:

- ST1016: This checks that method reciever names are consistent. The code already satisfied this check.
- ST1020: This checks that exported functions are documented with a comment that begins with the function's name. Enabling this check caught a couple of typos in the code.
- ST1022: This checks that exported variables are documented with a comment that begins with the variable's name. Enabling this check caught one bad comment.

I also considered enabling the other default checks, but wasn't sure if it was worth it:

- ST1000: This checks that every package has a package comment. For some internal packages, like those in weavertest/internal, this check felt unnecessary.
- ST1003: This checks for bad variable names. The vast majority of flagged variable names involved the capitalization of "ID". We often use variables like "DeploymentId", but the linter wants "DeploymentID". I couldn't find any official guidance on this kind of capitalization, so I left the check disabled.
- ST1021: This checks that exported types are documented with a comment that begins with the type's name. I want to enable this check, but it doesn't handle generic types. For example, the comment above RoutedBy begins `// RoutedBy[T] ...`, but the check wants `// RoutedBy ...`

### DIFF
--- a/runtime/codegen/decoder.go
+++ b/runtime/codegen/decoder.go
@@ -68,7 +68,7 @@ func (d *Decoder) DecodeProto(value proto.Message) {
 	}
 }
 
-// DecodeBinaryMarshaler deserializes the value from a byte slice using
+// DecodeBinaryUnmarshaler deserializes the value from a byte slice using
 // UnmarshalBinary.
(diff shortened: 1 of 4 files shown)


ghemawat (approved):
Thanks!

PR #546, opened 2023-08-18
Implemented basic simulation.

mwhittaker:
This PR implements the core logic of the deterministic simulator. The simulator maintains a buffer of pending calls and pending replies. It repeatedly performs one of the following actions:

1. Start an op.
2. Deliver a method call.
3. Deliver a method reply.

## Starting An Op

Recall the definition of an op:

```go
type Op[T any] struct {
    Name string
    Gen func(*rand.Rand) T
    Func any
}
```

When a simulator starts an op, it calls the `Gen` function to generate a random value. It then invokes `Func`, passing it handles to the necessary components. These handles are generated using reflect stubs created with the simulator's `call` method. `Func` runs freely until it calls a component method. At that point, the reflect stub is invoked, the arguments of the call are recorded for later execution, and the call blocks on a channel to receive return values. When the op terminates, the simulation is aborted if the op returns an error.

## Delivering a Method Call

When the simulator delivers a call, it picks a component replica to execute the call and then calls the component's method directly. From here, the executing method acts like an executing op. When the call returns, the return values are recorded for later delivery.

## Delivering a Method Reply

When the simulator delivers a method reply, it writes the return values of the method to the channel on which the caller is blocked. This will signal the caller to wake up and continue executing.

## Control Flow

Most of the complexity of the simulator comes from its complicated control flow. Execution is similar to [cooperative threading][1]. The various ops and method calls all happen on different goroutines and are never preempted. Instead, they relinquish control back to the simulator whenever they call a component method or return.

[1]: https://en.wikipedia.org/wiki/Cooperative_multitasking

### DIFF
--- a/internal/sim/components.go
+++ b/internal/sim/components.go
@@ -22,24 +22,121 @@ import (
 
 //go:generate ../../cmd/weaver/weaver generate .
 
-// This file contains components used to test the simulator in sim_test.go.
+// This file contains components used to test the simulator in sim_test.go. The
+// callgraph is intentionally overcomplicated to exercise the simulator.
 
-type swapper interface {
-	Swap(context.Context, int, int) (int, int, error)
+// Component interfaces.
+
+type divMod interface {
+	// DivMod(n, d) returns n/d, n%d.
+	DivMod(context.Context, int, int) (int, int, error)
 }
 
-type swapperImpl struct {
-	weaver.Implements[swapper]
+type div interface {
+	// Div(n, d) returns n/d.
+	Div(context.Context, int, int) (int, error)
 }
 
-func (s *swapperImpl) Swap(_ context.Context, x, y int) (int, int, error) {
-	return y, x, swapperError{}
+type mod interface {
+	// Mod(n, d) returns n%d.
+	Mod(context.Context, int, int) (int, error)
 }
 
-type swapperError struct {
+type identity interface {
+	// Identity(x) returns x.
+	Identity(context.Context, int) (int, error)
+}
+
+type blocker interface {
+	// Block blocks until the provided context is cancelled.
+	Block(context.Context) error
+}
+
+// Component implementation structs.
+
+type divModImpl struct {
+	weaver.Implements[divMod]
+	div weaver.Ref[div]
+	mod weaver.Ref[mod]
+}
+
+type divImpl struct {
+	weaver.Implements[div]
+	identity weaver.Ref[identity]
+}
+
+type modImpl struct {
+	weaver.Implements[mod]
+	identity weaver.Ref[identity]
+}
+
+type identityImpl struct {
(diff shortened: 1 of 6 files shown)


internal/sim/components.go
```diff
@@ -22,24 +22,121 @@ import (
 
 //go:generate ../../cmd/weaver/weaver generate .
 
-// This file contains components used to test the simulator in sim_test.go.
+// This file contains components used to test the simulator in sim_test.go. The
```
spetrovic77, line 25: Can we add "test" to the name of the file, to make it clear it's only used in local tests?

internal/sim/components.go
```diff
@@ -22,24 +22,121 @@ import (
...
+// callgraph is intentionally overcomplicated to exercise the simulator.
 
-type swapper interface {
-	Swap(context.Context, int, int) (int, int, error)
+// Component interfaces.
+
+type divMod interface {
+	// DivMod(n, d) returns n/d, n%d.
```
spetrovic77, line 31: I thought that you are computing `(n/d) % d`, which is an awesomer (and more useless) operation. :smile:

Although, maybe I just came up with a tool to solve the collatz conjecture! :rofl:

internal/sim/sim.go
```diff
@@ -220,54 +313,133 @@ func validateOp[T any](s *Simulator, o Op[T]) (op, error) {
...
-	// Call the method.
-	out := reflect.ValueOf(replicas[i]).MethodByName(method).Call(in)
+	// Compute the set of candidate steps.
+	const (
+		runOp = iota
+		deliverCall
+		deliverReply
+	)
+	var candidates []int
+	if s.numOps < s.opts.NumOps {
```
spetrovic77, line 348: This is an interesting algorithm for picking work items. I don't have anything better to suggest, just wanted to point out that it may converge fairly quickly to `s.opts.NumOps`. But correct me if I'm wrong, that is what we want since we'll start with smaller NumOps and then increase as needed?

spetrovic77 (approved):
This is really nice work!

internal/sim/components.go
mwhittaker: Unfortunately not. If we do, then `weaver generate` ignores it.

internal/sim/components.go
mwhittaker: :laughing: I took the idea from Python's [divmod](https://docs.python.org/3/library/functions.html#divmod).

internal/sim/sim.go
mwhittaker: Ah, you're giving me too much credit. I just implemented the simplest possible thing I could think of! In the future, I hope to think harder about smart ways to interleave things.

mwhittaker (commented):
Thanks Srdjan!

PR #548, opened 2023-08-19
Inject RemoteCallErrors into simulations.

mwhittaker:
Recall that component method calls can fail with RemoteCallErrors. This PR extends the simulator to randomly return RemoteCallErrors from method calls. Note that the simulator sometimes returns a RemoteCallError without executing a method call and sometimes returns a RemoteCallError after executing a method. Both are possible.

### DIFF
--- a/internal/sim/sim.go
+++ b/internal/sim/sim.go
@@ -26,6 +26,7 @@ import (
 	"reflect"
 	"sync"
 
+	core "github.com/ServiceWeaver/weaver"
 	"github.com/ServiceWeaver/weaver/internal/config"
 	"github.com/ServiceWeaver/weaver/internal/reflection"
 	"github.com/ServiceWeaver/weaver/internal/weaver"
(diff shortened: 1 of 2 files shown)


internal/sim/sim.go
```diff
@@ -348,16 +349,18 @@ func (s *Simulator) step() {
...
+		deliverCallError
+		deliverReplyError
 	)
 	var candidates []int
 	if s.numOps < s.opts.NumOps {
 		candidates = append(candidates, runOp)
 	}
 	if len(s.calls) > 0 {
-		candidates = append(candidates, deliverCall)
+		candidates = append(candidates, deliverCall, deliverCallError)
```
spetrovic77, line 360: Will this lead to shallow executions, where too many calls/replies end up failing?

Fine to leave as-is for now.

spetrovic77 (approved):

internal/sim/sim.go
mwhittaker: Yeah, I think you're 100% right. I implemented the simplest thing for now, but in the future, I'll have to improve things. I don't have anything super concrete in mind, but one idea for example, is that we should probably simulate executions without any injected failures first and only inject errors if we can't find a failing execution.

PR #557, opened 2023-08-23
Sped up bin_test.go using t.Parallel().

mwhittaker:
I noticed on GitHub Actions that bin_test was taking up to two minutes to execute [1]. This PR speeds things up by running things in parallel. The test also redundantly builds the same binary again and again. We could speed up the tests even further by not rebuilding things, but I'll leave that for another PR as it introduces a bit of complexity.

[1]: https://github.com/ServiceWeaver/weaver/actions/runs/5954182034/job/16150143324

### DIFF
--- a/runtime/bin/bin_test.go
+++ b/runtime/bin/bin_test.go
@@ -27,12 +27,17 @@ import (
 )
 
 func TestReadComponentGraph(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mwhittaker:
Huh, this actually made things slower. I'll have to debug further.

PR #560, opened 2023-08-23
Fixed bug in GitHub Actions go caching.

mwhittaker:
The setup-go action's default caching behavior does not work well with matrix tests. By default, setup-go caches the contents of `go env GOCACHE` and `go env GOMODCACHE` under a key derived from the OS, the go version, and a hash of the contents of go.sum [1].

When we run a matrix test, every job uses the same cache key. As a result, the jobs race, and the first to finish determines the contents of the cache. Because caches are write-once, all other jobs fail to update the cache. In subsequent runs, all jobs use the contents of the cache, even if the cache was generated by a different job.

For example, if the "lint" job finishes first, then all future jobs like "build" and "test" will use the cache produced by the "lint" job. However, this cache is likely missing a lot of cached package and build outputs, causing to the "build" and "test" jobs to redundantly download and build dependencies again and again.

To avoid this problem, we manually cache the contents of `go env GOCACHE` and `go env GOMODCACHE` using a key derived from the OS, the go version, the contents of go.sum, and the matrix command. See [2] for more documentation on custom cache keys and paths.

[1]: https://github.com/actions/setup-go/blob/93397bea11091df50f3d7e59dc26a7711a8bcfbe/docs/adrs/0000-caching-dependencies.md
[2]: https://github.com/actions/cache#creating-a-cache-key


spetrovic77 (approved):
Wowza. Nice find!

mwhittaker:
> Wowza. Nice find!

Thanks! Some of the tests were running super slow which led me down a rabbit hole to figure out why things weren't working properly.

PR #561, opened 2023-08-23
Verify registered listeners when starting the app.

spetrovic77:
This ensures that the application listeners match the information stored in the read-only section of the application binary, and used by various deployers. If there is a mismatch, the user will be prompted to run `weaver generate`.

This fixes one of the common issues that came up in workshops, where people forgot to re-run `weaver generate` after declaring the listener.

### DIFF
--- a/dev/build_and_test
+++ b/dev/build_and_test
@@ -54,7 +54,7 @@ function cmd_vet() {
 
 function cmd_lint() {
   if ! exists staticcheck; then
-    printf "staticcheck not found; install via\ngo install honnef.co/go/tools/cmd/staticcheck@0.4.3\n" >&2
+    printf "staticcheck not found; install via\ngo install honnef.co/go/tools/cmd/staticcheck@v0.4.3\n" >&2
     exit 1
   fi
 
--- a/internal/tool/generate/generator.go
+++ b/internal/tool/generate/generator.go
@@ -510,6 +510,16 @@ func extractComponent(opt Options, pkg *packages.Package, file *ast.File, tset *
 		return nil, err
 	}
 
+	// Check that listener names are unique.
+	seenLis := map[string]struct{}{}
+	for _, lis := range listeners {
+		if _, ok := seenLis[lis]; ok {
+			return nil, errorf(pkg.Fset, spec.Pos(),
+				"component implementation %s declares multiple listeners with name %s. Please disambiguate.", formatType(pkg, impl), lis)
+		}
+		seenLis[lis] = struct{}{}
+	}
+
 	// Warn the user if the component has a mistyped Init method. Init methods
 	// are supposed to have type "func(context.Context) error", but it's easy
 	// to forget to add a context.Context argument or error return. Without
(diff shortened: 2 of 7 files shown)


dev/build_and_test
```diff
@@ -54,7 +54,7 @@ function cmd_vet() {
 
 function cmd_lint() {
   if ! exists staticcheck; then
-    printf "staticcheck not found; install via\ngo install honnef.co/go/tools/cmd/staticcheck@0.4.3\n" >&2
+    printf "staticcheck not found; install via\ngo install honnef.co/go/tools/cmd/staticcheck@v0.4.3\n" >&2
```
mwhittaker, line 57: Oh, good catch, thanks!

validate.go
```diff
@@ -48,38 +48,36 @@ func validateRegistrations(regs []*codegen.Registration) error {
...
+					err := fmt.Errorf("component implementation struct %v has a listener field %v, but listener %v hasn't been registered; maybe you forgot to run 'weaver generate'", reg.Impl, name, name)
 					errs = append(errs, err)
 				}
 			}
 		}
 	}
 	return errors.Join(errs...)
 }
-
-// isValidListenerName returns whether the provided name is a valid
```
mwhittaker, line 69: I replaced `token.IsIdentifier` with `isValidListenerName` to remove the dependency on the `token` package, as recommended by Sanjay here: https://github.com/ServiceWeaver/weaver/pull/473#discussion_r1268593783. Happy to revisit the decision and switch back to using `token`, but just wanted to give the full context.

validate.go
```diff
@@ -48,38 +48,36 @@ func validateRegistrations(regs []*codegen.Registration) error {
...
+					name = tag
+				}
+				found := false
+				for _, lis := range reg.Listeners {
+					if lis == name {
+						found = true
+						break
+					}
+				}
+				if !found {
```
mwhittaker, line 75: How about `if !slices.Contains(reg.Listeners, name) {`?

validate_test.go
```diff
@@ -83,8 +86,9 @@ func TestValidateUnregisteredRef(t *testing.T) {
 // TestValidateInvalidListenerNames tests that validateRegistrations fails on
 // invalid listener names.
 func TestValidateInvalidListenerNames(t *testing.T) {
-	type foo struct{}
+	type foo interface{}
 	type fooImpl struct {
+		_ Listener // undeclared name
```
mwhittaker, line 91: Mind updating the error list starting on line 109 to have an error to match this listener?

internal/tool/generate/generator.go
```diff
@@ -510,6 +510,16 @@ func extractComponent(opt Options, pkg *packages.Package, file *ast.File, tset *
 		return nil, err
 	}
 
+	// Check that listener names are unique.
```
mwhittaker, line 513: Nice!

mwhittaker (approved):
Very nice! Thanks Srdjan!

PR #566, opened 2023-08-28
Added basic tracing to the simulator.

mwhittaker:
Recall that a simulation consists of a number of discrete events:

- An op starts.
- An op finishes.
- A method call is made.
- A method call is delivered.
- A method call returns.
- A method return return is delivered.
- An error is injected.

This PR adds basic tracing to the simulator to keep track of these events. When a simulation fails, a history of the events is returned to the user. They can inspect the history to understand why the simulation failed. In a future PR, I'll add a way to visualize a history of events.

In the future, traces could be used for more than just visualization. For example, we may use traces for minimization or for checking that a simulation is deterministic. Thus, I expect the contents of the traces to change over time as we refine the simulator. I went with something simple for now.

### DIFF
--- a/internal/sim/sim.go
+++ b/internal/sim/sim.go
@@ -82,11 +82,88 @@ type Simulator struct {
 	ctx   context.Context // simulation context
 	group *errgroup.Group // group with all running goroutines
 
-	mu      sync.Mutex // guards the following fields
-	rand    *rand.Rand // random number generator
-	numOps  int        // number of spawned ops
-	calls   []*call    // pending calls
-	replies []*reply   // pending replies
+	mu          sync.Mutex // guards the following fields
+	rand        *rand.Rand // random number generator
+	numOps      int        // number of spawned ops
+	calls       []*call    // pending calls
+	replies     []*reply   // pending replies
+	history     []Event    // history of events
+	nextTraceId int        // next trace id
+	nextSpanId  int        // next span id
+}
+
+// An Event represents an atomic step of a simulation.
+type Event interface {
+	isEvent()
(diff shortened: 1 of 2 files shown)


internal/sim/sim.go
```diff
@@ -82,11 +82,88 @@ type Simulator struct {
...
+	SpanId  int // span id
+}
+
+func (OpStart) isEvent()       {}
+func (OpFinish) isEvent()      {}
+func (Call) isEvent()          {}
+func (DeliverCall) isEvent()   {}
+func (Return) isEvent()        {}
+func (DeliverReturn) isEvent() {}
+func (DeliverError) isEvent()  {}
```
spetrovic77, line 161: Do you want to add a bunch of:

```
var _ Event = OpStart{}
...
```

internal/sim/sim.go
```diff
@@ -112,6 +191,30 @@ type reply struct {
 	returns []reflect.Value // the call's return values
 }
 
+// We store trace and span ids in the context using the following keys.
+type traceIdKey struct{}
```
spetrovic77, line 195: nit: s/Id/ID/

(everywhere)

spetrovic77 (approved):

internal/sim/sim.go
mwhittaker: Done.

internal/sim/sim.go
mwhittaker: Done.

mwhittaker (commented):
Thanks Srdjan!

PR #568, opened 2023-08-29
Added basic implementation of a simulator corpus.

mwhittaker:
See corpus.go for a description of what the simulator corpus is, how it's designed, why it's important, and what the next steps are. Note that this PR implements a very primitive corpus. There's lots to improve in the future.

Here's an example of a corpus:

```shell
$ tree testdata
testdata/
└── sim
    ├── TestCancelledSimulation
    │   └── a52f5ec5f94e674d.json
    ├── TestSimulateCorpusEntries
    │   ├── 2bfe847328319dae.json
    │   └── a52f5ec5f94e674d.json
    └── TestUnsuccessfulSimulation
        ├── 2bfe847328319dae.json
        └── a52f5ec5f94e674d.json

5 directories, 5 files
$ cat testdata/sim/TestCancelledSimulation/a52f5ec5f94e674d.json
{
    "version": 1,
    "seed": 0,
    "num_replicas": 10,
    "num_ops": 1000
}
```


internal/sim/corpus.go
```diff
@@ -0,0 +1,149 @@
...
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// # Simulator Corpus
```
spetrovic77, line 26: Reading the change description, I thought that a "corpus" was going to represent a series of `Simulators`, i.e., a collection of simulation runs with different setting and randomizations. Instead, a corpus is a set of failed runs, which is less intuitive IMO.

Maybe a better name for a collection of failed runs is something that includes failure in the name. "RIPs" is a playful example of this. If you're less worried about the name confusion, fine to leave as-is too.

internal/sim/corpus.go
```diff
@@ -0,0 +1,149 @@
...
+//
+// When the simulator runs a failed execution, it persists the failing inputs
+// to disk. The collection of saved failing inputs is called a *corpus*, and
+// each individual entry is called a *corpus entry*. When a simulator is
+// created, the first thing it does is load and re-simulate all corpus entries.
```
spetrovic77, line 31: You mentioned offline that these re-runs may not be very useful to re-run, if the code (and thereby the interleavings) change in the process of fixing the failed runs. I suppose there's some value still, just wanted to get your (updated) opinion.

internal/sim/corpus.go
mwhittaker: What about corpses instead of corpus :stuck_out_tongue_closed_eyes:, just kidding. 

Good point. The go fuzzing library uses the term corpus, but they do differ from us in that you can manually add non-failing inputs to the corpus. What do you think about renaming corpus to *graveyard* and corpus entry to *graveyard entry*? Also happy with something a little more bland like *failing corpus* or *counterexample corpus* or something like that.

internal/sim/corpus.go
mwhittaker: Great question. Here's a recap of my thought process. 

Initially, I wanted to persist the history of a failing execution and replay it step-by-step. If the re-execution diverged from the original execution, I wanted to return an error. Later, I realized that this didn't make any sense because as the user changes their code, the execution will understandably change. 

However, I still think there is benefit to re-executing failing simulations on a best effort basis. For example, if a user encounters a failing simulation, I think it would be very natural for them to add print statements to their code and re-run the simulation. Adding print statements shouldn't cause the simulation to diverge.

I think in the future, we could take this a step further. We could store the history of failing simulations and try to replay them as close to the original execution as possible. As I said above, there may be divergence, but I think there will be cases where we can get close to the original execution. For example, if a user changes how they generate random inputs, this affects the seed of the random number generator which causes the entire simulation to behave differently. However, their code might be the same as before, and we can re-execute the failing simulation perfectly.

mwhittaker (commented):
Thanks Srdjan!

internal/sim/corpus.go
spetrovic77: "Graveyard" sounds great to me!

internal/sim/corpus.go
spetrovic77: SGTM!

internal/sim/corpus.go
mwhittaker: Renamed!

internal/sim/graveyard.go
```diff
@@ -0,0 +1,153 @@
...
+
+// # Simulator Graveyard
+//
+// When the simulator runs a failed execution, it persists the failing inputs
+// to disk. The collection of saved failing inputs is called a *graveyard*, and
+// each individual entry is called a *graveyard entry*. When a simulator is
+// created, the first thing it does is load and re-simulate all graveyard
+// entries.
+//
+// We borrow the design of go's fuzzing library' corpus with only minor changes
```
spetrovic77, line 34: s/library'/library's/

spetrovic77 (approved):

internal/sim/graveyard.go
mwhittaker: Oops, good catch.

PR #569, opened 2023-08-29
Added hyperparameter for simulator failure rate.

mwhittaker:
Recall that the simulator can arbitrarily choose to fail a component method call, either before or after delivering the call. Before this PR, the simulator had an incredibly naive algorithm for deciding when to fail calls. This naive approach failed a lot of calls.

This PR introduces a hyperparameter `FailureRate` to configure how often the simulator fails calls. When FailureRate is 0, the simulator never fails a call, and when FailureRate is 1, the simulator fails every call.

### DIFF
--- a/internal/sim/sim.go
+++ b/internal/sim/sim.go
@@ -194,9 +194,19 @@ type op struct {
 type fate int
 
 const (
-	dontFail           fate = iota // dont fail a call
-	failBeforeDelivery             // fail a call before it is delivered
-	failAfterDelivery              // fail a call after it is delivered
+	// Don't fail the call.
+	dontFail fate = iota
+
+	// Fail the call before it is delivered to a component replica. In this
+	// case, the method call does not execute at all.
+	failBeforeDelivery
+
+	// Fail the call after it has been delivered to a component replica. In
+	// this case, the method call will fail after fully executing.
+	//
+	// TODO(mwhittaker): Deliver a method call, and then fail it while it is
+	// still executing.
+	failAfterDelivery
 )
 
 // call is a pending method call.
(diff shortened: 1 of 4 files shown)


internal/sim/sim.go
```diff
@@ -189,10 +190,20 @@ type op struct {
 	components []reflect.Type // Op.Func component argument types
 }
 
+// fate dictates if and how a call should fail.
+type fate int
+
+const (
+	dontFail           fate = iota // dont fail a call
```
spetrovic77, line 197: nit: don't

internal/sim/sim.go
```diff
@@ -189,10 +190,20 @@ type op struct {
...
+	failBeforeDelivery             // fail a call before it is delivered
```
spetrovic77, line 198: Can you clarify here what "delivered" means.

spetrovic77 (approved):
Very clean!

internal/sim/sim.go
mwhittaker: Fixed.

internal/sim/sim.go
mwhittaker: Done. Let me know if you think I should elaborate further :)

internal/sim/sim.go
```diff
@@ -362,9 +373,23 @@ func (s *Simulator) call(caller string, replica int, reg *codegen.Registration,
 	spanID := s.nextSpanID
 	s.nextSpanID++
 
+	// Determine the fate of the call.
+	fate := dontFail
+	if flip(s.rand, s.opts.FailureRate) {
+		// TODO(mwhittaker): Have two parameters to control the rate of failing
+		// before and after delivery? This level of control might be
+		// unnecessary. For now, we pick between them equiprobably.
```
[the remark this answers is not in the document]
mwhittaker: Thanks :nerd_face:. I took this word from [the discrete math class](https://www.cs.cornell.edu/~rafael/discmath.pdf) I took freshman year of college.

mwhittaker (commented):
Thanks Srdjan!

PR #570, opened 2023-08-30
Allow methods to be marked as non-retriable.

ghemawat:
A component method can now be marked as non-retriable by adding a blank assignment of the form:

```
var _ weaver.NotRetriable = Component.Method
```

We currently don't auto-retry, so such annotations are no-ops, but will become meaningful after we enable auto-retries.

Refactored some stub creation code out of remoteweavelet.go and made it easier to test.

### DIFF
--- a/internal/tool/generate/generator.go
+++ b/internal/tool/generate/generator.go
@@ -247,6 +247,19 @@ func newGenerator(opt Options, pkg *packages.Package, fset *token.FileSet, autom
 			components[c.fullIntfName()] = c
 		}
 	}
+
+	// Find method attributes.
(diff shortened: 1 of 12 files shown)


internal/sim/testdata/sim/TestCancelledSimulation/a52f5ec5f94e674d.json
mwhittaker, line 1: Oops, I didn't mean for this `testdata` directory to be committed. Just opened #571 to ignore it.

mwhittaker (approved):
Awesome, very nice!

PR #572, opened 2023-08-30
Added simulator yield rate to tune interleaving.

mwhittaker:
This PR introduces a new hyperparameter to the simulator, *yield rate*, that dictates how much an execution is interleaved. Recall that a simulation consists of a number of operations (ops), and every op executes a series of steps, mostly calling component methods.

After executing a step, an op makes a decision to keep executing or to yield control back to the scheduler. The yield rate determines how often the op yields. With a yield rate of 0, the op never yields. Ops are run to completion, and there is no interleaving. With a yield rate of 1, ops yield after every step, leading to a very high degree of interleaving. With a yield rate of $\frac{1}{n}$, an op yields after $n$ steps in expectation. In general, yielding is governed by a [geometric distribution][geometric].

Note that some of the code in this PR is really inefficient. I left it as is to keep thing easy to understand, but will optimize things in a future PR.

[geometric]: https://en.wikipedia.org/wiki/Geometric_distribution

### DIFF
--- a/internal/sim/sim.go
+++ b/internal/sim/sim.go
@@ -94,9 +94,9 @@ type Simulator struct {
 	rand        *rand.Rand       // random number generator
 	current     int              // currently running op
 	numStarted  int              // number of started ops
-	notFinished map[int]struct{} // set of not finished ops
-	calls       map[int][]*call  // pending calls, by op id
-	replies     map[int][]*reply // pending replies, by op id
+	notFinished map[int]struct{} // not finished op trace ids (a set for efficient removal)
+	calls       map[int][]*call  // pending calls, by trace id
+	replies     map[int][]*reply // pending replies, by trace id
 	history     []Event          // history of events
 	nextTraceID int              // next trace id
 	nextSpanID  int              // next span id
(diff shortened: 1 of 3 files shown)


internal/sim/sim.go
```diff
@@ -89,14 +90,16 @@ type Simulator struct {
...
-	calls       []*call    // pending calls
-	replies     []*reply   // pending replies
-	history     []Event    // history of events
-	nextTraceID int        // next trace id
-	nextSpanID  int        // next span id
+	mu          sync.Mutex       // guards the following fields
+	rand        *rand.Rand       // random number generator
+	current     int              // currently running op
+	numStarted  int              // number of started ops
+	notFinished map[int]struct{} // set of not finished ops
```
spetrovic77, line 97: Can you explain what these are keyed by and why they are a map and not a slice? (To be able to detect quickly when everything is finished.)

internal/sim/sim.go
```diff
@@ -273,6 +279,10 @@ func New(name string, opts Options) (*Simulator, error) {
 	// Create simulator.
 	//
 	// TODO(mwhittaker): Take a *testing.T and use the test name as the name.
+	notFinished := map[int]struct{}{}
+	for i := 1; i < opts.NumOps+1; i++ {
+		notFinished[i] = struct{}{}
```
spetrovic77, line 284: Any way to replace `i` with `traceID`, to make it clearer? This is also assuming that trace ids map to ops 1-1. Also, you use the term "op id" in other places.

spetrovic77 (approved):

mwhittaker (commented):
Thanks Srdjan!

PR #573, opened 2023-08-30
Optimized how simulator picks ops.

mwhittaker:
Recall that the simulator has to keep track of the set of ops that haven't yet finished. The simulator performs two main operations on this set:

1. It picks random elements from it.
2. It removes elements from the set.

Before this PR, the simulator maintained this set as a builtin Go `map`. Deterministically picking a key from the map was slow. This PR replaces the `map` with a more optimized data structure. I took the idea from Sanjay's code in PR #498.

This speeds up TestSuccessfulSimulation from about 0.4 seconds to 0.1 second. There are still more optimizations to do in future PRs.

### DIFF
--- a/internal/sim/rand.go
+++ b/internal/sim/rand.go
@@ -73,3 +73,64 @@ func flip(r *rand.Rand, p float64) bool {
 	}
 	return r.Float64() <= p
 }
+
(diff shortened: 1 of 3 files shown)


spetrovic77 (approved):
Nice!

PR #575, opened 2023-08-30
Enable automatic retrying on a network error.

ghemawat:
Remote method calls are now automatically retried after network errors. This functionality can be disabled on a method-by-method basis by marking the method as not-retriable:

```go
var _ weaver.NotRetriable = Component.Method
```

weavertest's RPC runner helps to discover methods that should be marked as not-retriable by artificially retrying all unmarked methods an extra time.

### DIFF
--- a/internal/weaver/stub.go
+++ b/internal/weaver/stub.go
@@ -24,10 +24,11 @@ import (
 
 // stub holds information about a client stub to the remote component.
 type stub struct {
-	component string          // name of the remote component
-	conn      call.Connection // connection to talk to the remote component
-	methods   []stubMethod    // per method info
-	tracer    trace.Tracer    // component tracer
+	component     string          // name of the remote component
+	conn          call.Connection // connection to talk to the remote component
+	methods       []stubMethod    // per method info
+	tracer        trace.Tracer    // component tracer
+	injectRetries int             // Number of artificial retries per retriable call
 }
 
 type stubMethod struct {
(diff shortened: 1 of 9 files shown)


mwhittaker (approved):
Very elegant!

internal/weaver/stub.go
```diff
@@ -43,15 +44,20 @@ func (s *stub) Tracer() trace.Tracer {
...
+		Retry:    m.retry,
 		ShardKey: shardKey,
 	}
-	return s.conn.Call(ctx, m.key, args, opts)
+	n := 1
+	if m.retry {
+		n += s.injectRetries
+	}
+	for i := 0; i < n; i++ {
+		result, err = s.conn.Call(ctx, m.key, args, opts)
```
pmatseykanets, line 58: Q: Shouldn't retry logic have some sort of backoff e.g. exponential?

mwhittaker: Hi @pmatseykanets! Good question. The real retry logic does have exponential backoff and is performed within a call to `s.conn.Call`: 

https://github.com/ServiceWeaver/weaver/blob/2ed7c7f552b4b90b802fac25518a138325571e82/internal/net/call/call.go#L377-L384

This code is artificially injecting fake retries into tests, which doesn't need backoff.

pmatseykanets: @mwhittaker Nice! Thank you so much for explaining!

ghemawat: Thanks for the comment. Probably worth explaining with a comment, which I will do.

PR #579, opened 2023-09-01
More logging prettifications.

ghemawat:
* remoteweavelet no longer merges dynamic values into the initial log string.
* Explicitly save the values we need for log line comparisons instead of cloning the entire LogEntry.
* Logging level is never dimmed.
* Timestamp prefix is always dimmed if same as in previous entry.
* Change initial padding for component and node columns.
* File name and line are now always colored by hashing the file name.
* Change separator between log line header and log line contents.
* Log message is not colored.
* Attribute keys are now always dimmed.
* Attribute values are now colored (by hashing the key so that all attribute values for the same key can be spotted at a glance). As an exception, error values are always red, and component values are always colored by hashing the component name.

### DIFF
--- a/internal/weaver/remoteweavelet.go
+++ b/internal/weaver/remoteweavelet.go
@@ -201,7 +201,7 @@ func NewRemoteWeavelet(ctx context.Context, regs []*codegen.Registration, bootst
 		return nil
 	})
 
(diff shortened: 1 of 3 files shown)


spetrovic77 (approved):
Wow, that added up to a lot of changes!

PR #582, opened 2023-09-05
Remove the use of co-location groups from the dashboard.

rgrandl:
In #536 we removed the colocation group from status server. This PR removes the use of the colocation group from the dashboard too.


mwhittaker (approved):
Nice!

PR #584, opened 2023-09-05
Make sure the callgraph always includes a node for weaver.Main.

spetrovic77:
This node can be missed if, e.g., the binary only contains weaver.Main and no other components.


mwhittaker (approved):
Good catch!

PR #585, opened 2023-09-05
Small call_test.go cleanup of span propagation test.

ghemawat:
Change the span propagation test to use the testing facility for running custom functions at the server.


spetrovic77 (approved):
Nice, much cleaner.

PR #586, opened 2023-09-06
Cover hostname test with IPv6

tiennv1997:
Use [net.JoinHostPort](https://pkg.go.dev/net?tab=doc#JoinHostPort) to join host and port instead of `fmt.Sprintf`. It fixed case when `ips[0]` is `::1`


mwhittaker (approved):
Nice! Never knew about `net.JoinHostPort`.

PR #588, opened 2023-09-06
Added rough draft of simulator API.

mwhittaker:
See https://github.com/ServiceWeaver/weaver/pull/565 for details.


internal/sim/api.go
```diff
@@ -0,0 +1,340 @@
...
+// its first argument and must return a single error value. A simulation is
+// aborted when a method returns a non-nil error.
+//
+// TODO(mwhittaker): For now, the Init method is required. In the future, we
+// could make it optional and use default generators for methods.
+type Workload interface {
+	Init(Registrar) error
+}
+
+// WorkloadPointer[T] is a *T that implements the Workload interface.
```
spetrovic77, line 92: Can you expand on why this is needed? It seems like you need it for New(), but does it have to be public?

internal/sim/api.go
```diff
@@ -0,0 +1,340 @@
...
+// A Simulator deterministically simulates a Service Weaver application. See
+// the package documentation for instructions on how to use a Simulator.
+type Simulator struct{}
+
+// An Event represents an atomic step of a simulation.
+type Event interface {
+	isEvent()
+}
+
+// TODO(mwhittaker): Prefix all events with Event so they show up together in
```
spetrovic77, line 112: 👍

internal/sim/api.go
```diff
@@ -0,0 +1,340 @@
...
+	// generators must match the method. For example, given the method:
+	//
+	//     Foo(context.Context, int, bool) error
+	//
+	// we must register a Generator[int] and a Generator[bool]:
+	//
+	//     var r Registrar = ...
+	//     var i Generator[int] = ...
+	//     var b Generator[bool] = ...
+	//     r.RegisterGenerators("Foo", i, b)
```
spetrovic77, line 61: What happens if a generator isn't specified for a method? Is the plan to default to integers{}, float64s{} etc?

internal/sim/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+type divModWorkload struct{}
+
+func (d *divModWorkload) Init(r Registrar) error {
+	r.RegisterGenerators("DivMod", integers{}, positives{})
+	r.RegisterGenerators("Div", integers{}, positives{})
+	r.RegisterGenerators("Mod", integers{}, positives{})
+	return nil
+}
+
+func (d *divModWorkload) DivMod(context.Context, int, int) error { return nil }
```
spetrovic77, line 51: Have you considered passing Generators directly here, in the method signature?

I can see it being weird because these methods will be called with generated values, not with generators. But maybe it can be done this way, with global type registration:

```
type positive int
type positives struct{}
func (positives) Generate(r *rand.Rand) positive { return 1 + r.Intn(math.MaxInt)}
type integers struct{}
func (integers) Generate(r *rand.Rand) int { return r.Int() }

func init() {
  RegisterGenerator(integers{})
  RegisterGenerator(positives{})
}

func (d *divModWorkload) DivMod(context.Context, int, positive) error { return nil }
```

spetrovic77 (commented):
Very nice, only meaningful comment is on Generators. I found it weird that the user has to register generator types for every method: it feels duplicative.

internal/sim/api.go
```diff
@@ -0,0 +1,340 @@
...
+
+// A Generator[T] generates random values of type T.
+type Generator[T any] interface {
+	// Generate returns a randomly generated value of type T. While Generate is
+	// "random", it must be deterministic. That is, given the same instance of
+	// *rand.Rand, Generate must always return the same value.
+	Generate(*rand.Rand) T
+}
+
+// A Registrar is used to register stuff with a simulator.
```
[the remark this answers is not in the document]
mwhittaker: Done.

internal/sim/api.go
mwhittaker: I'm not sure yet. We could use a set of defaults (like integers{}), but I also don't think it's too wild to require that the user register generators for every method. For now, I require it, but let's revisit later?

internal/sim/api.go
```diff
@@ -0,0 +1,340 @@
...
+	RegisterGenerators(method string, generators ...any)
+}
+
+// A Workload defines the set of operations to run as part of a simulation.
+// Every workload is defined as a named struct. To perform a simulation, a
+// simulator constructs an instance of the struct, calls the struct's Init
+// method, and then randomly calls the struct's exported methods. For example,
+// the following is a simple workload:
+//
+//	type workload struct {}
```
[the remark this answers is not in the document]
mwhittaker: Done.

internal/sim/api.go
mwhittaker: I'll do this in a future PR then!

internal/sim/api.go
mwhittaker: Yeah, it is only needed for `New`. I think technically it doesn't have to exported, but since it is a type constraint in `New` (which is exported), I think it makes the most sense for it to also be exported? As an alternative, I could remove `WorkloadPointer` and simplify `New` to the following: 

```go
New[W Workload](*testing.T) *Simulator
```

The difference is that people would have to call `New` like `New[*myWorkload]` instead of `New[myWorkload]`. What's your preference? Happy to switch.

internal/sim/api_test.go
mwhittaker: Good question. We definitely explored the idea of methods receiving generators. I disliked the idea for a couple of reasons. Primarily, I placed a high value on method signatures being as simple and as direct as possible.

- In the body of your `DivMod`, for example, the user would have to cast `positive` to `int`. This is a bit roundabout and also error-prone as they might cast to the wrong type (e.g., `int32`). 
- We could work around this by having `positive` (or `positives`) have a `Get` method or something similar that returns an `int`, but that complicates things further. 
- In your `DivMod` example, I also feel it's a bit discordant that the first argument gets to be a plain `int` while `positive` does not. If you later change how you want to generate the first argument, you'd have to change its type and also update the body of `DivMod` to perform casting (or calling a `Get` method).

Overall, I felt our approach requires more typing but is also more straightforward with fewer subtleties. My plan is to stick with registering generators in `Init` for now, get the simulator working on some examples, and reconsider the API once I have a bit more experience with it. How does that sound?

mwhittaker (commented):
Thanks for the thorough review Srdjan!

internal/sim/api_test.go
spetrovic77: Good point about type casting. I'm okay with us moving forward with what you have.

Whatever we do, I like that both your API and my proposal have the property that it is clear these methods signatures are primitive values, and not some types like `Gen[int]`. The conceptual model for the user should be that once the method is called, the values have already been generated, and nothing conveys that better than the signature consisting of primitive types.

internal/sim/api.go
spetrovic77: SGTM

internal/sim/api.go
spetrovic77: I like your idea of getting rid of it, because these extra type will just confuse the user (e.g., should I be implementing the WorkloadPointer interface or Workload?)

mwhittaker: Sounds good. Mind if I do this in a follow up PR?

spetrovic77 (approved):

PR #589, opened 2023-09-07
Implemented bare bones simulation with new API.

mwhittaker:
This PR hooks up the new simulation API to the existing simulation implementation code. This mostly consists of massaging a workload struct into types that the underlying simulation engine understands. For now, the simulator only runs a single simulation on a fixed set of hyperparameters. In the future, it will sweep over hyperparameters given a time budget.


internal/sim/api.go
```diff
@@ -192,28 +225,52 @@ type Results struct {
...
+	// Gather the set of registered components.
+	regs := codegen.Registered()
+	regsByIntf := map[reflect.Type]*codegen.Registration{}
+	for _, reg := range regs {
+		regsByIntf[reg.Iface] = reg
+	}
+
+	// Call Init and validate the registered fakes and generators.
+	//
+	// A simulator calls a workload's Init method many times, once per
```
spetrovic77, line 259: I presume we create new workload instances and call Init once per instance?

Do you know how many workload instances we plan to create, and how much the repeated validation will contribute to the cost compared to actual simulation runs? It feels like the API is slightly broken if we have to do this type of caching re: registrations. If anything, I would err on the side of repeated work.

internal/sim/api.go
```diff
@@ -249,36 +306,84 @@ func validateWorkload(t reflect.Type) error {
...
+	w          reflect.Type                           // workload type
+	regsByIntf map[reflect.Type]*codegen.Registration // registered components, by interface
+	fakes      map[reflect.Type]struct{}              // registered fakes, by interface
+	generators map[string]struct{}                    // registered generators, by method name
 }
 
 // RegisterFake implements the Registrar interface.
-func (r *registrar) RegisterFake(weavertest.FakeComponent) {
-	// TODO(mwhittaker): Implement.
+func (r *validatingRegistrar) RegisterFake(fake FakeComponent) {
```
spetrovic77, line 355: I presume you had to split it this way because of cached validation?

spetrovic77 (approved):
Looks really good!

internal/sim/api.go
mwhittaker: Yeah. I merged the two registrar implementations now.

mwhittaker (commented):
Thanks Srdjan! Code is much better now with your suggestion.

PR #591, opened 2023-09-08
Sweep over hyperparameters, running multiple sims.

mwhittaker:
Recall the `Simulator.Run` method:

```go
func (s *Simulator) Run(duration time.Duration) Results
```

Before this PR, `Run` ran a single simulation. This PR enhances the simulator to continuously run simulations for the provided duration.

Recall that the behavior of a simulation is governed by a number of hyperparameters like the RNG seed, number of operations to run, the probability of injecting a failure, and so on. Given a time budget to run simulations, how do we explore this hyperparameter space?

That's a *very* hard and open ended question. This PR implements a very stupid yet simple algorithm that tries to search "simpler" simulations before exploring more complex simulations. This is only a temporary algorithm to get things working. In the future, we'll have to revisit and improve it. The algorithm will also get more complex when we implement minimization.

This PR also changes the signature of `Run` to not return an error. A simulation itself should never fail to execute properly.

### DIFF
--- a/internal/sim/api.go
+++ b/internal/sim/api.go
@@ -218,10 +218,12 @@ var _ Event = Return{}
 var _ Event = DeliverReturn{}
 var _ Event = DeliverError{}
 
-// Results are the results of running a simulation.
+// Results are the results of simulating a workload.
 type Results struct {
-	Err     error   // first non-nil error returned by an op
-	History []Event // a history of all simulation events
+	Err            error         // first non-nil error returned by an op
+	History        []Event       // a history of the error inducing run, if Err is not nil
+	NumSimulations int           // number of simulations ran
+	Duration       time.Duration // duration of simulations
 }
 
 // New returns a new Simulator that simulates workload W.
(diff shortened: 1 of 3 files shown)


internal/sim/api.go
```diff
@@ -306,20 +308,56 @@ func validateWorkload(t reflect.Type) error {
...
 	defer cancel()
-	opts := options{
-		Seed:        time.Now().UnixNano(),
-		NumOps:      10,
-		NumReplicas: 1,
-		FailureRate: 0.1,
-		YieldRate:   0.25,
+
+	// TODO(mwhittaker): Use a smarter algorithm to sweep over hyperparameters.
+	// TODO(mwhittaker): Run simulations multiple goroutines.
```
spetrovic77, line 318: nit: in

internal/sim/api_test.go
```diff
@@ -72,17 +73,42 @@ func (d *divModWorkload) Mod(ctx context.Context, x, y int) error {
...
-		t.Fatal(err)
-	}
+	r := s.Run(2 * time.Second)
+	t.Log(r.Summary())
 	if r.Err != nil {
 		t.Fatal(r.Err)
 	}
 }
 
+type divideByZeroWorkload struct {
```
spetrovic77, line 85: nice!

spetrovic77 (approved):

internal/sim/api.go
mwhittaker: Fixed.

mwhittaker (commented):
Thanks Srdjan!

PR #593, opened 2023-09-08
Remove sim.WorkloadPointer interface.

mwhittaker:
This PR removes the `WorkloadPointer` interface, as suggested by Srdjan in https://github.com/ServiceWeaver/weaver/pull/588. The `New` function now has the following signature:

```go
func New(t *testing.T, x Workload, opts Options) *Simulator
```

This PR also addresses a subtlety in how Go defines method sets. Consider the following workload:

```go
type myWorkload struct{}
func (myWorkload) Init(sim.Registrar) error { ... }
func (myWorkload) Foo(context.Context) error { ... }
func (*myWorkload) Bar(context.Context) error { ... }
```

`myWorkload` implements the `Workload` interface and is a valid workload. Thus, a simulator should call all of `myWorkload`'s methods. But, which methods are defined on `myWorkload`? The answer might surprise you!

According to [the Go spec][spec], the method set of `myWorkload` includes `Init` and `Foo` but not `Bar` because `Bar` has a pointer receiver. `*myWorkload`, on the other hand, has `Init`, `Foo`, and `Bar` in its method set.

Thus, if we're being careful, we can't just call the methods defined on the workload type. We may have to take a pointer to the type first. This PR implements that.

This PR also tweaks a flaky test.

[spec]: https://go.dev/ref/spec#Method_sets

### DIFF
--- a/internal/sim/api.go
+++ b/internal/sim/api.go
@@ -119,12 +119,6 @@ type Workload interface {
 	Init(Registrar) error
 }
 
-// WorkloadPointer[T] is a *T that implements the Workload interface.
-type WorkloadPointer[T any] interface {
-	*T
-	Workload
-}
-
 // Options configure a Simulator.
 type Options struct {
 	Config string // TOML config contents
@@ -226,8 +220,8 @@ type Results struct {
 	Duration       time.Duration // duration of simulations
 }
 
-// New returns a new Simulator that simulates workload W.
-func New[W any, P WorkloadPointer[W]](t *testing.T, opts Options) *Simulator {
+// New returns a new Simulator that simulates the provided workload.
+func New(t *testing.T, x Workload, opts Options) *Simulator {
 	// Note that because the Init method on a workload struct T often has a
 	// pointer receiver *T, it is the pointer *T (rather than T itself) that
 	// implements the Workload interface.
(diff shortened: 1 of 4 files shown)


internal/sim/api.go
```diff
@@ -243,8 +237,26 @@ func New[W any, P WorkloadPointer[W]](t *testing.T, opts Options) *Simulator {
...
+	//
+	//     type t struct{}
+	//     func (t) ValueReceiver() {}
+	//     func (*t) PointerReceiver() {}
+	//
+	// According to the Go spec, the method set of t includes only
+	// ValueReceiver, while the method set of *t includes ValueReceiver and
+	// PointerReceiver [1]. We want to call *every* exported method on a
+	// workload struct, so we need to massage the type of x into a pointer if
+	// it isn't already.
```
spetrovic77, line 251: Very nice!

internal/sim/api_test.go
```diff
@@ -74,7 +74,7 @@ func (d *divModWorkload) Mod(ctx context.Context, x, y int) error {
 }
 
 func TestPassingSimulation(t *testing.T) {
-	s := New[divModWorkload](t, Options{})
+	s := New(t, &divModWorkload{}, Options{})
```
spetrovic77, line 77: Do you have a test here that tests the value receiver?

spetrovic77 (approved):

internal/sim/api_test.go
mwhittaker: Oops, added one.

mwhittaker (commented):
Thanks Srdjan!

PR #595, opened 2023-09-11
Run simulations concurrently.

mwhittaker:
This PR changes the `Simulator.Run` method to run multiple simulations in parallel. This does speed things up a bit (5x throughput for TestPassingSimulations), but things still aren't as fast as I'd like. I also need a lot of parallelism to speed things up fully which I don't yet fully understand. I'll try to do some profiling and optimizations in future PRs.

### DIFF
--- a/internal/sim/api.go
+++ b/internal/sim/api.go
@@ -323,16 +323,19 @@ func (s *Simulator) Run(duration time.Duration) Results {
 	defer cancel()
 
 	// Spawn n concurrent simulators. The simulators read options from the opts
-	// channel and write errors and results to the errs and results channels.
+	// channel and write errors and failed results to the errs and
+	// failedResults channels. Simulation ends when we encounter an error or
+	// successfully find an invariant violation.
 	//
 	// TODO(mwhittaker): Optimize things and pick a smarter value of n.
-	n := 1000
+	const n = 1000
 	opts := make(chan options, n)
 	errs := make(chan error, n)
-	results := make(chan Results, n)
+	failedResults := make(chan Results, n)
 	done := sync.WaitGroup{}
 	numSimulations := int64(0)
 
+	// Spawn n simulating goroutines that read from the opts channel.
 	done.Add(n)
 	for i := 0; i < n; i++ {
 		go func() {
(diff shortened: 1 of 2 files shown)


internal/sim/api.go
```diff
@@ -313,50 +315,81 @@ func validateWorkload(t reflect.Type) error {
...
-	// TODO(mwhittaker): Run simulations in multiple goroutines.
-	// TODO(mwhittaker): Read and run graveyard entries.
+	// Spawn n concurrent simulators. The simulators read options from the opts
+	// channel and write errors and results to the errs and results channels.
+	//
+	// TODO(mwhittaker): Optimize things and pick a smarter value of n.
+	n := 1000
+	opts := make(chan options, n)
+	errs := make(chan error, n)
+	results := make(chan Results, n)
```
spetrovic77, line 332: Rename errs and results to describe what type of errors they store.

internal/sim/api.go
```diff
@@ -313,50 +315,81 @@ func validateWorkload(t reflect.Type) error {
...
 						NumOps:      numOps,
 						NumReplicas: 1,
 						FailureRate: failureRate,
 						YieldRate:   yieldRate,
 					}
-					results, err := s.runOne(ctx, opts)
-					if err != nil && err == ctx.Err() {
+
+					select {
+					case <-ctx.Done():
```
spetrovic77, line 378: This processing is confusing, perhaps better to run it in a separate goroutine?

internal/sim/api.go
```diff
@@ -313,50 +315,81 @@ func validateWorkload(t reflect.Type) error {
...
 	defer cancel()
 
-	// TODO(mwhittaker): Use a smarter algorithm to sweep over hyperparameters.
-	// TODO(mwhittaker): Run simulations in multiple goroutines.
-	// TODO(mwhittaker): Read and run graveyard entries.
+	// Spawn n concurrent simulators. The simulators read options from the opts
+	// channel and write errors and results to the errs and results channels.
+	//
+	// TODO(mwhittaker): Optimize things and pick a smarter value of n.
+	n := 1000
```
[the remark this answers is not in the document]
mwhittaker: Done.

internal/sim/api.go
mwhittaker: Done. I also added some more comments.

internal/sim/api.go
mwhittaker: Good idea. Done.

mwhittaker (commented):
Thanks Srdjan!

spetrovic77 (approved):
Looks great.

PR #610, opened 2023-09-13
Combine pipe.Cmd's WPipe and RPipe into a single method.

ghemawat:
Calls to WPipe and RPipe are always paired, so combine them into one method. The new method returns a struct so that we get good names for the four distinct values needed for the pair of pipes.


mwhittaker (approved):
Nice! I find ChildReader/ChildWriter and ParentReader/ParentWriter much easier to understand than the old way of toWeavelet/toEnvelope.

PR #622, opened 2023-09-18
Use the Go BuildInfo to read the tool and application module versions.

spetrovic77:
This remove the manual module version tracking (in the version.go file), which will hopefully lead to fewer versioning errors and allow us to track the version information at a finer granularity.

Note that the tool and the application module versions aren't ever directly compared: their values are only used for nicer error-message generation.


mwhittaker (approved):
Really nice change!

PR #624, opened 2023-09-19
Fix the website URL for the online boutique example.

spetrovic77:
We moved the Online Boutique code to a different repo, and the link has been broken ever since.


mwhittaker (approved):
Thanks Srdjan!

PR #626, opened 2023-09-19
Added some basic built-in generators to simulator.

mwhittaker:
This PR adds three built-in generators to the simulator. I'll add more in future PRs, but these three seemed like no-brainers. I'm open to alternative name suggestions too!


spetrovic77 (approved):
Nice!

PR #627, opened 2023-09-20
Added test that shows importance of minimization.

mwhittaker:
I've been playing around with various workloads and found an append-only string workload to be quite interesting. It emphasizes the importance of the following things:

1. Simulator speed is important. For rarely occurring bugs, the simulator being 100x slower can mean the difference between finding the bug overnight and not finding the bug in a month.
2. Test case minimization is more important for rarer bugs, as they are likely to arise in longer executions with irrelevant ops.
3. At any point in time, the search algorithm should give every possible execution a non-zero change of being run. Currently, we increase the number of ops per execution over time, but this means that if a bug only arises from an execution with few ops, we may never find it.

### DIFF
--- a/internal/sim/components.go
+++ b/internal/sim/components.go
@@ -52,6 +52,15 @@ type blocker interface {
 	Block(context.Context) error
 }
 
+// register is a string-valued register.
+type register interface {
+	// Append appends to the register.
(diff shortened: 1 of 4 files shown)


mwhittaker:
We chatted offline and decided to abandon this change.

PR #631, opened 2023-09-20
Catch panics in simulations.

mwhittaker:
This PR extends the simulator to catch any panics that happen during an execution and abort the execution. By catching panics, the simulator is able to show the user the history of events that led to the panic.

This PR also fixes a minor bug. Previously, we recorded the history for aborted events. With this PR, once an execution is aborted, no other events are recorded.

I also renamed some unit tests to be more descriptive.

### DIFF
--- a/internal/sim/executor_test.go
+++ b/internal/sim/executor_test.go
@@ -28,18 +28,18 @@ import (
 	"github.com/ServiceWeaver/weaver"
 )
 
-// See TestSuccessfulExecution.
-type successfulWorkload struct {
+// See TestPassingExecution.
+type passingWorkload struct {
 	divmod weaver.Ref[divMod]
 }
 
-func (s *successfulWorkload) Init(r Registrar) error {
+func (p *passingWorkload) Init(r Registrar) error {
 	r.RegisterGenerators("DivMod", intn{0, 100}, intn{1, 100})
 	return nil
 }
 
-func (s *successfulWorkload) DivMod(ctx context.Context, x, y int) error {
-	div, mod, err := s.divmod.Get().DivMod(ctx, x, y)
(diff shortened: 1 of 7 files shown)


internal/sim/executor_test.go
```diff
@@ -116,14 +116,70 @@ func TestExecuteGraveyardEntries(t *testing.T) {
...
+type panickingOpWorkload struct{}
+
+func (p *panickingOpWorkload) Init(r Registrar) error {
+	r.RegisterGenerators("Panic", Flip(0.1))
+	return nil
+}
+
+func (p *panickingOpWorkload) Panic(ctx context.Context, b bool) error {
+	if b {
+		panic("Panic!")
```
spetrovic77, line 152: The most stressful code review ever! 🤣

internal/sim/simulator.go
```diff
@@ -598,6 +598,9 @@ func (r *Results) Mermaid() string {
 		case EventDeliverError:
 			call := calls[x.SpanID]
 			fmt.Fprintf(&b, "    note right of %s%d: [%d:%d] RemoteCallError\n", call.Caller, call.Replica, x.TraceID, x.SpanID)
+		case EventPanic:
+			// TODO(mwhittaker): Show stack trace?
```
spetrovic77, line 602: I think we should.

spetrovic77 (approved):

internal/sim/executor_test.go
mwhittaker: :laughing:

internal/sim/simulator.go
mwhittaker: Done! The stack traces are a bit ugly for now, but we can improve them later.

83JDoBNZLuXJLGp

mwhittaker (commented):
Thanks Srdjan!

PR #632, opened 2023-09-20
Escape test names in simulator.

mwhittaker:


internal/sim/simulator.go
```diff
@@ -269,8 +269,15 @@ func (s *Simulator) newExecutor() *executor {
...
 func (s *Simulator) graveyardDir() string {
-	// TODO(mwhittaker): Escape test names.
-	return filepath.Join("testdata", "sim", s.t.Name())
+	// Test names often contain slashes ("/"). We replace "/" with "#" to
+	// safely use the test name as a directory name.
+	//
+	// TODO(mwhittaker): This mapping is sensitive to collisions. A test named
+	// "foo/bar" collides with a test named "foo#bar". I think in practice,
+	// this will likely not be a big issue. But, if people are running into
+	// problems, we can use a more complex collision resistant sanitization.
```
spetrovic77, line 278: I had to do this for GKE and best to avoid it for something as simple as this.

spetrovic77 (approved):

internal/sim/simulator.go
mwhittaker: Agreed!

PR #633, opened 2023-09-20
Detect bad context propagation in simulations.

mwhittaker:
An executor embeds trace and span ids in the contexts passed to a workload. The workload is responsible for propagating these contexts throughout the workload. If a context is not propagated, then an executor cannot tell which op a given call is part of. This makes it impossible for the executor to schedule things correctly. Plus, the reported history will be missing tracing information, making it much harder to understand what happened.

This PR detects and panics when contexts are not being propagated properly. I think with enough effort, we could modify the executor to gracefully handle missing trace information, but I'll leave that for future work.

### DIFF
--- a/internal/sim/executor.go
+++ b/internal/sim/executor.go
@@ -384,10 +384,17 @@ func (e *executor) call(caller string, replica int, reg *codegen.Registration, m
 		strings[i] = fmt.Sprint(arg)
 	}
 
+	// Extract the trace id.
+	traceID, _ := extractIDs(ctx)
+	if traceID == 0 {
+		// TODO(mwhittaker): Link to online documentation with better
+		// explanation of this error.
+		panic(fmt.Errorf("missing simulation trace context. Make sure you are correctly propagating contexts through your workload."))
+	}
+
 	// Record the call.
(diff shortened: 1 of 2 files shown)


internal/sim/executor.go
```diff
@@ -384,10 +384,17 @@ func (e *executor) call(caller string, replica int, reg *codegen.Registration, m
 		strings[i] = fmt.Sprint(arg)
 	}
 
+	// Extract the trace id.
+	traceID, _ := extractIDs(ctx)
+	if traceID == 0 {
+		// TODO(mwhittaker): Link to online documentation with better
+		// explanation of this error.
+		panic(fmt.Errorf("missing simulation trace context. Make sure you are correctly propagating contexts through your workload."))
```
spetrovic77, line 392: Can you be more explicit right now. In particular, there's something we require the user to do (e.g., propagate the contexts through to the component calls), might as well mention it briefly here?

spetrovic77 (approved):

internal/sim/executor.go
mwhittaker: Reworded to the error to hopefully be clearer.

PR #634, opened 2023-09-20
net/call now picks appropriate default for options.

ghemawat:
Previously callers of call.Serve and call.Connect were responsible for picking appropriate tuning parameter values. We now let the callers provide zero values and let the call implementation pick the right value.

### DIFF
--- a/internal/net/benchmarks/net_test.go
+++ b/internal/net/benchmarks/net_test.go
@@ -39,8 +39,8 @@ var (
 	subproccess           = flag.Bool("subprocess", false, "Is this a subprocess?")
 	network               = flag.String("network", "", "Network (e.g., tcp, unix, mtls)")
 	address               = flag.String("address", "", "Server address")
-	inlineHandlerDuration = flag.Duration("inline_handler_duration", 0, "ServerOptions.InlineHandlerDuration")
-	writeFlattenLimit     = flag.Int("write_flatten_limit", 0, "ServerOptions.WriteFlattenLimit")
+	inlineHandlerDuration = flag.Duration("inline_handler_duration", -1, "ServerOptions.InlineHandlerDuration")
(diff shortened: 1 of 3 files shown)


internal/net/benchmarks/net_test.go
```diff
@@ -39,8 +39,8 @@ var (
 	subproccess           = flag.Bool("subprocess", false, "Is this a subprocess?")
 	network               = flag.String("network", "", "Network (e.g., tcp, unix, mtls)")
 	address               = flag.String("address", "", "Server address")
-	inlineHandlerDuration = flag.Duration("inline_handler_duration", 0, "ServerOptions.InlineHandlerDuration")
-	writeFlattenLimit     = flag.Int("write_flatten_limit", 0, "ServerOptions.WriteFlattenLimit")
+	inlineHandlerDuration = flag.Duration("inline_handler_duration", -1, "ServerOptions.InlineHandlerDuration")
```
mwhittaker, line 42: Just a clarifying question, an `inlineHandlerDuration` of `-1` means don't run handlers inline, and a `writeFlattenLimit` of `-1` means don't flatten?

mwhittaker (approved):
Nice!

internal/net/benchmarks/net_test.go
ghemawat: Yes. This is implied, but a bit indirectly by the comments on the fields in options.go. I extended the comments to say this explicitly:

```
	// If positive, calls on the server are inlined and a new goroutine is
	// launched only if the call takes longer than the provided duration.
	// If zero, the system inlines call execution and automatically picks a
	// reasonable delay before the new goroutine is launched.
	// If negative, handlers are always started in a new goroutine.
```

```
	// All writes smaller than this limit are flattened into a single
	// buffer before being written on the connection. If zero, an appropriate
	// value is picked automatically. If negative, no flattening is done.
```

PR #635, opened 2023-09-20
Run more than one replica in the simulator.

mwhittaker:
Executions with more than one replica of each component are harder to reason about, but some bugs may only trigger when there are more than one replica. This PR changes the simulator to sometimes run 2 or 3 replicas of every component.

### DIFF
--- a/internal/sim/simulator.go
+++ b/internal/sim/simulator.go
@@ -404,21 +404,10 @@ func (s *Simulator) run(ctx context.Context, stats *stats) (result, error) {
 		defer done.Done()
 		seed := time.Now().UnixNano()
 		for numOps := 1; ; numOps++ {
-			for numReplicas := 1; numReplicas <= 3; numReplicas++ {
+			for _, numReplicas := range []int{1, 2, 3} {
 				for _, failureRate := range []float64{0.0, 0.01, 0.05, 0.1} {
 					for _, yieldRate := range []float64{0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0} {
-						// Run fewer executions with more replicas.
-						var iterations int
-						switch numReplicas {
-						case 1:
-							iterations = 1000
-						case 2:
-							iterations = 100
-						case 3:
-							iterations = 10
-						}
-
-						for i := 0; i < iterations; i++ {
+						for i := 0; i < 1000; i++ {
 							seed++
 							p := hyperparameters{
 								Seed:        seed,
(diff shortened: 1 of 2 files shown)


internal/sim/simulator.go
```diff
@@ -404,21 +404,34 @@ func (s *Simulator) run(ctx context.Context, stats *stats) (result, error) {
...
-							NumReplicas: 1,
-							FailureRate: failureRate,
-							YieldRate:   yieldRate,
+			for numReplicas := 1; numReplicas <= 3; numReplicas++ {
+				for _, failureRate := range []float64{0.0, 0.01, 0.05, 0.1} {
+					for _, yieldRate := range []float64{0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0} {
+						// Run fewer executions with more replicas.
+						var iterations int
+						switch numReplicas {
+						case 1:
```
spetrovic77, line 413: I was surprised to see orders of magnitude drops like these. The execution runs at the same speed methinks regardless of the number of replicas? (Can we confirm this via a benchmark?)

It would be clearer to keep another tightly nested loop at the top:

```
for _, numReplicas := range []int{1, 2, 3} {
...
}
```

You can use 1000 for all of these. It's an arbitrary value anyway.

spetrovic77 (approved):

internal/sim/simulator.go
mwhittaker: For every execution, an executor has to create and initialize every replica (e.g., fill its refs, listeners, and config). So, executions do slow down with more replicas. I added a benchmark to demonstrate: 

```
name                        time/op
NumReplicas/1-Replicas-128  29.1µs ± 3%
NumReplicas/2-Replicas-128  33.1µs ± 1%
NumReplicas/3-Replicas-128  37.0µs ± 1%

name                        alloc/op
NumReplicas/1-Replicas-128  2.83kB ± 0%
NumReplicas/2-Replicas-128  3.54kB ± 0%
NumReplicas/3-Replicas-128  4.25kB ± 0%

name                        allocs/op
NumReplicas/1-Replicas-128    78.0 ± 0%
NumReplicas/2-Replicas-128    93.0 ± 0%
NumReplicas/3-Replicas-128     108 ± 0%
```

But, I didn't vary the number of iterations for performance reasons. Instead, I had an intuition that most bugs don't need more than one replica to surface. And, executions with multiple replicas are much harder to understand. So, I wanted to bias towards fewer replicas to increase the chances of returning a simpler failing execution. 

But I agree, this is all too complicated, so I switched to the simple for loop you suggested! :smile:

mwhittaker (commented):
Thanks Srdjan!

PR #636, opened 2023-09-21
Added more generators to simulator.

mwhittaker:


internal/sim/generators.go
```diff
@@ -43,6 +35,109 @@ func Flip(p float64) Generator[bool] {
...
 }
 
+// Numerics
+
+// Nat returns a [Generator] that returns non-negative numbers, also known as
+// natural numbers. Note that Nat does not return all natural numbers
+// equiprobably. Instead, it biases towards numbers closer to zero and other
+// pathological numbers that are more likely to induce bugs (e.g.,
+// math.MaxInt).
+func Nat() Generator[int] {
```
spetrovic77, line 45: Can we rename `Nat` -> `UInt`?

NAT is overloaded and when doing the code review above, I really had no idea what Nat was.

internal/sim/generators.go
```diff
@@ -43,6 +35,109 @@ func Flip(p float64) Generator[bool] {
...
+	}
+	return Weight(choices)
+}
+
+// Int returns a [Generator] that returns integers. Note that Int does not
+// return all integers equiprobably. Instead, it biases towards numbers closer
+// to zero and other pathological numbers that are more likely to induce bugs
+// (e.g., math.MaxInt, math.MinInt).
+func Int() Generator[int] {
+	choices := []Weighted[int]{
```
spetrovic77, line 71: This `Weighted` type is brilliant!

internal/sim/generators.go
```diff
@@ -43,6 +35,109 @@ func Flip(p float64) Generator[bool] {
...
+		const stdev = 1e6
+		return r.NormFloat64() * stdev
+	})
+}
+
+// Rune returns a [Generator] that returns runes equiprobably.
+func Rune() Generator[rune] {
+	return generatorFunc[rune](func(r *rand.Rand) rune {
+		// Note that rune is an alias for int32.
+		return rune(r.Intn(math.MaxInt32 + 1))
```
spetrovic77, line 128: This random rune thing would break our code left and right! 🤣

internal/sim/generators.go
```diff
@@ -55,6 +150,148 @@ func Range(low, high int) Generator[int] {
...
+// String returns a [Generator] that returns moderately sized readable strings,
+// with a bias towards smaller strings.
+func String() Generator[string] {
+	// Prefer smaller strings.
+	size := Weight([]Weighted[int]{
+		{100, Range(0, 10)},
+		{10, Range(10, 100)},
+		{1, Range(100, 4096)},
+	})
+	alphabet := OneOf(
```
spetrovic77, line 164: Do you want to throw in some non-English alphabet characters in there? Or English character-like runes that take up two bytes?

spetrovic77 (approved):
Very nice!!

internal/sim/generators.go
mwhittaker: Yeah, it's very useful! I can't take credit for the idea. I'm mostly just stealing from OCaml's quickcheck library: https://ocaml.org/p/qcheck-core/latest/doc/QCheck/Gen/index.html.

internal/sim/generators.go
mwhittaker: :rofl:

internal/sim/generators.go
mwhittaker: I find `UInt` a bit confusing since we're generating `int`s and not `uint`s. I could change the type to return `uint`s but I also found that a bit annoying because I feel like people work with `int`s way more than they do `uint`s. 

Do you like any of these names?

- `Natural`
- `NaturalNumber`
- `NonNegative`
- `NonNeg`

internal/sim/generators.go
mwhittaker: Good idea. Added lowercase Greek letters, which are two bytes each in UTF-8.

mwhittaker (commented):
Thanks Srdjan!

internal/sim/generators.go
mwhittaker: We chatted offline and decided on `NonNegativeInt`.

PR #638, opened 2023-09-21
Add a graph library and use it to represent the component call-graph.

spetrovic77:
This graph library is based on an earlier C++ implementation by ghemawat@.

The binary package now returns a component call-graph, as opposed to returning a set of component edges. This fixes the following bug we ran into multiple times:
  * The program contains only the `weaver.Main` component.
  * The set of component edges is therefore empty.
  * The deployer iterates over the edges, and ends up ignoring `weaver.Main` component.

The graph package will be handy in the future, as we plan to perform a Topo-sort traversal of the graph when generating Kubernetes YAML files. Also, the package may be useful for detecting component-graph cycles.

### DIFF
--- a/godeps.txt
+++ b/godeps.txt
@@ -363,11 +363,9 @@ github.com/ServiceWeaver/weaver/internal/tool
     runtime/debug
 github.com/ServiceWeaver/weaver/internal/tool/callgraph
     fmt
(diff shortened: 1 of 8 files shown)


ghemawat (approved):
Well this looks familiar :-)

PR #639, opened 2023-09-21
Added local MySQL instructions for chat app.

mwhittaker:
This PR adds some instructions on how to run a MySQL instance locally using Docker for the example `chat` application. This is part of a larger effort to understand how hard it is for people to run their own databases locally.

In a future PR, I will explore how we can run MySQL instances for unit tests. [dockertest][] and [testcontainers] seem to be popular choices. After that, I will likely remove our use of sqlite entirely.

[dockertest]: https://github.com/ory/dockertest
[testcontainers]: https://testcontainers.com/

### DIFF
--- a/examples/chat/README.md
+++ b/examples/chat/README.md
@@ -19,8 +19,39 @@ graph TD
 
 ## How to Run Locally
 
-```sh
-$ go run .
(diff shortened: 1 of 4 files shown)


spetrovic77 (approved):
Nice!

PR #641, opened 2023-09-25
Use docker containers to test the chat app.

mwhittaker:
Recall that the chat app implements a simple online messaging forum. It persists chat threads in a database. Before this PR, the chat app used MySQL for production deployments and used SQLite for local deployments and tests.

This PR removes the use of SQLite and instead always uses MySQL, even locally and in tests. This has two advantages:

1. Running the same database in production and in tests helps discover more bugs. In fact, in this PR, I found a bug in the code by running MySQL in tests.
2. It's not always possible to run an app against SQLite. An app might rely on features SQLite does not offer, for example.

Always using MySQL does have some downsides. Launching a docker container for every test is slow. The chat tests now take 24 seconds. Most of this time comes from spinning up MySQL, not from launching the actual container. We could put in some effort to re-use a MySQL instance across multiple tests, though it would complicate things.

Overall, I think faking is still very important. `chat/server_test.go` uses a `sqlStore` fake, for example. I think this will be especially necessary for simulation testing.

## Details

- To run MySQL in tests, we leverage [testcontainers][], a Go module that makes it easy to run containers in tests.
- I renamed the chat database from `serviceweaver_chat_example` to just `chat`.
- I fixed a bug in `sqlStore` where posts were not being ordered correctly.

[testcontainers]: https://testcontainers.com/


spetrovic77 (approved):
Looks very clean!

PR #643, opened 2023-09-27
Add signal handler cleanup registration.

ghemawat:
Support registering functions that are called when a signal that will terminate the process is received. Use this facility to remove temporary directory and registry entry when a multi deployer application is interrupted.


mwhittaker (approved):
Very nice! :signal_strength: :satellite:

PR #644, opened 2023-09-27
Avoid 'weaver multi' zombie main components.

mwhittaker:
Recall that when you run `weaver multi deploy`, you start a `weaver multi deploy` process that runs a bunch of remote weavelets in subprocesses. The weavelets communicate with the `weaver multi deploy` process over a set of pipes. Here's an example of what `top` shows when deploying the collatz app:

```
-bash
 └─ cmd/weaver/weaver multi deploy weaver.toml
    ├─ examples/collatz/collatz
    ├─ examples/collatz/collatz
    ├─ examples/collatz/collatz
    ├─ examples/collatz/collatz
    ├─ examples/collatz/collatz
    └─ examples/collatz/collatz
```

If you `kill` the `weaver multi deploy` process, it doesn't kill the child weavelet processes, at least not directly. However, killing the `weaver multi deploy` process closes the pipes to the weavelets. This should cause the weavelet processes to exit.

This was working correctly for all weavelets not running the main component. weavelets running the main component blocked on the user provided lambda. If this lambda did something like run an HTTP server, then the process wouldn't exit, even if the underlying weavelet was broken. This led to a bunch of zombie processes lingering around forever.

This PR fixes the bug by exiting whenever the weavelet encounters a broken pipe.

Also note that `examples/examples_test.go` runs into this zombie behavior. So, if you've been running `./dev/build_and_test.sh`, you might to run `ps aux | grep weaver` to see if you have a bunch of zombies.

### DIFF
--- a/weaver.go
+++ b/weaver.go
@@ -163,18 +163,18 @@ func runLocal[T any, _ PointerToMain[T]](ctx context.Context, app func(context.C
 		return err
 	}
 
-	runner, err := weaver.NewSingleWeavelet(ctx, regs, opts)
+	wlet, err := weaver.NewSingleWeavelet(ctx, regs, opts)
 	if err != nil {
 		return err
 	}
 
 	go func() {
-		if err := runner.ServeStatus(ctx); err != nil {
+		if err := wlet.ServeStatus(ctx); err != nil {
 			fmt.Fprintln(os.Stderr, err)
 		}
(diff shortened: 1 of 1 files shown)


spetrovic77 (approved):
Nice fix, I indeed had a bunch of zombies on my local machine!

PR #646, opened 2023-09-28
Move Local GKE under the GKE deployer

rgrandl:
Local GKE is a version of the GKE deployer that allows you to emulate running applications in GKE, but on the local machine.

Local GKE is useful:
* for the Service Weaver team to quickly iterate over the GKE deployer (e.g., debugging, new functionalities)
* for a user that wants to use the GKE deployer but wants to see how it works when running on the local machine.

Given that the local GKE deployer makes sense only in correlation with the GKE deployer, it makes more sense to add it under the GKE deployer in our documentation instead of a standalone deployer.


spetrovic77 (approved):
Looks good. This will leave more room for `weaver-kube`.

PR #647, opened 2023-10-01
Controlling application logging verbosity

zailic:
This PR introduces the ability to set the application logging verbosity by a configurable value, like in the example bellow:
```
[serviceweaver]
log_level = debug
...
```


mwhittaker:
Hi @zailic, thanks for the PR!

I chatted with the team, and we're unsure if this is the best approach because

- we want to keep the number of configuration options as small as possible,
- there is a way to change log level currently that doesn't require changes to Service Weaver, and
- there are other ways someone might modify their loggers (besides changing the log level) that this approach doesn't cover. 

Here are two alternatives. Let me know what you think. We're not sure ourselves what the best approach is.

## Option 1: Let Users Modify Loggers 

We discussed this option in the Discord, but I'll repeat it here for others to see as well. Currently, a developer writing a Service Weaver application can modify their loggers using the `slog` package. For example, here's some code for a `WithLevel` function that modifies the log level of a logger: 

```go
type leveled struct {
        level   slog.Level
        handler slog.Handler
}

func (l leveled) Enabled(ctx context.Context, level slog.Level) bool {
        return level >= l.level
}

func (l leveled) Handle(ctx context.Context, record slog.Record) error {
        return l.handler.Handle(ctx, record)
}

func (l leveled) WithAttrs(attrs []slog.Attr) slog.Handler {
        return leveled{l.level, l.handler.WithAttrs(attrs)}
}

func (l leveled) WithGroup(name string) slog.Handler {
        return leveled{l.level, l.handler.WithGroup(name)}
}

func WithLevel(logger *slog.Logger, level slog.Level) *slog.Logger {
        return slog.New(leveled{level, logger.Handler()})
}
```

The good thing about this approach is that developers can modify their loggers in arbitrary ways, and we don't have to involve Service Weaver at all. 

The bad thing, as you mentioned in Discord, is that this approach can be inconvenient. For example, if an app has 100 components and we have to pass a log level to all 100, that is cumbersome. This might be solved by passing log level as an environment variable. Or, maybe we could explore adding a global config section to Service Weaver config files. Even still, it is inconvenient to modify your loggers in a bunch of different places.

## Option 2: A Generic Logger API 

If we do decide to involve Service Weaver in modifying loggers, then we may want to provide a more generic API that developers can use beyond just setting log level. I'm not exactly sure what that would look like, but to give a concrete example, maybe `weaver.Run` takes a default `slog.Handler`:

```go
func main() {
    if err := weaver.Run(context.Background(), someDefaultHandler(), serve); err != nil {
        ...
    }
}
```

The benefit of this approach is that you can modify your loggers in a single place, and hopefully, the API is also generic enough to support relatively arbitrary customization. The disadvantage is that it introduces a new API to Service Weaver, which comes with some new complexity.

zailic:
hi @mwhittaker,

Thank you for your thoughtful feedback and suggestions.
I completely understand the expressed concerns about this approach and analyzed the pros and cons of the suggested alternatives. It's clear that there's no _one-size-fits-all_ solution. I'll look forward to see where this discussion leads us. Meanwhile we may consider also a third option:

## Option 3: A package-global base logger
where the user has the option to customize it at the "up-stack" level before `weaver.Run` call:

_weaver logging package:_
```go
var (
    Log = slog.New(someServiceWeaverDefaultLogHander())
)
```
_user weaver app:_
```go
func main() {
    wlogging.SetLogger(someUserCustomizedLogHandler())
    if err:=weaver.Run(ctx, serve); err != nil {
        ...
    }
}
```

Pros:
- the possibility to configure the logger in a single place
- no need to pass the logger around
- minimized ServiceWeaver involvement.

Cons:
- may lead to global state, which can make testing and debugging more complex.

rgrandl:

PR #666, opened 2023-10-19
Fixed panicking packages.Load call.

mwhittaker:
`weaver generate`'s initial call to `packages.Load` started to panic. I'm not entirely sure why, but upgrading the `golang.org/x/tools` module fixed the problem.


spetrovic77 (approved):
Thanks for fixing!

PR #667, opened 2023-10-21
Colocation groups config

rgrandl:
Right now we specify the components that should run together in a colocation group. For all the others we run them in their own groups. However, given that the group is the unit of deployment, we'll have to add configuration knobs that make sense per colocation group.

This PR provides some options to set per colocation group knobs in the config.


rgrandl:
We never reached a conclusion on what is the right way to specify collocation groups. The way we let users specify collocation groups as of now seems fine.

PR #677, opened 2023-11-13
Add goreleaser to publish signed binaries

veggiemonk:
Hi team,

I just wanted to introduce goreleaser as it simplifies the publishing, distribution and signing of assets.
You can see an example here: https://github.com/anchore/grype/releases/tag/v0.73.1

The next step would be to provide a container with weaver in it as it could be used as a base image for container instead of a [vanilla container like ubuntu](https://github.com/ServiceWeaver/weaver-kube/blob/main/internal/impl/docker.go#L140)

Happy to discuss this further as **this PR probably won't work on the first try**. Those things are a bit hard to get right without testing them first.

Let me know your thoughts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,3 +1,17 @@
+# Copyright 2023 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 name: release
 
 on:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+name: release
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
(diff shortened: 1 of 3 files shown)


mwhittaker:
Hi @veggiemonk, what benefits does gorelease offer in comparison to tagging and releasing manually with something like `go tag -a v0.X.Y && git push origin v0.X.Y`?

veggiemonk:
Hi @mwhittaker, it doesn't change the workflow of releasing with `git tag -a v0.X.Y && git push origin v0.X.Y`. 

When a new tag is created (see https://github.com/veggiemonk/weaver/blob/ef19a2c09677f3c6e42e04e5e91918cd71cdbc42/.github/workflows/release.yml#L19-L21), it will automatically build and push signed binaries that will be available with the release on GitHub as shown here https://github.com/anchore/grype/releases/tag/v0.73.1 .

It is also possible to publish a homebrew tap (brew.sh, famous package manager on Mac) but this is not setup in this PR. 

The next step would be to provide a container image with weaver in it. It can serve as a base image for application container and remove the need for compiling weaver in the container build step and allows for `ko` to build weaver applications without ever writing a Dockerfile, therefore reducing the number of steps to deploy. 

I think GitHub provides some kind of metrics about the number of downloads if the container is hosted on ghcr.io (GitHub Container Registry). Maybe you want to host that on GCR/Artefacts Registry? 

Goreleaser can use `ko` to build and deploy a Go application in a container. Therefore, the migration a project that already uses Goreleaser to include service weaver would be seamless. 

`ko` produces checksums and SBOMs for applications. Supply chain security is something that needs to be setup at the beginning of a project. It is not impossible to add later but it could become a chore.  

This has been my thinking so far. I think a full example would be useful to show the possibilities and the tradeoffs.

PR #682, opened 2023-11-15
Otel tracer: update service name to the app name

rgrandl:
In the current implementation, we create a global trace provider where the service name is the weavelet id. However, this is meaningless and confusing especially if you use Jaeger where the services to select traces for are based on the weavelet ids.

With this PR, the user sees only one service, which is the app name, and all the traces associated with it.

In an ideal scenarion, we may want to export a tracing service for each colocation group. However, this is not easy, given that we have a global trace provider that has to set attributed for the service name. We may revisit this in a future PR.

### DIFF
--- a/internal/weaver/tracing.go
+++ b/internal/weaver/tracing.go
@@ -15,7 +15,6 @@
 package weaver
 
 import (
-	"fmt"
 	"os"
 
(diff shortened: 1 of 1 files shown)


mwhittaker (approved):
Nice, thanks Robert!

PR #684, opened 2023-11-25
Update docs.md to mention encrypted secure conn via mTLS

liufuyang:
Allow users to easily find information related to encrypted and secure connection between components.


mwhittaker (approved):
Thanks!

PR #687, opened 2023-12-18
Drop UpdateComponentsRPC.

ghemawat:
An earlier changed switched to using a controller component method call to update the set of components served by a weavelet. This change finishes that by dropping the UpdateComponentsRPC from the pipe protocol.

The weavertest multi deployer code had to be refactored a little bit to give it access to the controller component (either local for components hosted in the main process, or remote for components hosted in other processes).

Also dropped the GetHealth method from the controller component since it is currently not used. (The only user of GetHealth is an out of tree deployer that has not yet been switched to using the controller component.)


mwhittaker (approved):
Nice!

PR #691, opened 2023-12-19
Make profile collection a component method.

ghemawat:
Also, stop tracking whether or not CPU profiling is already on explicitly; let the Go runtime do it.


mwhittaker (approved):
Nice!

PR #701, opened 2024-01-03
Move GetMetrics from a pipe RPC to a controller component method.

ghemawat:


internal/envelope/conn/weavelet_conn.go
```diff
@@ -145,21 +143,6 @@ func (w *WeaveletConn) handleMessage(handler WeaveletHandler, msg *protos.Envelo
...
-	case msg.GetMetricsRequest != nil:
-		// Inject Service Weaver specific labels.
-		update := w.metrics.Export()
-		for _, def := range update.Defs {
-			if def.Labels == nil {
-				def.Labels = map[string]string{}
-			}
-			def.Labels["serviceweaver_app"] = w.einfo.App
-			def.Labels["serviceweaver_version"] = w.einfo.DeploymentId
-			def.Labels["serviceweaver_node"] = w.einfo.Id
```
mwhittaker, line 157: Do we need to inject these labels in `RemoteWeavelet`'s implementation of `GetMetrics`?

mwhittaker (approved):
Thanks Sanjay! Looks great.

internal/envelope/conn/weavelet_conn.go
ghemawat: Good catch. Done.

PR #704, opened 2024-01-08
Rename control components.

ghemawat:
1. Rename the control.Controller component to control.WeaverControl (to distinguish it from the DeployerControl component). Consistently rename corresponding types in other packages.
2. Introduce a control.DeployerControl component that will hold methods implemented by deployers. Remove the logger component and move LogBatch into control.DeployerControl.
3. Do not make the multi deployer provide a full component implementation (by using "weaver generate"). Instead, just export its override under the non-overridden component name.


mwhittaker (approved):
Nice!

PR #705, opened 2024-01-08
Move ActivateComponents RPC from pipe to component method call.

ghemawat:
Continuing the work to remove pipe-based RPCs, the ActivateComponents call to the deployer is now a component method call.

Also made NewEnvelope() responsible for serving calls to the deployer component. (Callers that are directly using NewEnvelopeConn() still have to manage such calls themselves.)


mwhittaker (approved):
Awesome!

PR #707, opened 2024-01-09
Fix MTLS handling.

ghemawat:
We were using the wrong component name when setting up group data structures for MTLS (because of Go's funny loop variable capture semantics). Fixed this.

Added a test that uses multi deployer with MTLS.

Increased timeout for the test from 4s to 10s.


spetrovic77 (approved):
Thanks for the fix!

PR #708, opened 2024-01-09
Switch certificate RPCs from pipe to component calls

ghemawat:


examples/examples_test.go
```diff
@@ -113,22 +113,24 @@ func TestExamples(t *testing.T) {
...
 			// "weaver multi deploy" the application with MTLS.
-			t.Run("weaver-multi-mtls", func(t *testing.T) {
-				if os.Getenv("GITHUB_RUN_ID") != "" {
-					t.Skip("test takes too long (over 30s) on github")
-				}
-				cmd := startCmd(ctx, t, nil, "../../cmd/weaver/weaver", "multi", "deploy", "weaver_mtls.toml")
-				t.Cleanup(terminateCmdAndWait(t, cmd))
-				run(t, test)
-			})
+			if name == "collatz" {
```
mwhittaker, line 116: Are we only running mTLS on collatz to speed up the tests?

mwhittaker (approved):
Awesome! Happy to see so much code getting deleted!

examples/examples_test.go
ghemawat: Yes, it gets really slow otherwise

PR #711, opened 2024-01-11
Make tests use NewEnvelope instead of NewEnvelopeConn.

ghemawat:
Introduce a new Child interface that NewEnvelope uses to run the weavelet. Typically this is a real process, but tests that want to run the weavelet in-process can supply a custom implementation.

### DIFF
--- a/internal/tool/multi/deployer.go
+++ b/internal/tool/multi/deployer.go
@@ -353,10 +353,12 @@ func (d *deployer) startColocationGroup(g *group) error {
 			d.stop(err)
 			return err
 		})
-		if pid, ok := e.Pid(); ok {
-			if err := d.registerReplica(g, wlet, pid); err != nil {
-				return err
-			}
+		pid, ok := e.Pid()
+		if !ok {
+			panic("multi deployer child must be a real process")
+		}
+		if err := d.registerReplica(g, wlet, pid); err != nil {
+			return err
 		}
 		if err := e.UpdateComponents(components); err != nil {
 			return err
--- a/internal/tool/multi/deployer.go
+++ b/internal/tool/multi/deployer.go
@@ -353,8 +353,10 @@ func (d *deployer) startColocationGroup(g *group) error {
 			d.stop(err)
 			return err
 		})
-		if err := d.registerReplica(g, wlet, e.Pid()); err != nil {
-			return err
+		if pid, ok := e.Pid(); ok {
+			if err := d.registerReplica(g, wlet, pid); err != nil {
+				return err
+			}
 		}
 		if err := e.UpdateComponents(components); err != nil {
 			return err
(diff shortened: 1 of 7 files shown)


runtime/envelope/child.go
```diff
@@ -0,0 +1,142 @@
...
+	Start(context.Context, *protos.AppConfig) error
+
+	// Wait for the child to exit.
+	// REQUIRES: Start has been called.
+	// REQUIRES: Wait has not been called.
+	Wait() error
+
+	// Different IO streams connecting us to the child
+	Reader() io.ReadCloser  // Can be used to receive messages from the Child
+	Writer() io.WriteCloser // Can be used to receive messages from the Child
```
mwhittaker, line 43:
```suggestion
	Writer() io.WriteCloser // Can be used to send messages to the Child
```

runtime/envelope/envelope.go
```diff
@@ -153,19 +155,24 @@ func NewEnvelope(ctx context.Context, wlet *protos.EnvelopeInfo, config *protos.
...
-	}
-	runtime.OnExitSignal(func() { os.RemoveAll(tmpDir) }) // Cleanup when process exits
-
-	// Arrange to delete tmpDir if this function returns an error.
-	removeDir := true // Cleared on a successful return
-	defer func() {
-		if removeDir {
-			os.RemoveAll(tmpDir)
+	var removeDir bool
+	tmpDir := options.TmpDir
```
mwhittaker, line 159: If a user provides the `TmpDir` option, are we promising not to delete the directory? It seems like here, we don't delete it, but on line 257 in the `Serve` method, we do?

internal/tool/multi/deployer.go
```diff
@@ -353,8 +353,10 @@ func (d *deployer) startColocationGroup(g *group) error {
 			d.stop(err)
 			return err
 		})
-		if err := d.registerReplica(g, wlet, e.Pid()); err != nil {
-			return err
+		if pid, ok := e.Pid(); ok {
```
mwhittaker, line 356: What do you think about panicking if `!ok`? If I understand correctly, the multi deployer should never create an `Envelope` with an in-process `Child`, so `ok` should always be true. And if `ok` isn't true and we don't call `registerReplica`, I think the deployer may end up in an unexpected state and either crash or hang. 

And same for the SSH deployer.

mwhittaker (approved):
Very nice! I like the `Child` API and am happy to see a lot of duplicated `EnvelopeConn` code go away :partying_face:

runtime/envelope/child.go
ghemawat: Done

runtime/envelope/envelope.go
ghemawat: Great catch. Thanks.

internal/tool/multi/deployer.go
ghemawat: Done

ghemawat (commented):
Thanks!

PR #712, opened 2024-01-12
Remove envelope<->weavelet pipes.

ghemawat:
Initialization info sent by the envelope to the weavelet is now sent in an environment variable.

Initialization info sent by the weavelet to the envelope is now sent as the response to a control.WeaveletControl.InitWeavelet() call. This call is made by NewEnvelope() as an initial hand-shake right after creating the weavelet.

Deleted all the envelope/weavlet connection management code.

Deleted low-level tests that were testing details of the connection management protocol.

Dropped unnecessary code that carried a Bootstrap object in a context. We had switched away from our use of a Bootstrap context value a long time ago.


runtime/bootstrap.go
```diff
@@ -19,99 +19,43 @@ package runtime
...
-		if !ok {
-			return Bootstrap{}, fmt.Errorf("invalid type %T for bootstrap info in context", val)
-		}
-		return bootstrap, nil
-	}
-
-	str1 := os.Getenv(ToWeaveletKey)
-	str2 := os.Getenv(ToEnvelopeKey)
-	if str1 == "" && str2 == "" {
+	argsEnv := os.Getenv("WEAVELET_ARGS") //WeaveletArgsKey)
```
mwhittaker, line 44:
```suggestion
	argsEnv := os.Getenv(WeaveletArgsKey)
```

runtime/bootstrap.go
```diff
@@ -19,99 +19,43 @@ package runtime
...
 	}
 	return Bootstrap{
-		ToWeaveletFd: uintptr(toWeaveletFd),
-		ToEnvelopeFd: uintptr(toEnvelopeFd),
+		Args: args,
 	}, nil
 }
 
-// HasPipes returns true if pipe information has been supplied. This
+// Exists returns true if bootstream information has been supplied. This
```
mwhittaker, line 57:
```suggestion
// Exists returns true if bootstrap information has been supplied. This
```

mwhittaker (approved):
Awesome change!

PR #713, opened 2024-01-12
Rename WeaveletInfo to InitWeaveletReply.

ghemawat:
Fix a TODO by renaming WeaveletInfo to InitWeaveletReply. This makes the type name consistent with the types of other method return value types.

Reintroduce a version verification check that was mistakenly dropped in an earlier change.

Envelope API provides just the weavelet properties that make sense for a deployer author to worry about (like DialAddr) instead of the entire InitWeaveletReply proto, which may contain internal details.


mwhittaker (approved):
Nice!

PR #714, opened 2024-01-12
Move bulky data from EnvelopeInfo to InitWeaveletRequest.

ghemawat:
EnvelopeInfo is sent in an environment variable, so it should not be too big. Some fields in it (Sections in particular) have unbounded length, so they are moved to InitWeaveletRequest.

NewRemoteWeavelet now waits for InitWeavelet() before returning. This ensures that all information the weavelet needs is available after NewRemoteWeavelet returns.

Renamed EnvelopeInfo to WeaveletArgs.


mwhittaker (approved):
Nice!

PR #715, opened 2024-01-12
Update deployer documentation.

ghemawat:
Patch up some stale information.

Drop Appendix that describes the defunct pipe protocol.

Drop copies of Envelope and EnvelopeHandler types; instead rely on live links to the documentation of these types.

Drop stale links from the table of links at the end.


mwhittaker (approved):
Thanks Sanjay!

PR #722, opened 2024-02-05
Add instructions on how to run apps on CloudRun

rgrandl:
Running a Service Weaver application on Cloud Run (as a single container) is straightforward. Added instructions on how to do that on the Service Weaver website based on Michael's example.


mwhittaker (approved):
Nice, thanks Robert!

PR #726, opened 2024-02-14
Expose the deployment identifier to the application.

spetrovic77:
This was a feature requested by one of our users so that they can distinct between multiple deployments of the same application.

As far as the user is concerned, the string is an opaque identifier they can use to compare different deployments.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -81,11 +81,11 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/go/bin/staticcheck
-          key: staticcheck-v0.4.3
+          key: staticcheck-v0.4.6
         if: ${{ matrix.command == 'lint' }}
 
       - name: Install linter
-        run: go install honnef.co/go/tools/cmd/staticcheck@v0.4.3
+        run: go install honnef.co/go/tools/cmd/staticcheck@v0.4.6
         if: ${{ matrix.command == 'lint' }}
 
       - name: Build the weaver binary
--- a/internal/sim/executor.go
+++ b/internal/sim/executor.go
@@ -30,6 +30,7 @@ import (
 	"github.com/ServiceWeaver/weaver/runtime"
 	"github.com/ServiceWeaver/weaver/runtime/codegen"
 	"github.com/ServiceWeaver/weaver/runtime/protos"
+	"github.com/google/uuid"
 	"golang.org/x/sync/errgroup"
 )
 
@@ -286,6 +287,15 @@ func (e *executor) reset(workload Workload, fakes map[reflect.Type]any, ops []*o
 	e.nextTraceID = 1
 	e.nextSpanID = 1
 
+	// Pick a deterministic deployment ID.
+	depID, err := newUUID(e.rand)
+	if err != nil {
+		return err
+	}
+	weaverInfo := &weaver.WeaverInfo{
+		DeploymentID: depID.String(),
+	}
+
 	// Fill ref fields inside the workload struct.
 	if err := weaver.FillRefs(workload, func(t reflect.Type) (any, error) {
 		return e.getIntf(t, "op", 0)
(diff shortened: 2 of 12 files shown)


internal/weaver/remoteweavelet.go
```diff
@@ -435,6 +435,7 @@ func (w *RemoteWeavelet) GetImpl(t reflect.Type) (any, error) {
 //
 // TODO(mwhittaker): Deduplicate with localweavelet.go.
 func (w *RemoteWeavelet) createComponent(ctx context.Context, reg *codegen.Registration) (any, error) {
+	w.syslogger.Debug("creating component", "component", reg.Name)
```
mwhittaker, line 438: Just checking whether these were for debugging or you intended to check them in!

internal/sim/executor.go
```diff
@@ -286,6 +287,8 @@ func (e *executor) reset(workload Workload, fakes map[reflect.Type]any, ops []*o
 	e.nextTraceID = 1
 	e.nextSpanID = 1
 
+	id := uuid.New().String()[:8]
```
mwhittaker, line 290: I'm not sure if it's 100% necessary, but we might want the deployment id to be deterministic. What do you think about using a function like the following to generate the id? We can call it with `e.rand` as an argument.

```go
func newUUID(r *rand.Rand) (uuid.UUID, error) {
    var bytes [16]byte
    r.Read(bytes[:])
    return uuid.FromBytes(bytes[:])
}
```

mwhittaker (approved):
Sweet, thanks Srdjan!

internal/sim/executor.go
spetrovic77: Thanks, done!

internal/weaver/remoteweavelet.go
spetrovic77: oops, my bad. Removed.

spetrovic77 (commented):
Thanks!

I incorporated offline comments from @ghemawat, PTAL.

ghemawat (approved):

weaver.go
```diff
@@ -210,6 +210,15 @@ func runRemote[T any, _ PointerToMain[T]](ctx context.Context, app func(context.
...
 
+// WeaverInfo holds runtime information about a deployed application.
+type WeaverInfo struct {
+	// Unique identifier for the application deployment.
+	DeploymentID string
+
+	// TODO(spetrovic): Add other runtime fields here (e.g., application start
+	// time, name of the deployer).
+}
+
```
mwhittaker, line 221: If you want to check that `WeaverInfo` and `weaver.WeaverInfo` have the same underlying type (ignoring struct tags), I think you can do the following:

```go
var (
    _ = WeaverInfo(weaver.WeaverInfo{})
    _ = weaver.WeaverInfo(WeaverInfo{})
)
```

https://go.dev/ref/spec#Conversions describes how this conversion is legal only if the two types (i.e. `WeaverInfo` and `weaver.WeaverInfo`) have the same underlying type.

weaver.go
```diff
@@ -268,6 +279,17 @@ func (i *Implements[T]) setLogger(logger *slog.Logger) {
 	i.logger = logger
 }
 
+// Weaver returns runtime information about the deployed application.
+func (i Implements[T]) Weaver() WeaverInfo {
+	return WeaverInfo{
```
mwhittaker, line 284: I think `return WeaverInfo(i.weaverInfo)` might also work?

.github/workflows/go.yml
```diff
@@ -81,11 +81,11 @@ jobs:
         uses: actions/cache@v3
         with:
           path: ~/go/bin/staticcheck
-          key: staticcheck-v0.4.3
+          key: staticcheck-v0.4.6
         if: ${{ matrix.command == 'lint' }}
 
       - name: Install linter
-        run: go install honnef.co/go/tools/cmd/staticcheck@v0.4.3
+        run: go install honnef.co/go/tools/cmd/staticcheck@v0.4.6
```
mwhittaker, line 88: Thanks for fixing this!

mwhittaker (approved):

weaver.go
spetrovic77: Great suggestion, done!

weaver.go
spetrovic77: It does, thanks!

.github/workflows/go.yml
spetrovic77: Thanks!

PR #738, opened 2024-03-07
Update doc regarding how to set the project/account for GKE.

rgrandl:
The documentation is obsolete. If you want to deploy an app in a different Google Cloud Project and/or using a different Google Cloud account, the way to do it is to pass this information using flags. The docs state that this should be specified in the config.

Also, updated the doc to mention that the user should specify at least one region where to rollout the application.


rgrandl:
This was fixed in #744.

PR #761, opened 2024-05-09
Check the ready status of component before it becomes reachable.

spetrovic77:
This fixes an situation where the component takes a while to initialize: previously, we were declaring the component as activated/healthy even during the pending init, resulting in subsequent method calls failing.

Thanks to rgrandl@ for debugging the issue.

Other changes:
  * Don't add a component to the routing info until it has been initialized. This fixes the health-checking situation where one component replica becomes healthy, but we subsequently send method calls to other replicas which fail.
  * Fix go vet in a couple of places.


mwhittaker (approved):
Very elegant!

rgrandl (approved):

PR #767, opened 2024-05-29
status: map weavelet id to pids for `multi`, `single` and `ssh` deployment.

flouthoc:
As of now weaver logs shows the weavelet id but not `pids` and `weaver multi status` shows the respective `pid` but its hard to map weavelet id from the logs to the running process, following commit adds weavelet id with the pids of the components so its easier to debug.

weavelet_id


flouthoc:
@rgrandl @mwhittaker PTAL

rgrandl (commented):
Thank you for the change. This is a great idea. 

You'll also need to fix the single deployer and the ssh deployer, because they use the same status package.

internal/status/status.proto
```diff
@@ -37,6 +37,7 @@ message Component {
   string name = 1;              // component name (e.g., Cache)
   repeated int64 pids = 3;      // PIDs of component replicas
   repeated Method methods = 4;  // methods
+  map<int64, string> weaveletId = 5;
```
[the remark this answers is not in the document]
flouthoc: Yes sure that can be done I am okay to add a new replica field, i added a new field because its a non-breaking change to the previously defined proto API and I was not sure if okay to do breaking changes.

If you are okay with it then sure I can add a new struct :)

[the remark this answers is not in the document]
rgrandl: Good point. I think the only thing that can break is the `ssh` deployer, although I am not aware of anyone using it in prod.

I think we can add the new struct.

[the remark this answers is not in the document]
flouthoc: Cool thanks.

flouthoc:
@rgrandl PTAL again. I have tested both `single` and `multi` deploy but I was not able to test `ssh` although I think my code changes should work without any issues. Do we have test for `ssh` in CI ? or should i setup `ssh` in my environment.

flouthoc:
@rgrandl I might be able to test on `ssh` as well today, but irrespective of that I think PR is ready for a review again.

flouthoc:
> @rgrandl I might be able to test on `ssh` as well today, but irrespective of that I think PR is ready for a review again.

Well I configured `ssh` setup only to figure out there is no `status` command in `ssh` deployer at all, there is a `dashboard` command but it is a webview. @rgrandl is there any `cli` command in `ssh` which is equivalent to `status`.

internal/status/status.go
```diff
@@ -111,22 +111,26 @@ func formatComponents(w io.Writer, statuses []*Status) {
...
+				return component.Replicas[i].Pid < component.Replicas[j].Pid
 			})
-			pids := make([]string, len(component.Pids))
-			for i, pid := range component.Pids {
-				pids[i] = fmt.Sprint(pid)
+			pids := make([]string, len(component.Replicas))
+			weaveletIds := make([]string, len(component.Replicas))
+			for i, replica := range component.Replicas {
+				pids[i] = fmt.Sprint(replica.Pid)
+				if len(replica.WeaveletId) > 0 {
```
rgrandl, line 129: I think we should let it crash here. If we don't set the weavelet id somehow, we shouldn't hide it.

flouthoc:
@rgrandl Please take a look again.

rgrandl (approved):
Looks great! Thanks a lot!

rgrandl:
> > @rgrandl I might be able to test on `ssh` as well today, but irrespective of that I think PR is ready for a review again.
> 
> Well I configured `ssh` setup only to figure out there is no `status` command in `ssh` deployer at all, there is a `dashboard` command but it is a webview. @rgrandl is there any `cli` command in `ssh` which is equivalent to `status`.

There is no status command in the `ssh` deployer apparently.

internal/tool/ssh/impl/manager.go
```diff
@@ -316,12 +315,11 @@ func (m *manager) Status(ctx context.Context) (*status.Status, error) {
 		cs := maps.Keys(g.components.Val)
 		g.components.Unlock()
 		g.mu.Lock()
```
rgrandl, line 317: Missed this one. Can you remove lines 317/318?

rgrandl:
Can you also run `./dev/build_and_test tidy generate build vet test testrace`

internal/tool/ssh/impl/manager.go
```diff
@@ -313,15 +312,12 @@ func (m *manager) Status(ctx context.Context) (*status.Status, error) {
 	var components []*status.Component
 	for _, g := range m.allGroups() {
 		g.components.Lock()
-		cs := maps.Keys(g.components.Val)
-		g.components.Unlock()
-		g.mu.Lock()
```
rgrandl, line 318: let's remove just this:
`
g.mu.Lock()
pids := slices.Clone(g.pids)
g.mu.Unlock()
`

flouthoc: Yes fixed it now, though I was doing something wrong the last time :)

rgrandl (approved):

Issue #556, opened 2023-08-23
feat: add check rules to avoid cycle Ref


sysulq:
When we change the examples/chat below, it's not hard to realise localCache and scaler are mutually dependent, which would lead to a stack overflow.

```
type localCache struct {
	weaver.Implements[LocalCache]
	mu    sync.Mutex
	test  weaver.Ref[ImageScaler]
	cache *lru.Cache[string, string]
	// TODO: Eviction policy.
}


type scaler struct {
	weaver.Implements[ImageScaler]
	weaver.Ref[LocalCache]
}
```

```
➜  chat git:(main) ✗ go run .                               
╭───────────────────────────────────────────────────╮
│ app        : chat                                 │
│ deployment : 1d053701-be60-4e2e-b170-d2d40d12423a │
╰───────────────────────────────────────────────────╯
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc02058c388 stack=[0xc02058c000, 0xc04058c000]
fatal error: stack overflow

runtime stack:
runtime.throw({0xd9b445?, 0x7f49e1f1fc70?})
        /home/liqi/.gvm/gos/go1.21.0/src/runtime/panic.go:1077 +0x5c fp=0x7f49e1f1fc20 sp=0x7f49e1f1fbf0 pc=0x43be7c
runtime.newstack()
        /home/liqi/.gvm/gos/go1.21.0/src/runtime/stack.go:1107 +0x5ac fp=0x7f49e1f1fdd0 sp=0x7f49e1f1fc20 pc=0x4563ec
runtime.morestack()
        /home/liqi/.gvm/gos/go1.21.0/src/runtime/asm_amd64.s:593 +0x8f fp=0x7f49e1f1fdd8 sp=0x7f49e1f1fdd0 pc=0x46daef
```

mwhittaker:
Oh great find! Yeah, we should either detect and disallow cycles or allow cycles. We'll work on this.

mwhittaker:
Oh great find! Yeah, we should either detect and disallow cycles or allow cycles. We'll work on this.

renanbastos93:
@mwhittaker Can you help to understand more about for me fix that?
I believe that we need to improve checker when to generate files `*_gen.go` to ensure this merge cycle, right?

sysulq:
@renanbastos93 
I fixed this problem by checking the cycle at runtime, details can be found in the link below, just for your information :-)

[...]

PR #772, opened 2024-06-07
feat(#556): added check circular dependencies

renanbastos93:
Added validation to ensure that dependencies do not form circular dependencies. 
An error is raised if such a condition is detected.

ISSUE: #556 
Repository reference to simulate it: https://github.com/renanbastos93/weaver-issue-556

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/BurntSushi/toml v1.3.2
 	github.com/DataDog/hyperloglog v0.0.0-20220804205443-1806d9b66146
 	github.com/alecthomas/chroma/v2 v2.2.0
+	github.com/dominikbraun/graph v0.23.0
 	github.com/fsnotify/fsnotify v1.6.0
 	github.com/go-sql-driver/mysql v1.7.1
 	github.com/goburrow/cache v0.1.4
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	github.com/BurntSushi/toml v1.3.2
 	github.com/DataDog/hyperloglog v0.0.0-20220804205443-1806d9b66146
 	github.com/alecthomas/chroma/v2 v2.2.0
-	github.com/dominikbraun/graph v0.23.0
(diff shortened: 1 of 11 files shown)


renanbastos93:
@mwhittaker @sysulq @rgrandl  

Does anyone have comments on this? Is it good, or do I need to change this commit?

rgrandl:
@renanbastos93, thanks for the PR. 

Few comments:
1) We already have a graph package, so we would prefer not to get a dependency on another graph library (`github.com/dominikbraun/graph`)
2) A better place to check for circular dependencies is in the deployer. For example, in the `multi` deployer, you can do the check when you read the component graph `https://github.com/ServiceWeaver/weaver/blob/main/internal/tool/multi/deployer.go#L250`. Similar path should exist for the other deployers.

renanbastos93:
> @renanbastos93, thanks for the PR.
> 
> Few comments:
> 
> 1. We already have a graph package, so we would prefer not to get a dependency on another graph library (`github.com/dominikbraun/graph`)
> 2. A better place to check for circular dependencies is in the deployer. For example, in the `multi` deployer, you can do the check when you read the component graph `https://github.com/ServiceWeaver/weaver/blob/main/internal/tool/multi/deployer.go#L250`. Similar path should exist for the other deployers.

awesome, i will use that. Thanks a lot soon i will do a new commit.

renanbastos93:
@rgrandl done, check please.

rgrandl (commented):
Sorry for the late reply on this, Renan. 

In dfs.go, we have a dfs function that does a DFS traversal of the graph already. Is it possible to modify that function to detect cycles instead of implementing another traversal?

renanbastos93:
> Sorry for the late reply on this, Renan.
> 
> In dfs.go, we have a dfs function that does a DFS traversal of the graph already. Is it possible to modify that function to detect cycles instead of implementing another traversal?

No worries! I can do this, but I will need more time because I've started a new job and my free time has been limited.

rgrandl:
This is not something critical and has been deprioritized.

PR #780, opened 2024-07-10
Reduce number of buckets for auto-generated metrics

rgrandl:
For the auto-generated histogram metrics, we use a large number of buckets. The largest bucket captures values up to 50 quintillion.

However, having a large number of buckets results in a significant size for exported metrics, even if they carry 0 data (as most buckets will have 0 values). This leads to: (1) higher costs for the user (e.g., Google Cloud monitoring charges you based on the volume of metrics exported), and (2) increased system load due to handling large metrics data.

This PR reduces the number of buckets to capture values up to 5 billion (5e9),
which seems reasonable in practice. E.g., auto-generated metrics using these buckets:
1) http_request_latency_micros - even if this is 10 minutes, the value
   is still 6e7
2) http_request_bytes_received - http request size of more than 5 GB seem
   rare
3) http_request_bytes_returned - http response size of more than 5 GB
   seem rare
4) serviceweaver_method_latency_micros - component method execution of
   more than 1.3 hours seem rare
5) serviceweaver_method_bytes_request - request size of more than 5 GB
   seem rare
6) serviceweaver_method_bytes_reply - response size of more than 5 GB
   seem rare

### DIFF
--- a/runtime/codegen/metrics.go
+++ b/runtime/codegen/metrics.go
@@ -17,42 +17,34 @@ package codegen
 import (
 	"time"
 
+	imetrics "github.com/ServiceWeaver/weaver/internal/metrics"
 	"github.com/ServiceWeaver/weaver/metrics"
-)
-
-// Names of automatically populated metrics.
-const (
-	MethodCountsName       = "serviceweaver_method_count"
-	MethodErrorsName       = "serviceweaver_method_error_count"
-	MethodLatenciesName    = "serviceweaver_method_latency_micros"
-	MethodBytesRequestName = "serviceweaver_method_bytes_request"
-	MethodBytesReplyName   = "serviceweaver_method_bytes_reply"
 )
 
 var (
 	// The following metrics are automatically populated for the user.
 	methodCounts = metrics.NewCounterMap[MethodLabels](
-		MethodCountsName,
+		imetrics.MethodCountsName,
 		"Count of Service Weaver component method invocations",
 	)
 	methodErrors = metrics.NewCounterMap[MethodLabels](
-		MethodErrorsName,
+		imetrics.MethodErrorsName,
 		"Count of Service Weaver component method invocations that result in an error",
 	)
 	methodLatencies = metrics.NewHistogramMap[MethodLabels](
-		MethodLatenciesName,
+		imetrics.MethodLatenciesName,
 		"Duration, in microseconds, of Service Weaver component method execution",
-		metrics.NonNegativeBuckets,
+		imetrics.GeneratedBuckets,
 	)
 	methodBytesRequest = metrics.NewHistogramMap[MethodLabels](
-		MethodBytesRequestName,
+		imetrics.MethodBytesRequestName,
 		"Number of bytes in Service Weaver component method requests",
-		metrics.NonNegativeBuckets,
+		imetrics.GeneratedBuckets,
 	)
 	methodBytesReply = metrics.NewHistogramMap[MethodLabels](
-		MethodBytesReplyName,
+		imetrics.MethodBytesReplyName,
 		"Number of bytes in Service Weaver component method replies",
-		metrics.NonNegativeBuckets,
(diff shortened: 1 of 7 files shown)


metrics/buckets.go
```diff
@@ -27,15 +27,15 @@ var NonNegativeBuckets = []float64{
...
-	100000000000, 200000000000, 500000000000,
-	1000000000000, 2000000000000, 5000000000000,
-	10000000000000, 20000000000000, 50000000000000,
-	100000000000000, 200000000000000, 500000000000000,
-	1000000000000000, 2000000000000000, 5000000000000000,
-	10000000000000000, 20000000000000000, 50000000000000000,
-	100000000000000000, 200000000000000000, 500000000000000000,
-	1000000000000000000, 2000000000000000000, 5000000000000000000,
-	10000000000000000000, 20000000000000000000, 50000000000000000000, // i.e., 5e19
+	1000000000, 2000000000, 5000000000, // i.e., 5e9
```
spetrovic77, line 30: Can we leave `NonNegativeBuckets` as-is, since it's not just used by the auto-generated metrics but is public to the user?

Instead, let's create a different internal variable in `weaver/http.go` and use that for the auto-generated metrics.

rgrandl: Good point.

weaver/http is in package weaver, so given that I have to use custom buckets in 3 different places, I decided to  define the buckets locally in each file.

Regarding `NonNegativeBuckets` - I feel that the number of buckets is really large for any customer, and they will keep complaining if this is the only option we provide. They can define their custom buckets when they create histograms though. I am not sure if we should provide any `NonNegativeBuckets` at all (this way the user will always have to define their buckets), or if we should provide a few options with different number of buckets.

rgrandl (commented):
Thank you

runtime/codegen/metrics.go
```diff
@@ -42,18 +42,33 @@ var (
...
 	)
 	methodBytesReply = metrics.NewHistogramMap[MethodLabels](
 		MethodBytesReplyName,
 		"Number of bytes in Service Weaver component method replies",
-		metrics.NonNegativeBuckets,
+		nonNegativeBuckets,
 	)
+
+	// nonNegativeBuckets provides rounded bucket boundaries for the autogenerated histograms.
+	nonNegativeBuckets = []float64{
```
spetrovic77, line 59: Can this be defined in `weaver/internal/metrics.go`?

If so, maybe best to put it that so that it can be shared. If not, it's fine as you have it.

metrics/buckets.go
spetrovic77: I'd be in favor of removing it, if it's not used by any other code.

rgrandl (commented):
Thank you

runtime/codegen/metrics.go
rgrandl: Good point. It requires a little bit of massaging though.

metrics/buckets.go
rgrandl: Sounds good. I will remove it in a future PR.

internal/metrics/buckets.go
```diff
@@ -0,0 +1,34 @@
...
+// limitations under the License.
+
+package metrics
+
+// NonNegativeBuckets provides rounded bucket boundaries for histograms
+// that will only store non-negative values.
+//
+// Note that these buckets are intended to be used only by the metrics generated
+// by the weaver runtime.
+var NonNegativeBuckets = []float64{
```
spetrovic77, line 22: Maybe move to `generated.go` and rename to `GeneratedBuckets`, so it's clear?

spetrovic77 (approved):

internal/metrics/buckets.go
rgrandl: Sounds good. Done

rgrandl (commented):
Thanks, Srdjan

PR #794, opened 2024-08-21
Fix traffic graph

rgrandl:
After we introduced system components (weavelet control and deployment control), the traffic graph doesn't display anything in the weaver multi dashboard.

This is because when we display the graph, we look at the components of the application and then we try to match them to the traffic between them. However, the edges show traffic for components that are missing from the status (e.g., the system components). Hence, when we display the graph there is no 1:1 mapping between components and the traffic between them.

This PR fixes this issue by not displaying traffic for system components (in fact the user shouldn't care about that). Maybe a better fix would be to do this in the javascript itself, but we can do that in the future if needed.

### DIFF
--- a/cmd/weaver/main.go
+++ b/cmd/weaver/main.go
@@ -165,7 +165,7 @@ func run(deployer string, args []string) (int, error) {
 	binary := "weaver-" + deployer
 	if _, err := exec.LookPath(binary); err != nil {
 		msg := fmt.Sprintf(`"weaver %s" is not a weaver command. See "weaver --help". If you're trying to invoke a custom deployer, the %q binary was not found. You may need to install the %q binary or add it to your PATH.`, deployer, binary, binary)
-		return 1, fmt.Errorf(wrap(msg, 80))
+		return 1, fmt.Errorf("%s", wrap(msg, 80))
 	}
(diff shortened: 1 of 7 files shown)


cmd/weaver/main.go
```diff
@@ -165,7 +165,7 @@ func run(deployer string, args []string) (int, error) {
 	binary := "weaver-" + deployer
 	if _, err := exec.LookPath(binary); err != nil {
 		msg := fmt.Sprintf(`"weaver %s" is not a weaver command. See "weaver --help". If you're trying to invoke a custom deployer, the %q binary was not found. You may need to install the %q binary or add it to your PATH.`, deployer, binary, binary)
-		return 1, fmt.Errorf(wrap(msg, 80))
+		return 1, fmt.Errorf(wrap(msg, 80), "%s")
```
mwhittaker, line 168: Sorry for not understanding, but what does the `"%s"` do?

mwhittaker (approved):

rgrandl (commented):
Thanks, Michael

cmd/weaver/main.go
mwhittaker: Ah, gotcha!

PR #796, opened 2024-09-04
Add build tags support

rgrandl:
`go build` allows the user to specify build tags. These tags enable the user to discard files based on various tags when they build the application binary.

`weaver generate` which is a wrapper around `go build` doesn't allow the user to specify build tags.

This PR adds built tags support for the `weaver generate` command.

The syntax of passing build tags to the `weaver generate` command is similar to how build tags are specified when using `go build`.

E.g.,

weaver generate -tags good,prod
weaver generate --tags=good

### DIFF
--- a/cmd/weaver/main.go
+++ b/cmd/weaver/main.go
@@ -74,25 +74,19 @@ func main() {
 	switch flag.Arg(0) {
 	case "generate":
 		generateFlags := flag.NewFlagSet("generate", flag.ExitOnError)
-		tags := generateFlags.String("tags", "not specified", "Optional tags for the generate command")
+		tags := generateFlags.String("tags", "", "Optional tags for the generate command")
 		generateFlags.Usage = func() {
 			fmt.Fprintln(os.Stderr, generate.Usage)
 		}
 		generateFlags.Parse(flag.Args()[1:])
-		buildTags := "--tags=ignoreWeaverGen"
-		if *tags != "not specified" { // tags flag was specified
-			if *tags == "" || *tags == "." {
-				// User specified the tags flag but didn't provide any tags.
-				fmt.Fprintln(os.Stderr, "No tags provided.")
-				os.Exit(1)
-			}
+		buildTags := "ignoreWeaverGen"
+		if *tags != "" { // tags flag was specified
 			// TODO(rgrandl): we assume that the user specify the tags properly. I.e.,
 			// a single tag, or a list of tags separated by comma. We may want to do
 			// extra validation at some point.
 			buildTags = buildTags + "," + *tags
 		}
-		idx := len(flag.Args()) - len(generateFlags.Args())
-		if err := generate.Generate(".", flag.Args()[idx:], generate.Options{BuildTags: buildTags}); err != nil {
+		if err := generate.Generate(".", generateFlags.Args(), generate.Options{BuildTags: buildTags}); err != nil {
 			fmt.Fprintln(os.Stderr, err)
 			os.Exit(1)
 		}
--- a/cmd/weaver/main.go
+++ b/cmd/weaver/main.go
@@ -74,11 +74,25 @@ func main() {
 	switch flag.Arg(0) {
 	case "generate":
 		generateFlags := flag.NewFlagSet("generate", flag.ExitOnError)
+		tags := generateFlags.String("tags", "not specified", "Optional tags for the generate command")
 		generateFlags.Usage = func() {
 			fmt.Fprintln(os.Stderr, generate.Usage)
 		}
 		generateFlags.Parse(flag.Args()[1:])
-		if err := generate.Generate(".", flag.Args()[1:], generate.Options{}); err != nil {
+		buildTags := "--tags=ignoreWeaverGen"
+		if *tags != "not specified" { // tags flag was specified
+			if *tags == "" || *tags == "." {
+				// User specified the tags flag but didn't provide any tags.
+				fmt.Fprintln(os.Stderr, "No tags provided.")
+				os.Exit(1)
+			}
+			// TODO(rgrandl): we assume that the user specify the tags properly. I.e.,
+			// a single tag, or a list of tags separated by comma. We may want to do
+			// extra validation at some point.
+			buildTags = buildTags + "," + *tags
+		}
+		idx := len(flag.Args()) - len(generateFlags.Args())
+		if err := generate.Generate(".", flag.Args()[idx:], generate.Options{BuildTags: buildTags}); err != nil {
 			fmt.Fprintln(os.Stderr, err)
 			os.Exit(1)
 		}
--- a/internal/tool/generate/generator.go
+++ b/internal/tool/generate/generator.go
@@ -54,7 +54,7 @@ const (
 	Usage = `Generate code for a Service Weaver application.
 
 Usage:
-  weaver generate [tags] [packages]
+  weaver generate [-tags taglist] [packages]
 
 Description:
   "weaver generate" generates code for the Service Weaver applications in the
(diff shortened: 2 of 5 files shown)


cmd/weaver/main.go
```diff
@@ -74,11 +74,25 @@ func main() {
 	switch flag.Arg(0) {
 	case "generate":
 		generateFlags := flag.NewFlagSet("generate", flag.ExitOnError)
+		tags := generateFlags.String("tags", "not specified", "Optional tags for the generate command")
```
ghemawat, line 77: I am not quite sure why we need "not specified". Can't we just say that if empty, we add no tags except for our internal weaver tag, and if non-empty, we add the specified tags + our internal weaver tag. 

If we really need to detect whether or not tags were specified on the command line, a less messy alternative to "not specified" would be to call flags.Visit() to see if it reports "tags" as one of the flags set on the command line.

cmd/weaver/main.go
```diff
@@ -74,11 +74,25 @@ func main() {
...
 		generateFlags.Usage = func() {
 			fmt.Fprintln(os.Stderr, generate.Usage)
 		}
 		generateFlags.Parse(flag.Args()[1:])
-		if err := generate.Generate(".", flag.Args()[1:], generate.Options{}); err != nil {
+		buildTags := "--tags=ignoreWeaverGen"
+		if *tags != "not specified" { // tags flag was specified
+			if *tags == "" || *tags == "." {
```
ghemawat, line 84: Why do we need a special case for "."?

cmd/weaver/main.go
```diff
@@ -74,11 +74,25 @@ func main() {
...
+				// User specified the tags flag but didn't provide any tags.
+				fmt.Fprintln(os.Stderr, "No tags provided.")
+				os.Exit(1)
+			}
+			// TODO(rgrandl): we assume that the user specify the tags properly. I.e.,
+			// a single tag, or a list of tags separated by comma. We may want to do
+			// extra validation at some point.
+			buildTags = buildTags + "," + *tags
+		}
+		idx := len(flag.Args()) - len(generateFlags.Args())
```
ghemawat, line 94: Why do we need to use flag.Args()[idx:] instead passing generateFlags.Args() to generate.Generate?

internal/tool/generate/generator.go
```diff
@@ -54,7 +54,7 @@ const (
 	Usage = `Generate code for a Service Weaver application.
 
 Usage:
-  weaver generate [packages]
+  weaver generate [tags] [packages]
```
ghemawat, line 57: Change to:

```
weaver generate [-tags taglist] [packages]
```

internal/tool/generate/generator.go
```diff
@@ -86,13 +89,21 @@ Examples:
...
+  top of the file.
+  weaver generate -tags good,prod`
 )
 
 // Options controls the operation of Generate.
 type Options struct {
-	// If non-nil, use the specified function to report warnings.
-	Warn func(error)
+	Warn      func(error) // If non-nil, use the specified function to report warnings
+	BuildTags string
```
ghemawat, line 106: It seems like the BuildTags fields includes the "-tags=" prefix. That seems a bit wrong to me. Could just contain the tags, and we would add the -tags= prefix below when setting BuildFlags:

```
cf.BuildFlags = []string{"-tags", opt.BuildTags}
```

cmd/weaver/main.go
```diff
@@ -74,11 +74,25 @@ func main() {
 	switch flag.Arg(0) {
 	case "generate":
 		generateFlags := flag.NewFlagSet("generate", flag.ExitOnError)
+		tags := generateFlags.String("tags", "not specified", "Optional tags for the generate command")
 		generateFlags.Usage = func() {
 			fmt.Fprintln(os.Stderr, generate.Usage)
 		}
 		generateFlags.Parse(flag.Args()[1:])
-		if err := generate.Generate(".", flag.Args()[1:], generate.Options{}); err != nil {
+		buildTags := "--tags=ignoreWeaverGen"
```
ghemawat, line 82: See below. I think we should just have "ignoreWeaverGen" here and supply --tags= prefix later.

internal/tool/generate/generator_test.go
```diff
@@ -237,6 +238,80 @@ func TestGenerator(t *testing.T) {
 	}
 }
 
+// TestGeneratorBuildWithTags runs "weaver generate" on all the files in
+// testdata/tags and checks if the command succeeds. Each file should have some build tags.
+func TestGeneratorBuildWithTags(t *testing.T) {
```
ghemawat, line 243: Why can't we use runGenerator for this test?

ghemawat (commented):
Looks reasonable, just some nits.

cmd/weaver/main.go
rgrandl: Good point. I think I implemented this before understanding that I need to define tags as a flag on generateFlags instead of defining a new flag set.

cmd/weaver/main.go
rgrandl: We don't. Removed all this complicated logic.

cmd/weaver/main.go
rgrandl: Great point again.

internal/tool/generate/generator.go
rgrandl: Done

internal/tool/generate/generator.go
rgrandl: Good point. Fixed

cmd/weaver/main.go
rgrandl: Dobe

internal/tool/generate/generator_test.go
rgrandl: runGenerator copies every single file in a temp directory, and tries to build it. I chose this approach because I wanted to run go build on all the files in the directory and check that weaver_gen contains things only for the good files. That adds a little bit of boilerplate code for sure. If you have a strong preference, I can switch to calling runGenerator instead.

rgrandl (commented):
Thanks Sanjay

internal/tool/generate/generator_test.go
ghemawat: I am fine with you keeping what you have, but I don't quite understand the downside of using runGenerator. It would copy all of the files and build it. You can still check that weaver_gen.go does not contain the components defined in the bad file.

rgrandl: Done

rgrandl (commented):
Thanks, Sanjay

internal/tool/generate/generator_test.go
```diff
@@ -219,7 +225,7 @@ func TestGenerator(t *testing.T) {
 			}
 
 			// Run "weaver generate".
-			output, err := runGenerator(t, dir, filename, contents, []string{"sub1", "sub2"})
+			output, err := runGenerator(t, dir, filename, contents, []string{"sub1", "sub2"}, []string{})
```
ghemawat, line 228: Can you just pass nil for the empty slice?

internal/tool/generate/generator_test.go
```diff
@@ -135,7 +136,12 @@ func runGenerator(t *testing.T, directory, filename, contents string, subdirs []
 	if err := tidy.Run(); err != nil {
 		t.Fatalf("go mod tidy: %v", err)
 	}
-	gobuild := exec.Command("go", "build")
+
+	var buildTagsArg string
+	if len(buildTags) > 0 {
```
ghemawat, line 141: I thought the earlier code always added "ignoreWeaveGen". So can we supply -tags unconditionally?

ghemawat (approved):
Sorry about the delay. Looks great. A couple of tiny comments you can ignore if you wish.

internal/tool/generate/generator_test.go
rgrandl: Done

internal/tool/generate/generator_test.go
rgrandl: Great observation. Fixed

rgrandl (commented):
Thanks a lot, Sanjay.

PR #803, opened 2024-11-05
Turndown message

rgrandl:


README.md
```diff
@@ -1,5 +1,42 @@
...
+
+<div class="announcement">
+  <p style="color: inherit">Service Weaver began as an exploratory initiative to
+    understand the challenges of developing, deploying, and maintaining distributed
+    applications. We were excited by the strong interest from the developer community,
+    which led us to open-source the project. 
+  </p>
+  <p style="color: inherit">We greatly appreciate the continued advocacy and support
+    of the Service Weaver community. However, we have learned what we needed from
+    our exploration, and we regret to announce that, <b>effective December 5, 2024,
```
ghemawat, line 28: Alternate wording suggestion:

However, we realized that it was hard for users to adopt Service Weaver directly since it required rewriting large parts of existing applications. Therefore Service Weaver did not see much direct use and therefore <b>effective ...

website/index.html
```diff
@@ -27,6 +27,13 @@ <h1>
   <a href="docs.html">Read the Docs</a>
 </div>
 
+<div class="announcement-banner">
+  <p>We regret to announce that, effective <b>December 5, 2024</b>, we will no longer be able to
+    continue active development on the Service Weaver open-source framework. For more details, click
```
ghemawat, line 32: s/no longer be able to continue active/no longer continue/

README.md
```diff
@@ -1,5 +1,42 @@
...
+    we will transition Service Weaver into maintenance mode</b>. We recommend that users
```
spetrovic77, line 29: The sentence "We recommend that..." is a bit in the way of the overall flow. I think it can be removed, or maybe moved to the end of this paragraph.

README.md
rgrandl: Updated the text.

README.md
rgrandl: Good point. Yeah, I felt the same. However, we should let users know what would be a better approach maybe (to fork the repo). I moved that sentence to the end of the paragraph.

rgrandl (commented):
Thank you for the feedback

ghemawat (approved):

